PR #4, opened 2010-11-27
Fix tests

lemenkov:
This patch fixes the only constantly failing test. 

The actual result of this test is opposite to what was expected. We're trying to know whether or nor we need to rebalance current list of nodes [{nonode@nohost,4}, {othernode2@otherhost2,6}, {othernode@otherhost,6}]. The expected number of partitions is 5, so no need to re-balance yet.


jonmeredith:
Thanks for the patch - that test case wasn't updated for the refactor back in 72967874 to prevent over-eager rebalancing.

PR #7, opened 2010-12-01
Bz882 cluster info

slfritchie:
Reviewer: anyone interested

Update packaging for cluster_info, which is now under the basho org @ github.
Add a really basic set of info doodads for riak_core.  Arguably, some of
the cluster_info goop that I've added for riak_kv really belong to riak_core?
But as I've implemented the goop now, riak_core's info doesn't (I hope) call
any riak_kv code.


beerriot:
+1 looks good & works for me

PR #9, opened 2010-12-14
Bz891 disable nagle

dizzyd:
Adding an option to disable nagle on HTTP bindings. Note that this option is global across all bindings, for the moment. Default is to leave Nagle enabled.


jonmeredith:
+1 - default is false, will be interesting to see if it changes the number of packets exchanged with it turned enabled.

PR #10, opened 2010-12-21
Missing parameter in call to error_logger:error_msg/2 when handoff exits abnormally

rtilder:
The resulting spew from riak_err tends to mask the root problem.  Many graciases for the merging of the Abby Normal fixes.

--Ryan


rtilder:
This is the most amazing pull request I've ever reviewed.  Doubleplusmerged.

--Ryan

jonmeredith:
+1

PR #11, opened 2011-01-28
Bz971 stats latency

slfritchie:
Pull request 3 of 3 for bz971

I expect this one to be controversial.  The changes to slide.erl require some place to put temp files on disk.  There is no global config option right now for the data dir, and you can't rely on your OS proc's cwd to be the data directory or even that cwd is writable by the "riak" user.  For 1M operations/min, these temp files use about 15MB of space per slide.

The spiral change may also be controversial, but I'm sticking to my guns: we aren't using any stats beyond 1 minute, so we shouldn't maintain them.  And there's genuine evidence that the extra spiral data maintenance is causing latency problems (see bz971 and related ZenDesk ticket).  Therefore, the surgery to remove minute, hour, and day stats is justified.  The defense rests.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -3,6 +3,7 @@
 {deps, [
   {protobuffs, "0.5.1", {git, "git://github.com/basho/erlang_protobuffs",
                               {tag, "protobuffs-0.5.1"}}},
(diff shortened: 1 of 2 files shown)


jonmeredith:
+1 merge away

PR #12, opened 2011-01-28
Merging default_bucket_props from app.config

beerriot:
The first of these patches fixes the problem where specifying default bucket properties in app.config is an all-or-nothing venture.  Anything specified in app.config is now merged with the defaults specified in riak_core_app.erl (the same defaults as before, just in a more accessible place).  The settings in app.config override the hard-coded ones in the source, of course.

The second patch fixes the problem where app.config bucket properties might cause duplicate property entries in the list.  This fixes some confusion and fragility.

The third patch just cleans up and standardizes the prop-merging code.

### DIFF
--- a/ebin/riak_core.app
+++ b/ebin/riak_core.app
@@ -76,14 +76,6 @@
          {wants_claim_fun, {riak_core_claim, default_wants_claim}},
          {choose_claim_fun, {riak_core_claim, default_choose_claim}},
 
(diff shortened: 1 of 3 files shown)


jonmeredith:
Looks good to me - I suggest we file another bug to cover the startup bad buckets case and close bz://123.

It's much better than what we have.  +1.

PR #14, opened 2011-02-03
Bz823 handoff processing

slfritchie:
Reviewer: whoever's free.

As mentioned in the BZ 823 comments, this one seems quite odd.
BZ 719 fixed the function_clause error that triggered the first cascade,
but it isn't clear why the riak_kv_handoff_listener would still be alive
4 full seconds after a proc that it linked to had crashed.


jonmeredith:
+1 - it's a weird one for backward compatibility, and the fix tests safely for me.

I think we've had enough releases that the proxys can go away soon.

PR #15, opened 2011-02-08
Correct riak_core_bucket's eunit test

beerriot:
This test fixes riak-core_bucket's eunit test case by making sure that riak_core's application environment default_bucket_props is defined to at least be the empty list.  In a recent change, the riak_core application starts up with that env undefined, and then sets it in riak_core_app:start/2, but that function is not run during eunit.

Reviewer: Scott (the reporter)


slfritchie:
+1

PR #16, opened 2011-02-11
Bz520 double wrap req

slfritchie:
Reviewer: anyone interested in purifying FSMs for better testing.
Reviewing the new code is probably best done looking at the
final files than at all the diffs in between.

This is the first stab at trying to purify a FSM for testing purposes.
I'm thinking that the new module, gen_fsm_test_driver.erl, should
be pulled out of riak_core/test because I'm thinking it will be
immediately useful for testing KV FSMs (I've already started on
the get & put FSMs).  So, some thought about where to put it
(in yet another repo? {shrug}) would be helpful.

The unit test for verifying 520's fix got a bit more involved than I'd
originally thought it'd be.  See the comment in the middle about
adding -ifdef(TEST)-only code for simplifying testing.

Also, it's possible that more complicated testing scenarios may
require more flexibility, e.g. return value X the first time that an
impure-function-simulated-as-pure is called but return Y a 2nd time.
It isn't clear to me how'd be best to do that: use meck or other
more traditional code mocking tool, resort to relying on stateful
3rd-party processes to interact with (thus tainting otherwise
beautiful purity), or something else.

Note on the use of erlang:apply() ... it isn't as slow as it used to be,
and it _really_ makes the code a lot, lot, lot shorter and easier to
read.  For contrast, see changeset 97caae5c3e55065ceffe2de8e70b0ed793764157
for what the code looked like before switching to using erlang:apply().
Or see the diff from moving to the not-yet-finished purification
using the old scheme -> new apply scheme:
https://github.com/basho/riak_core/commit/93f49d19f1b20f0fd3cda56ec0094e93ff453377

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -196,16 +196,16 @@ handle_sync_event({handoff_data,BinObj}, _From, StateName,
 handle_info(_Info, StateName, State) ->
     {next_state, StateName, State, State#state.inactivity_timeout}.
 
-terminate(Reason, _StateName, #state{mod=Mod, modstate=ModState} = State) ->
+terminate(Reason, _StateName, State=#state{mod=Mod, modstate=ModState}) ->
     impure(State, {mod, Mod}, terminate, [Reason, ModState]),
     ok.
 
 code_change(_OldVsn, StateName, State, _Extra) ->
     {ok, StateName, State}.
 
-should_handoff(#state{index=Idx, mod=Mod} = S) ->
-    {ok, Ring} = impure(S, riak_core_ring_manager, get_my_ring, []),
-    Me = impure(S, erlang, node, []),
+should_handoff(State=#state{index=Idx, mod=Mod}) ->
+    {ok, Ring} = impure(State, riak_core_ring_manager, get_my_ring, []),
+    Me = impure(State, erlang, node, []),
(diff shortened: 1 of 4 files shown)


src/riak_core_vnode.erl
```diff
@@ -244,3 +269,22 @@ reply({raw, Ref, From}, Reply) ->
...
+    ?VNODE_REQ{request=Request}.
+
+%% Impure handling stuff
+
+imp_interp(Fun, Arg) when not is_tuple(Fun), is_function(Fun, 1) ->
+    Fun(Arg);
+imp_interp(Else, _) ->
+    Else.
+
+impure(#state{pure_p = false}, {mod, Mod}, Func, ArgList) ->
```
argv0, line 282: Suggest renaming this to something like impure_mod_call(M, F, A, State) for better readability.  Maybe make argument order  (M, F, A, State) ?

argv0:
Love it, modulo readability concern in riak_core_vnode.

slfritchie:
Per Skype call with Jon, we're going to go with the bugfix only,
and will revisit the unit test for regression.  Jon's got a variation
of the functional purity mechanism that he'd like to try.

slfritchie:
Merged tag bz520-bugfix-for-0.14-branch into master,
closing pull request.

PR #17, opened 2011-02-16
Riak core config module

kellymclaughlin:
A module to provide uniform access to riak_core env data internally and to other modules.

### DIFF
--- a/src/riak_core_config.erl
+++ b/src/riak_core_config.erl
@@ -43,23 +43,25 @@
 %% @doc Get the HTTP IP address and port number environment variables.
 http_ip_and_port() ->
     case get_riak_core_env(http) of
-        {ok, [{WebIp, WebPort} | _]} ->            
+        [{WebIp, WebPort} | _] ->            
             {WebIp, WebPort};
-        {ok, []} ->
+        [] ->
             error;
         undefined ->
             %% Fallback to pre-0.14 HTTP config.                                                                                                                        
             %% TODO: Remove in 0.16                                                                                                                                     
-            WebIpCfg = get_riak_core_env(web_ip),
-            WebPortCfg = get_riak_core_env(web_port),
-            case {WebIpCfg, WebPortCfg} of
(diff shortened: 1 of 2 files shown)


src/riak_core_config.erl
```diff
@@ -0,0 +1,184 @@
+ %% -------------------------------------------------------------------
+%%
+%% riak_core: Core Riak Application
+%%
+%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
```
jonmeredith: It's 2011, but maybe we should fix that everywhere as a separate ticket :)

src/riak_core_config.erl
```diff
@@ -0,0 +1,184 @@
...
+ring_state_dir_test_case() ->
+    ?assertEqual("data/ring", ring_state_dir()).
+
+non_existent_var_test_case() ->
+    ?assertEqual(undefined, get_riak_core_env(bogus)).
+
+setup() ->
+    
+    %% Start the applications required for riak_core to start 
+    application:start(sasl),
```
jonmeredith: you might be able to simplify these tests by just issuing an application:load(riak_core), I don't see anything that requires starting the app or any dependencies.

src/riak_core_config.erl
```diff
@@ -0,0 +1,184 @@
...
+            %% TODO: Remove in 0.16                                                                                                                                     
+            WebIp = get_riak_core_env(web_ip),
+            WebPort = get_riak_core_env(web_port),
+            if
+                WebIp == undefined ->
+                    error;
+                WebPort == undefined ->
+                    error;
+                true ->
+                    error_logger:info_msg("Found HTTP config for riak_jmx using pre-0.14 config "
```
jonmeredith: HTTP config is for webmachine too - this message should only reference riak_core.

jonmeredith:
+1 looks good to me.

PR #18, opened 2011-02-21
merkerl: fixes potential clobbering when turning a leaf into an inner

andrewoswald:
When adding several values to a tree, I occasionally noticed that some values didn't get picked up.  After looking at the clause for inserting an MI into a "leaf" (line #158), the code that inserted the leaves into an orddict via their offset_key/2 results looked suspicious.  Indeed, if those two results were identical, the orddict values would clobber.  The solution is to simply create the inner using _only_ the initial leaf and then to subsequently call the mi_insert1 function providing the newly created inner as its Tree arg.

I added a test that demonstrates what would otherwise be a clobbered value (assuming the offset is either 0 or 8).  Its keys, <<"riak.com42">> and <<"riak.com452">>, respectively "sha" to:
<<21,236,177,213,57,7,41,130,41,112,195,213,10,85,84,31,115,84,156,40>>
<<21,236,167,160,142,11,39,35,69,14,183,49,144,41,75,9,17,141,189,202>>

Interested in your thoughts.

thanks!
-andy

### DIFF
--- a/src/merkerl.erl
+++ b/src/merkerl.erl
@@ -156,12 +156,10 @@ mi_insert({Offset,MI},Tree) ->
 		Key -> % replacing!
 		    mkleaf(MI);
 		_ -> % turning a leaf into an inner
-		    K0 = orddict:new(),
-		    K1 = orddict:store(offset_key(Offset,Key),
-				       mkleaf(MI),K0),
-		    TKey = Tree#merk.key,
-		    Kids = orddict:store(offset_key(Offset,TKey),Tree,K1),
-		    mkinner(Offset,Kids)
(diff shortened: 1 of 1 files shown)


jonmeredith:
Hi Andrew, I'm sorry this has lingered for a while - I've updated our task tracking system to make sure we look at it sooner rather than later.

andrewoswald:
No worries Jon.  I should have also submitted a proper ticket for tracking purposes.  Thanks again!

argv0:
Looks good and tests pass - verified fix with Justin Sheehy (original author).

andrewoswald:
Thanks Andy.  Hope all is well!

PR #19, opened 2011-02-24
Changes to support purifying KV FSMs

jonmeredith:
I've spent some time refactoring the get FSM today in preparation for changes and thought I'd share progress to see if we think this is a viable approach.  I've created branches in riak_core and riak_kv called jdm-purify-fsms and have pushed my latest and greatest including some code I forgot to commit end of Wednesday.

Here's the broad strokes of what I've done:
- split the initialized state into prepare and execute.  Any side effects are isolated to the prepare function.
- added an alternate start function that takes a proplist to fudge the state and drops into the execute state.
- refactored the vnode response handling into a single function
- refactored the reply-to-client decision into a function
- replaced the preference list code with the riak_core_vnode_master based commands and added an alternate variant of the preflist code that provides the vnode type of primary/fallback (needed for deletes, but will be vital for adding RP-responses-from-primaries condition).
- taught the vnode_master commands to recognize Pids in the preference list and send messages there directly instead of going through a vnode master.
- taught riak_kv_vnode:put to accept bucket properties in options so it didn't try and get the ring while in 'pure' mode
- made read repair supply bucket properties.
- cleaned up the strange spawning/individual sync deletes - I don't believe they provide any additional protection against races.  We need to add conditional matching on the vclock soon anyway.

The refactored code still passes all the original quickcheck tests and I've added a couple of examples of how it can be used in 'pure' mode without any other parts of the system spun up.

./rebar eunit app=riak_kv suite=riak_kv_get_fsm if you'd like to see it run.

riak_kv_put_fsm is causing problems with running the whole test suite, I'll clean that up when I refactor the put FSM.

Looking for comments on
- if this seems like a good approach
- reasons why any of the changes are bad or breaks something
- if the delete change is acceptable
- if the readrepair change is acceptable, and if we think it is ok in production or if it needs to be conditional for testing.  It will mean more data sent per read repair.

### DIFF
--- a/src/riak_core_apl.erl
+++ b/src/riak_core_apl.erl
@@ -24,7 +24,7 @@
 %% -------------------------------------------------------------------
 -module(riak_core_apl).
 -export([active_owners/1, active_owners/2,
-         get_apl/3, get_apl/4, get_apl2/4,
+         get_apl/3, get_apl/4, get_apl_ann/4,
          get_primary_apl/3, get_primary_apl/4
         ]).
 
@@ -64,13 +64,13 @@ get_apl(DocIdx, N, Service) ->
 -spec get_apl(binary(), n_val(), ring(), [node()]) -> preflist().
 get_apl(DocIdx, N, Ring, UpNodes) ->
     [{Partition, Node} || {{Partition, Node}, _Type} <- 
-                              get_apl2(DocIdx, N, Ring, UpNodes)].
+                              get_apl_ann(DocIdx, N, Ring, UpNodes)].
 
 %% Get the active preflist taking account of which nodes are up
 %% for a given ring/upnodes list and annotate each node with type of
 %% primary/fallback
--spec get_apl2(binary(), n_val(), ring(), [node()]) -> preflist2().
-get_apl2(DocIdx, N, Ring, UpNodes) ->
+-spec get_apl_ann(binary(), n_val(), ring(), [node()]) -> preflist2().
+get_apl_ann(DocIdx, N, Ring, UpNodes) ->
     UpNodes1 = ordsets:from_list(UpNodes),
     Preflist = riak_core_ring:preflist(DocIdx, Ring),
     {Primaries, Fallbacks} = lists:split(N, Preflist),
--- a/src/riak_core_apl.erl
+++ b/src/riak_core_apl.erl
@@ -24,7 +24,7 @@
 %% -------------------------------------------------------------------
 -module(riak_core_apl).
 -export([active_owners/1, active_owners/2,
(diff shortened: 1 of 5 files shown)


jonmeredith:
Matching K/V pull request here - https://github.com/basho/riak_kv/pull/41

kellymclaughlin:
From riak_kv if I checkout the jdm-purify-fsms branch and change rebar.config to pull the same branch from riak_core I get the following error running make: 

==> riak_core (compile)
src/riak_core_tracer.erl:22: can't find include lib "riak_kv/include/riak_kv_vnode.hrl"
make: **\* [compile] Error 1

I may have overlooked something, but I don't see that riak_core_tracer references anything form riak_kv_vnode.hrl so maybe that include is not needed? 

Commenting out that include_lib in riak_core_tracer.erl does give a successful build, but I haven't run tests yet.

slfritchie:
Kelly, that's a symptom of a short-term hack.  If you build with riak_kv
or riak_core as your top level, rebar doesn't assemble an include path that
a riak_core-based module can find a riak_kv header file.  But if you build
with riak as the top level, then rebar includes both include dirs when
building both.

jonmeredith:
the riak_kv header shouldn't be there - lack of review when I repurposed it - thanks Kelly.

kellymclaughlin:
Ok, that stinks. My question would still be if that include_lib is even necessary. All riak_kv and riak_core eunit tests pass for me with it commented out.

kellymclaughlin:
Jon beat me to it.

jonmeredith:
Cleaned up the k/v dependency stuff - should go in as riak_kv_tracer if we want it.

src/riak_core_apl.erl
```diff
@@ -62,6 +63,14 @@ get_apl(DocIdx, N, Service) ->
...
 -spec get_apl(binary(), n_val(), ring(), [node()]) -> preflist().
 get_apl(DocIdx, N, Ring, UpNodes) ->
+    [{Partition, Node} || {{Partition, Node}, _Type} <- 
+                              get_apl2(DocIdx, N, Ring, UpNodes)].
+
+%% Get the active preflist taking account of which nodes are up
+%% for a given ring/upnodes list and annotate each node with type of
+%% primary/fallback
+-spec get_apl2(binary(), n_val(), ring(), [node()]) -> preflist2().
+get_apl2(DocIdx, N, Ring, UpNodes) ->
```
beerriot: Could 'get_apl2' be renamed to something like 'get_apl_ann' (for "ANNotated"), to provide a little more context for the difference between it and 'get_apl'?

PR #21, opened 2011-02-25
Slf otp app restart fixes

slfritchie:
Small fixes to permit "./rebar eunit" to work from riak repo.

### DIFF
--- a/src/riak_core_ring_handler.erl
+++ b/src/riak_core_ring_handler.erl
@@ -73,6 +73,11 @@ ensure_vnodes_started([H|T], Ring, Acc) ->
 
 ensure_vnodes_started(Mod, Ring) ->
     Startable = startable_vnodes(Mod, Ring),
+    %% NOTE: This following is a hack.  There's a basic
+    %%       dependency/race between riak_core (want to start vnodes
+    %%       right away to trigger possible handoffs) and riak_kv
+    %%       (needed to support those vnodes).  The hack does not fix
+    %%       that dependency: internal techdebt todo list #A7 does.
     spawn_link(fun() ->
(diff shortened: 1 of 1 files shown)


jonmeredith:
Tried doing evil things to it and managed to do no harm.

(dev1@127.0.0.1)8> F = fun(Ring, Seq) -> {new_ring, riak_core_ring:update_meta(erlang:phash2(self()), Seq, Ring)} end.
(dev1@127.0.0.1)15> spawn(fun() -> [riak_core_ring_manager:ring_trans(F, X) || X <- lists:seq(1, 1000)] end).
<0.1486.1>
(dev1@127.0.0.1)16> spawn(fun() -> [riak_core_ring_manager:ring_trans(F, X) || X <- lists:seq(1, 1000)] end).
<0.1576.1>
(dev1@127.0.0.1)17> spawn(fun() -> [riak_core_ring_manager:ring_trans(F, X) || X <- lists:seq(1, 1000)] end).
<0.1664.1>

_However_ this breaks layers separation between riak_core and riak_kv and will not help with search start up issues - as search starts vnodes too.  Search will be no worse as a result, as I think riak_kv is a dependency so it starts after it anyway.

It's significant work to fix vnode startup correctly.  When riak_core/kv got split we created a race condition on vnode startup - if riak_kv manages to start and hook the ring before core loads the ringfile we start all vnodes and catch any fallback nodes.  If core loads the ring and sets it then only owned vnodes are started.  The item is on the techdebt list as A7 which should hopefully be resolved in the next few weeks.

I think the fix is acceptable if we add a big fat comment block saying that we're breaking encapsulation, why we're doing it and that it should be resolved in the near future.

PR #23, opened 2011-03-01
Bz218 ring file cleanup on leave

kellymclaughlin:
These changes fix the problem of a node not cleaning up its ring file properly when it leaves a cluster. 

The changes also fix what appears to be a race condition where a node could stop before properly distributing an updated ring file to the other nodes in the cluster. The race condition seems unlikely ever to happen in practice, but good to clean it up anyway.

### DIFF
--- a/src/riak_core_gossip.erl
+++ b/src/riak_core_gossip.erl
@@ -201,7 +201,7 @@ remove_from_cluster(ExitingNode) ->
     distribute_ring(ExitRing),
 
     % Instruct the exiting node to clean up and leave the ring
-    rpc:call(ExitingNode, riak_core_ring_manager, leave_the_ring, [ExitRing]).
+    rpc:call(ExitingNode, riak_core_ring_manager, leave_the_ring, [ExitRing, Ring]).
 
 
 attempt_simple_transfer(Ring, Owners, ExitingNode) ->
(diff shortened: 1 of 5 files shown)


slfritchie:
If I create a 3-node cluster (2 is probably sufficient, though), wait for
the rings to settle and partition transfers to finish, then put some keys:

```
{ok, C} = riak:local_client().
[C:put(riak_object:new(<<"b">>,list_to_binary(integer_to_list(X)),<<"val">>)) || X <- lists:seq(1,500)].
```

Then tell node #3 to leave.  I expect to see the following in the error
log:

```
=INFO REPORT==== 1-Mar-2011::18:19:13 ===
Starting handoff of partition riak_kv_vnode 1347321821914426127719021955160323408745312813056 to 'dev2@127.0.0.1'
=INFO REPORT==== 1-Mar-2011::18:19:13 ===
Handoff of partition riak_kv_vnode 1347321821914426127719021955160323408745312813056 to 'dev2@127.0.0.1' completed: sent 30 objects in 0.02 seconds
[... repeat for many other partitions ...]
=INFO REPORT==== 1-Mar-2011::18:19:13 ===
"node removal completed, exiting."
```

When patched with this branch, I see only the final message.  It's a good
thing because the leave op hasn't always shut down the leaving node,
and this branch appears to fix that problem.  But shutting down before
partition transfer can start & finish isn't so good.

FWIW, the basho_expect systest_cluster_membership test passes,
though it doesn't explicitly check for successful handoffs by a leaving
node.

kellymclaughlin:
These two commits, the last one in particular, should address the problem. I was trying to make the solution more complicated than it needed to be so these latest changes make the overall fix simpler and functional.

slfritchie:
Kelly: merge it.

PR #26, opened 2011-03-03
Bz1032 stabilize vclock pruning

eriksoe:
Cf. https://issues.basho.com/show_bug.cgi?id=1032


jonmeredith:
Thanks for the patch - I've added a small test and pulled it in to mainline

PR #27, opened 2011-03-08
Td p16 improve supervision and linking

kellymclaughlin:
Change a call to spawn to spawn_link in riak_core_vnode_master.

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -176,9 +176,18 @@ handle_sync_event({handoff_data,BinObj}, _From, StateName,
              State#state.inactivity_timeout}
     end.
 
+handle_info({'EXIT', _Pid, _Reason}, StateName, StateData) ->
+    %% A linked processes has died so the vnode
+    %% process should terminate and be respawned
+    %% by riak_core_vnode_master to prevent 
+    %% messages from stacking up on the process message
+    %% queue and never being processed.
+    {stop, linked_process_death, StateData};
 handle_info(_Info, StateName, State) ->
     {next_state, StateName, State, State#state.inactivity_timeout}.
 
+terminate(linked_process_death, _StateName, _State) ->
+    crash;
 terminate(Reason, _StateName, #state{mod=Mod, modstate=ModState}) ->
     Mod:terminate(Reason, ModState),
     ok.
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -46,7 +46,8 @@ behaviour_info(callbacks) ->
      {encode_handoff_item,2},
      {is_empty,1},
      {terminate,2},
-     {delete,1}];
+     {delete,1},
+     {handle_exit,3}];
 behaviour_info(_Other) ->
     undefined.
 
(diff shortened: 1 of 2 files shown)


src/riak_core_vnode.erl
```diff
@@ -176,9 +177,16 @@ handle_sync_event({handoff_data,BinObj}, _From, StateName,
              State#state.inactivity_timeout}
     end.
 
+handle_info({'EXIT', Pid, Reason}, StateName, State=#state{mod=Mod}) ->
+    %% A linked processes has died so use the
+    %% handle_exit callback to allow the vnode 
+    %% process to take appropriate action.
+    Mod:handle_exit({Pid, Reason}, StateName, State);
```
jonmeredith: We should wrap this with a catch and check if the function is defined or not - don't want to break any third parties using riak_core for their projects.

src/riak_core_vnode.erl
```diff
@@ -176,9 +177,16 @@ handle_sync_event({handoff_data,BinObj}, _From, StateName,
...
 handle_info(_Info, StateName, State) ->
     {next_state, StateName, State, State#state.inactivity_timeout}.
 
+terminate(process_crash, _StateName, _State) ->
```
jonmeredith: Why don't you want to call terminate in this case?

kellymclaughlin: Oddly, that won't cause the process to crash (at least for some backends) and leaves messages dangling on the vnode process queue. I'm looking more into this to see if it's the case for all backends or a case of inconsistent behavior.

jonmeredith: Wasn't very clear - I meant why not call Mod:terminate/2 if the process returns process_crash?

src/riak_core_vnode.erl
```diff
@@ -46,7 +46,8 @@ behaviour_info(callbacks) ->
      {encode_handoff_item,2},
      {is_empty,1},
      {terminate,2},
-     {delete,1}];
+     {delete,1},
+     {handle_exit,3}];
```
jonmeredith: this will make older implementations complain and might break the riak_search build if it is failing on warnings.
it would be better as a comment for anybody who looks as it is optional at the moment.

src/riak_core_vnode.erl
```diff
@@ -176,6 +177,18 @@ handle_sync_event({handoff_data,BinObj}, _From, StateName,
...
 
+handle_info({'EXIT', Pid, Reason}, StateName, State=#state{mod=Mod}) ->
+    %% A linked processes has died so use the
+    %% handle_exit callback to allow the vnode 
+    %% process to take appropriate action.
+    %% If the function is not implemented default
+    %% to crashing the process.
+    try
+        Mod:handle_exit({Pid, Reason}, StateName, State)
+    catch
```
jonmeredith, line 204: I think it's worth specifying just the undef case here and letting it crash.  Otherwise people won't be able to see why it failed.

Something like

<pre>
_:{undef, [{Mod,handle_exit,_}|_]}
</pre>

jonmeredith:
+1 looks good for merge.

jonmeredith:
Must have been rebased before merge.  All commits are on master.

PR #29, opened 2011-03-17
Bz977 merge diff ts

jonmeredith:
Support for bz://977.

Exposes a variant of vclock:increment that can be given an explicit timestamp so that all vnodes create the same vclock.


slfritchie:
Nice to see that having a type with the same name as an arity-zero function
doesn't confuse the compiler.  +1 to merge

PR #30, opened 2011-03-28
Fix support for multiple vnode types running on riak_core (bz://1052).

argv0:
Fix registered name in riak_core_ring_handler:ensure_vnodes_started to
be unique per-vnode module instead of a constant.

riak_core:register_vnode_module() now stores the application name that
owns the module - wait_for_kv_app changed to wait_for_app that takes
an AppName argument.

This was affecting Search and some other experimental riak_core based apps.

### DIFF
--- a/src/riak_core.erl
+++ b/src/riak_core.erl
@@ -39,11 +39,27 @@ vnode_modules() ->
     end.
 
 register_vnode_module(VNodeMod) when is_atom(VNodeMod)  ->
+    {ok, App} = case application:get_application(self()) of
+        {ok, AppName} -> {ok, AppName};
+        undefined -> app_for_module(VNodeMod)                        
+    end,
     case application:get_env(riak_core, vnode_modules) of
         undefined ->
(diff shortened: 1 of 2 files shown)


argv0:
To verify with search:

Check out search master, run 'make rel', start up a search console.

1)  Notice no merge_index initialization messages
2)  Run riak_core_vnode_master:all_nodes(riak_kv_vnode), note the list of pids
3)  Run riak_core_vnode_master:all_nodes(riak_search_vnode), note empty list

Switch branches to this one and repeat the above.

jonmeredith:
+1 for merge.  Checked riak and riak_search.

In the future we could consider a riak_core:register_vnode_module/2 that takes an app and module name so we could register multiple vnode types per application.  I don't have an immediate use case in mind so I don't think it's worth retesting the pull request - the only possibility I can think of is we needed multiple vnode types during a migration (such as changing data partitioning).

jonmeredith:
Braindead comment on my part - would work with other vnode types.

PR #32, opened 2011-03-28
Remove unused exclusion list from riak_core_vnode_master state (bz://1055

argv0:
The exclusions list is a list of vnodes that have completed handoff and should
not be restarted by the code in riak_core_ring_handler. This list used to be
stored in the vnode master, but is now kept in riak_core_handoff_manager.

This pull request removes the list from riak_core_vnode_master's state record.


jonmeredith:
+1 to merge.  Not required there any more and doesn't break anything.

PR #35, opened 2011-03-31
fix edoc errors and the "make docs" target

argv0:
Fixed a few minor edoc errors and a stale "make docs" target.


beerriot:
+1, works for me

PR #36, opened 2011-03-31
"Guarded" gen_event handlers

argv0:
This pull request adds a supervisor for "event handler guards" (riak_core_eventhandler_guard) -  gen_servers that call gen_event:add_sup_handler in their init/1 functions and exit when they receive gen_event_EXIT messages - the supervisor restarts them and the event handler is re-added. 

This was originally motivated by bz://698 (riak_core_ring_handler crashing and not being restarted) and Scott's work in riak_sysmon and riak_err.

Hairy bits:
- riak_core_node_watcher adds a "supervised callback", which is a gen_event handler parameterized with a function of arity one that receives the event.  node_watcher previously passed in a closure over its own current pid and handled the gen_event_EXIT message itself.  With a guarded handler, there is no longer a link between the node_watcher and the handler (since it's installed by riak_core_eventhandler_guard).   In trying to make node_watcher use the guarded handlers I did the following:
- changed the handler id for callbacks from a ref() to a unique identifier for the function (via erlang:fun_info) 
- changed the callback to call using a registered name rather than a Pid. 

Now, when node_watcher tries to add a handler on startup, the ID is the same so the old handler stays installed, and since it's using a registered name that's fine - no stale pid in the callback function.

I'm happy to revert node_watcher to use a regular supervised handler and manage the gen_event_EXIT signal itself if this is deemed unacceptable.


src/riak_core_eventhandler_guard.erl
```diff
@@ -0,0 +1,64 @@
...
+    ok = gen_event:add_sup_handler(HandlerMod, Handler, Args),
+    {ok, #state{handlermod=HandlerMod, handler=Handler, exitfun=ExitFun}}.
+
+handle_call(_Request, _From, State) -> {reply, ok, State}.
+
+handle_cast(_Msg, State) -> {noreply, State}.
+
+
+handle_info({gen_event_EXIT, _Handler, shutdown}, State) ->
+    {stop, shutdown, State};
```
jonmeredith: Shouldn't this be normal rather than shutdown?  See the comment in the eventhandler guard supervisor below.

src/riak_core_eventhandler_guard.erl
```diff
@@ -0,0 +1,64 @@
...
+handle_info({gen_event_EXIT, _Handler, normal}, State) ->
+    {stop, normal, State};
+handle_info({gen_event_EXIT, _Handler, _Reason}, State=#state{exitfun=undefined}) ->
```
jonmeredith: Should the exit reason be 'Reason'?

src/riak_core_eventhandler_sup.erl
```diff
@@ -0,0 +1,50 @@
...
+start_guarded_handler(HandlerMod, Handler, Args, ExitFun) ->
+    case supervisor:start_child(?MODULE, handler_spec(HandlerMod, Handler, Args, ExitFun)) of
+        {ok, _Pid} -> ok;
+        Other -> Other
+    end.
+
+handler_spec(HandlerMod, Handler, Args, ExitFun) ->
+    {{HandlerMod, Handler},
+     {riak_core_eventhandler_guard, start_link, [HandlerMod, Handler, Args, ExitFun]},
+     permanent, 5000, worker, [riak_core_eventhandler_guard]}.
```
jonmeredith: I think transient would be better here if you make the changes above - that way if the handler is removed with delete_handler then the supervisor will go away too.  If it crashes it will be restarted.

jonmeredith:
Why did you change from make_ref() to riak_core_util:fun_uniq_id(Fn) for the callback arguments?  Was the make_ref() version broken in the new world?  Are you deliberately trying to avoid the same back being supervised multiple times?

I can't see any code inside the repo that actually calls the callback variants.  We could always add a variant of add_guarded_callback that takes a callback and id and defaults to the fun_uniq_id function if not provided, that way you can have your cake and eat it.

jonmeredith:
My mistake ... looks like nodewatcher uses the callbacks.

argv0:
implemented suggestions.

jonmeredith:
Re-reviewing this I think we're better off leaving node_watcher out of the new event watcher mechanism (and commenting to explain why).

I think there's a race with registering the ring callback and there are problems if the node watcher is killed with an exit/kill.  The supervisor will see it go down and because it is transient restart it with the old dead Pid.  node watcher will come up and register a new handler.  I think it would still work with the registered name plan as long as it gracefully handles trying to start up another child with the same id.

(dev1@127.0.0.1)2> whereis(riak_core_node_watcher).
<0.107.0>
(dev1@127.0.0.1)6> gen_event:which_handlers(riak_core_ring_events).
[riak_repl_ring_handler,riak_core_ring_handler,
 {riak_core_ring_events,<0.107.0>}](dev1@127.0.0.1)7> 
(dev1@127.0.0.1)7> exit(pid(0,107,0),kill).
true
(dev1@127.0.0.1)8> gen_event:which_handlers(riak_core_ring_events).
[{riak_core_ring_events,<0.2288.0>},
 riak_repl_ring_handler,riak_core_ring_handler,
 {riak_core_ring_events,<0.107.0>}]

argv0:
reverted to old handler usage.  if it ain't broke, don't fix it.

jonmeredith:
+1 on merge.

Checked killing node watcher re-registers

<pre>
(dev1@127.0.0.1)4> gen_event:which_handlers(riak_core_ring_events).                      
[riak_repl_ring_handler,riak_core_ring_handler,
 {riak_core_ring_events,#Ref<0.0.0.143>}]
(dev1@127.0.0.1)5> whereis(riak_core_node_watcher).
<0.107.0>
(dev1@127.0.0.1)6> exit(pid(0,107,0),kill).
true
(dev1@127.0.0.1)8> gen_event:which_handlers(riak_core_ring_events).
[{riak_core_ring_events,#Ref<0.0.0.2212>},
 riak_repl_ring_handler,riak_core_ring_handler]
</pre>


and the handlers survive bad rings

<pre>
(dev1@127.0.0.1)9> riak_core_ring_events:ring_update(not_a_ring).
(dev1@127.0.0.1)10> gen_event:which_handlers(riak_core_ring_events).
[riak_core_ring_handler,riak_repl_ring_handler,
 {riak_core_ring_events,#Ref<0.0.0.2501>}]
</pre>

PR #37, opened 2011-04-03
Remove "riak" prefix requirement for vnodes

rzezeski:
I see no reason to force a "riak_" prefix on all vnodes.  My guess is
this was just leftover from the original separation from riak_kv.

Without this patch one is forced to prefix his/her vnode module w/ `riak_`.


Vagabond:
Fixed differently by https://github.com/basho/riak_core/pull/42

PR #41, opened 2011-04-05
bz://716 : graceful riak_core_handoff_sender failure on crash of receiving node

argv0:
Added a clause in riak_core_vnode for the handoff sender to notify the vnode of a failed handoff attempt, which calls Mod:handoff_cancelled.

Added a clause for handle_info('EXIT', ...) for the handoff sender pid to avoid crashing the vnode.

Added error checking of tcp calls in fold function to avoid hard crash.


jonmeredith:
Tested and verified I got the new error.

Nice work catching handoff_error and retrying.
I wonder if we could have done try/catch around the other send/recv code to keep it simpler.

Vagabond:
Merged

PR #42, opened 2011-04-05
az244: More sane calculation of the owner application for a vnode module.

argv0:
Addresses same issue as https://github.com/basho/riak_core/pull/37
but enforces no convention on the vnode module name and uses the
canonical source for registered vnode modules
((riak_core, vnode_modules) in the app env).

Also gets rid of funky atom construction.


rzezeski:
+1 - I prefer this over mine, didn't even notice the vnode_modules env parameter.

Vagabond:
+1 for merge.

PR #43, opened 2011-04-07
Patches/0.14

cipriancraciun:
Fixed `riak_core_vnode:should_handoff` to accept VNode modules not prefixed with `riak`.

(There still exists the limitation that the name of a VNode module must have exactly
three components joined by an underscore, and the last one must be `_vnode`.)


seancribbs:
This is already addressed by https://github.com/basho/riak_core/pull/42 in a cleaner fashion.

PR #44, opened 2011-04-08
Cleanup and performance

eriksoe:
Minor code simplifications; non-aggressive performance-improving changes.

### DIFF
--- a/src/chash.erl
+++ b/src/chash.erl
@@ -101,8 +101,14 @@ successors(Index, CHash) ->
 %                                                             -> [NodeEntry]
 successors(Index, CHash, N) ->
     Num = max_n(N, CHash),
-    {Res, _} = lists:split(Num, ordered_from(Index, CHash)),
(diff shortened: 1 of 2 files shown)


Vagabond:
Some of my benchmarking results on this change:

For a basho_bench run with 20 concurrent workers, I saw master peak at about 390% CPU on a quad-core i7 (with hyperthreading, so there's 8 CPUs). With the modifications I didn't see anything over 260% cpu usage. After running the benchmark twice the total CPU time for the beam processes were 57.38 and 45:46. respectively.

I also generated basho_bench graphs, but the graphs were basically identical, indicating that this won't improve throughput unless CPU time becomes limited. I'm going to rebench with -smp disabled and see what happens.

Vagabond:
Master
Trifork

This was with SMP disabled and with 100 concurrent workers.

Vagabond:
Merged.

PR #45, opened 2011-04-08
vclock: Make merge stable

eriksoe:
Use max of timestamps in the case where the counters are identical - cf bz://1061.


jonmeredith:
Thanks for the pull req, I think we got inspired by it and made a very similar change which avoided constructing the two tuples in the recursive call to merge.

eriksoe:
That in turn inspired me... see comments on https://github.com/basho/riak_core/commit/563bdcc73bcfc1b47eef3468492c9b4b805d024f .

PR #46, opened 2011-04-13
handle_exit

DangerDawson:
I want to use handle_exit in my riak_core based project although the current implementation expects you to know the internals of a State, when we should be passed back the ModState. I have modified this as best I could although it breaks the latest riak_kv as it expects a handle_exit/4 where as I have not re-implented it as a handle_exit/3 which I feels it should be.

If this is not the correct way then I would be interested to know how to spawn many process in a vnode with a process link and handle the exit's in a graceful way without taking down the vnode as well?

Dave

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -39,6 +39,7 @@
 behaviour_info(callbacks) ->
     [{init,1},
      {handle_command,3},
+     {handle_exit,3},
      {handoff_starting,2},
      {handoff_cancelled,1},
      {handoff_finished,2},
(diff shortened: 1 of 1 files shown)


beerriot:
I have TODOs on an unrelease project to do something very much like this.  I think that in addition to {noreply, ...}, the case clause should also check for {stop, Reason, NewModState}, so it can re-wrap as {stop, Reason, State#state{modstate=NewModState}}.

Other TODOs nearby suggest also adding a handle_info({'DOWN', …}…) clause, which would call Mod:handle_down(…) in a similar manner.  That would make monitoring processes from a vnode possible, instead of linking.

dreverri:
Regarding the riak_kv_vnode:handle_exit/4 function, it doesn't seem to do much so it's probably not worth worrying about breaking it:
https://github.com/basho/riak_kv/blob/master/src/riak_kv_vnode.erl

dreverri:
The comment regarding handle_exit at the top of file should be updated to reflect the change in functionality:
https://github.com/DangerDawson/riak_core/blob/c5342b8be632c4756d529944168496722ea6b7e1/src/riak_core_vnode.erl#L55

DangerDawson:
I will update the comments, and then let you know when I am ready

PR #47, opened 2011-04-13
Az276 fix handle exit

argv0:
properly pass the vnode module state, not the core_vnode state to handle_exit.

added QC test to core_vnode_eqc to verify handle_exit behavior.


slfritchie:
Looks good to merge, thanks.

PR #50, opened 2011-04-22
AZ 307: SSL encryption of riak_core handoff traffic

slfritchie:
Added basics for SSL options for handoff sender & receiver.  It has been
lightly tested using SSL (it works), and the non-SSL case still appears to
work fine.  Because basho_expect can't easily patch stuff that isn't a .beam
file, e.g. etc config files, boot scripts, I haven't given this a full workout
with basho_expect yet.

Verification of SSL encryption: use Wireshark to record traffic on ports
8101 and 8102 directly, or use "tcpdump -w /path/to/file -i lo0 -s 9999
port 8101 or port 8102" and then use Wireshark to analyze as SSL.

### DIFF
--- a/src/riak_core_handoff_listener.erl
+++ b/src/riak_core_handoff_listener.erl
@@ -28,14 +28,18 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 -export([sock_opts/0, new_connection/2]).
--record(state, {portnum}).
+-record(state, {
+          portnum :: integer(),
+          ssl_opts :: list()
+         }).
 
 start_link() ->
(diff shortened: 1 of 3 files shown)


slfritchie:
Oops, I forgot a bit more commentary & description on how to use.

See https://gist.github.com/f423cd526d54a1f50d01 for instructions on
generating the key files that you need to test this patch.  Strictly speaking,
you only need one of them, the `erlserver.pem` file.  After copying it to
its etc directory place, edit the `riak_core` app's `handoff_ssl_options`
option in `etc/app.config` to point to its full path.

If I've left out other important details, just holler at me.

Vagabond:
+1 for merge.

PR #53, opened 2011-05-04
Fixed spiraltime module

goj:
- old code used to crash on `incr(1, N - 60, fresh(N))`
- improved efficency and code style
- added better tests


russelldb:
Ooops, closed without comment by mistake.

Meant to say: thanks for the fix, sorry for the delay, from Riak 1.2 on we use folsom for stats, which has a different impl of spiraltime.

PR #56, opened 2011-05-26
expose the pid for the vnode that just received the command

beerriot:
command_return_vnode returns {ok, VnodePid} instead of just 'ok', to
allow the caller to monitor the vnode that is handling its request

see https://agilezen.com/project/17206/story/386 for why this is helpful


slfritchie:
Makes a ton of sense, merge plz.

PR #58, opened 2011-06-09
Balanced choose claim

grourk:
Re-creating pull request from separate branch.

### DIFF
--- a/src/riak_core_claim.erl
+++ b/src/riak_core_claim.erl
@@ -52,9 +52,9 @@
 %% default is 3.
 
 -module(riak_core_claim).
--export([default_wants_claim/1, default_choose_claim/1,
+-export([default_wants_claim/1, default_choose_claim/1, balanced_choose_claim/1,
          never_wants_claim/1, random_choose_claim/1]).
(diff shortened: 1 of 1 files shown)


grourk:
This seems to be working in my testing, at least for adding nodes.  I'm not quite sure yet how it well it works with node leave/remove.  It could also use some cleaning up, which I hope to do soon to make the algorithm more clear.

rzezeski:
Hey Greg,

Sorry this has been dormant for so long.  A few months ago we looked at your algorithm and it indeed performs better than our current default.  In fact, based on yours we wrote one of our own that not only balances better for larger clusters, like yours, but also does so for small clusters.  We are in the process of testing our new algorithm and should it prove worthy it will become the new default.  For now I'll leave this PR open but once our new algorithm has been vetted I'll close it.  Even though we aren't directly using this patch we are highly appreciative for it.

-Ryan

rzezeski:
We just merged a new default claim to address the issues that prompted this PR.  Thanks again for the submission, Greg.

https://github.com/basho/riak_core/pull/126

PR #59, opened 2011-07-06
Az496 coverage code changes

kellymclaughlin:
This pull request generalizes the _coverage_ behavior used by key listing and makes it directly usable for other operations. All key and bucket listing done in `riak_kv` has been updated to use this code on the `az496-coverage-code` branch including the key listing eqc test. 

My benchmarking shows that there is no real change in latency of any operations as a result of this change. 

I added function to `riak_core_ring` to determine the ring index that is the successor of a `{Bucket, Key}` directly given the number of ring partitions instead of having to calculate the entire preference list. There is also a related function added to the `chash module and a quickcheck test to exercise this function. 

This pull request also includes a few commits to tidy up some of the core modules by adding some formatting, converting edoc to specs, etc.


kellymclaughlin:
This pull request is related to [this](https://github.com/basho/riak_kv/pull/126) `riak_kv` pull request and the two should be reviewed together.

rebar.config
```diff
@@ -1,6 +1,8 @@
 {erl_first_files, ["src/gen_nb_server.erl", "src/gen_server2.erl"]}.
 {cover_enabled, true}.
 {edoc_opts, [{preprocess, true}]}.
+{erl_opts, [debug_info, fail_on_warning]}.
```
beerriot: `fail_on_warning` has been deprecated in favor of `warnings_as_errors`

src/chash.erl
```diff
@@ -24,6 +24,9 @@
 %%      coincides with SHA-1 hashes, and so any two keys producing the same
 %%      SHA-1 hash are considered identical within the ring.
 %%
+%$      Warning: It is not recommended that code outside this module make use
```
beerriot: typo: `%$` instead of `%%`

src/riak_core_coverage_filter.erl
```diff
@@ -0,0 +1,144 @@
...
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc This module is used to construct a property list of VNode
+%%      indexes and a functions to filter results from a coverage
```
beerriot: "and a functions" should just be "and functions"?

src/riak_core_coverage_filter.erl
```diff
@@ -0,0 +1,144 @@
...
+        (ItemFilter == none) andalso (FilterVNodes == undefined) -> % no filtering
+            [];
+        (FilterVNodes == undefined) -> % only key filtering
+            %% Associate a key filtering function with each VNode
+            [{Index, build_filter(ItemFilter)} || {Index, _} <- VNodes];
+        (ItemFilter == none) -> % only vnode filtering required
+            {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+            PrefListFun = build_preflist_fun(Bucket, Ring),
+            %% Create VNode filters only as necessary
+            [{Index, build_filter(proplists:get_value(Index, FilterVNodes), PrefListFun)} || {Index, _} <- VNodes, proplists:is_defined(Index, FilterVNodes)];
```
beerriot: wrap to 80 column, please

src/riak_core_coverage_plan.erl
```diff
@@ -0,0 +1,296 @@
...
+%%      covering VNodes around the ring.
+-spec create_plan(bucket(), pos_integer(), req_id(), atom()) ->
+                         {error, term()} | coverage_plan().
+create_plan(Bucket, PVC, ReqId, Service) ->
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    case Bucket of
+        all ->
+            %% It sucks, but for operations involving all buckets
+            %% we have to check all vnodes because of variable n_val.
+            NVal = 1;
```
beerriot: Would the coverage be good enough if, instead of NVal=1, NVal were set to the minimum of the bucket-defaults n_val ++ the n_val of any bucket props stored in the ring?  If it worked, would it be a big enough benefit for a user that used only n_vals >1?  … for a user that used only a single n_val?

kellymclaughlin: That would probably be good thing to investigate and take some measurements, but I can see that being beneficial. I'll try to explore that a little.

beerriot:
`RINGTOP` is redefined in many places, and `?RINGTOP div Partitions` is used all over.  Maybe time to move them to a header file ± function somewhere?

beerriot:
+1 merge, very cool stuff

jonmeredith:
Here's a quick review of the things we discussed.

= Listing Buckets =

Listing buckets fails if any single node is down (as N=1), so perhaps it's better to introduce an atom of 'allup' for the coverage planner.  While you're in there you could remove the dependency on bucket properties and just pass N instead to reduce coupling.

= M/R =

Old school m/r crashes and burns when too many nodes are down.  I moved my systest code over from bitbucket https://github.com/basho/systest .  To fail set up a 4 node cluster, take 3 nodes down and run mrh:eg(). and it should crash and burn with lots of error messages.

= Architectural =

Some of the changes introduce some k/v ideas into core and we should really keep them at the appropriate level. 
- Using {Bucket, Key} instead of hash values in riak_core_ring:responsible_index.
- riak_core_coverage_filter is probably most useful at the k/v layer - the responsible_index abstraction in riak_core_ring should be enough for other vnode types.
- core_core_coverage_fsm makes references to luke and mapred.  They should be moved to the callback module and new calls added for them - init / start (not sure if you need a separate call), handle_response, finish, error.  You could make the callback module responsible for finish/error handling for maximum flexibility.

There's also some simplifications/interface cleanups you can do.  Currently the path to listkeys goes...
- riak_client:list_keys calls riak_kv_keys_fsm_sup:start_keys_fsm which calls riak_core_coverage_fsm:start_link with riak_kv_keys_fsm as the callback.  This couples the argument format to the FSM with riak_client.
- core_coverage_fsm calls vnode_master:coverage which in turn sends a ?NODE_REQ to the vnode masters in the plan  vnode master then passes on a COVERAGE_VNODE_REQ for each index (after looking up the filter
   which got sent as part of the message) 
- core_vnode passes COVERAGE_VNODE_REQ as a whole on to the vnode callback (no unwrapping).  This limits us in how we encode the vnode-level coverage requests and exposes innards of core to the callback.
- kv_vnode then does an apply on CoverageMod/CoverageFun with extra arguments of filter, Idx, Mod, ModState. The callback module is riak_kv_vnode:list_keys which calls the Mod:fold stuff, using process_keys to filter as needed.

This could be simplified.
- riak_core_coverage_fsm could send directly to each vnode (i.e. rewrite riak_core_vnode_master:coverage to send multiple casts).
- send a CORE_COVERAGE_REQ with an index, sender, request, keyfilter list
- core_vnode_master pass it onto the appropriate vnode like we currently do
- core_vnode builds a filter function based on the keylist filter and calls a new handle_coverage function in the vnode callback if present, or returns an error if not.
- handle_coverage callback gets passed 'from' which will probably be the coverage FSM, the request term and keylist filter to decide what to do.  It can call riak_core_vnode:reply to send responses back to the coverage FSM.

Also, look at using/extending riak_core_vnode:reply for sending responses, that way we can trace them rather than !ing messages around.

kellymclaughlin:
In regards to the M/R errors, that is due to existing behavior in the M/R code that predates any of the coverage or key listing changes. I think changes to that behavior are outside the intended scope of this changeset.

src/riak_core_coverage_plan.erl
```diff
@@ -0,0 +1,268 @@
...
+-type vnode_filters() :: [{node(), [{index(), [index()]}]}].
+-type coverage_plan() :: {coverage_vnodes(), vnode_filters()}.
+
+%% ===================================================================
+%% Public API
+%% ===================================================================
+
+%% @doc Create a coverage plan to distribute work to a set
+%%      covering VNodes around the ring.
+-spec create_plan(VNodeSelector :: all | allup, NVal :: pos_integer(),
```
rzezeski: Why are you listing the argument names in the spec?  For edoc?  Because edoc will automatically pull them in.  I think removing them and making types for these various args would look cleaner.

BTW, cool stuff here.

rzezeski:
I plan on using this feature to implement an experimental feature in Search.  While attempting to implement a coverage behavior I realized that it requires that the data is wrapped in one of two special tuples.

```
{results, R}
{final_results, R}
```

This means that if I want a vnode API to participate in a coverage plan it needs to make sure to wrap it's data in these tagged tuples.  This has some unfortunate effects IMO:
1. It couples the vnode to the coverage behavior implementation.
2. It unnecessarily limits what messages can be sent between the vnode and coverage behavior.  I.e. it forces you to wrap them even if you don't want to.
3. (And this is what sparked this comment) You can't utilize the vnode as it's currently written even if it already has the API you want to expose via coverage behavior.  I.e. If I want to use a coverage plan in search I have to modify the search vnode to detect when the coverage behavior is calling it's API and make sure to wrap the results in the correct tuples.  This seems silly.

IMO, a better approach would be to pass all messages directly to the `process_results` callback in the coverage behavior impl.  Along with this add a special return type that "stops" the behavior.  E.g. something like fsm's `{stop,Reason,NewStateData}`.

kellymclaughlin:
Pushed a set of changes to riak_kv and riak_core to implement suggestions by Ryan Zezeski. Thanks Ryan!

jonmeredith:
(comments cover https://github.com/basho/riak_kv/pull/126 too).

It's really nice to see all the abstraction in riak_kv_keys_fsm when you look at the whole 
pull request diff, and the bucket listing code works fine with nodes down now.

I think the improvements based on Ryan's comments are very good.  I like passing opaque messages
back and forth.  However, I don't like that waiting_results has to return Vnode.
That forces the vnode callback code to track what index it is, which all vnodes
may not want to do.  The core_vnode code should really be responsible for passing that on.

Seeing the refactor, it makes having a handle_coverage() / coverage_reply() pair of functions seems
like an even more useful abstraction as you can build in better error handling.

  handle_coverage(Req, Keyspaces, From, VnodeModState) ->
    {reply, Reply, UpdVnodeModState} | {noreply, UpdVnodeModState}

If you wanted to make multiple calls, could use

  coverage_reply(From, Reply)

Which would wrap up sucessful coverage calls with the reqid and vnode info.

If you wrap the handle_coverage call in a try/catch you can let the calling
FSM know that it died rather than wait for a timeout that never comes (the FSM is not
monitoring the vnode after all).  I think we'll want to add something like this for handle_command requests in the future too.

The pull request still needs to address rolling upgrades - clusters with 0.13.x and 0.14.x nodes.
It seems silly to remove the functionality when we merge this pull request, then add it back
in a subsequent one.  Have you mapped out how it needs to work?

The listkeys / listbuckets req is still passing a function around between nodes - you
mentioned you were a little concerned about that for rolling upgrades.

A couple of last minor things.

Still exposing the init args array in riak_client.

There's a commented out handle_cast in riak_core_vnode_master for old COVERAGE_VNODE_REQ

jonmeredith:
A few last comments.

For the core coverage, why match on From = {'raw', ReqId, _} - what if a gen_fsm/gen_server wants to use it?
Why is sender computed in the callback?  The core coverage FSM knows how it wants data sent back, not the callback modules.
It would be cleaner to treat the From passed into riak_core_coverage_fsm as opaque and create our own refs. 

I feel a bit ambivalent about the ReplyTo fun being passed into handle_coverage.  It's a local fun so it doesn't have the problems of passing them over dist erlang, but it's a different way of doing things than in handle_command, so if for example you just wanted to write a vnode module that passed on requests to the main handle_command you couldn't use the same reply function.  You could make riak_core_vnode_master:coverage update the Sender to replace ReqId with {ReqId, Vnode}.  It couples the vnode master/coverage FSM a bit, but maybe it's worth it.  That way you could just use the same reply code for normal requests.

kellymclaughlin:
Added pull requests to add the rolling upgrade switch to app.config in [riak](https://github.com/basho/riak/pull/47) and [riak_ee](https://github.com/basho/riak_ee/pull/18).

kellymclaughlin:
I dislike the idea of adding the vnode information to the Sender tuple because the vnode is the receiver. Maybe it would be better if I change the tuple prepended to the results in ReplyFun to be {index, Index}, change vnode_coverage so that returning {reply, R, S} does the same thing as in vnode_command, and change the coverage fsm to match on {index, Index}, but also accept other results. There has to be a way to know when all results have been sent for a vnode and so even if passing things to handle_command, the vnode module will have to indicate completion with the ReplyFun.

jonmeredith:
+1 merge.

PR #60, opened 2011-07-06
fixes differences between slide.erl and spiraltime.erl per bz://1124

schofield:
Chanes spiraltime.erl to use calendar:local_time like slide.erl. This fixes the one minute stats problem for gets and puts.


jtuple:
+1, merging in before 1.0.

Note: 1.0 has new stats engine that doesn't use spiraltime, but this fix is still good for legacy kv_stat as well as the new core_stat that still uses spiraltime.

PR #61, opened 2011-07-08
Cleanup and performance

eriksoe:
Improvements in vclock.

### DIFF
--- a/src/vclock.erl
+++ b/src/vclock.erl
@@ -142,19 +142,9 @@ timestamp() ->
     calendar:datetime_to_gregorian_seconds(erlang:universaltime()).
 
 % @doc Compares two VClocks for equality.
-%      Not very fast.
 -spec equal(VClockA :: vclock(), VClockB :: vclock()) -> boolean().
 equal(VA,VB) ->
-    VSet1 = sets:from_list(VA),
-    VSet2 = sets:from_list(VB),
-    case sets:size(sets:subtract(VSet1,VSet2)) > 0 of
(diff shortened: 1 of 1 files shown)


krestenkrab:
Looks clean except indentation in the prune_vclock fix

src/vclock.erl
```diff
@@ -136,25 +136,25 @@ increment(Node, IncTs, VClock) ->
...
+       ).
+
 % @doc Return a timestamp for a vector clock
 -spec timestamp() -> timestamp().
 timestamp() ->
-    calendar:datetime_to_gregorian_seconds(erlang:universaltime()).
+    %% Same as calendar:datetime_to_gregorian_seconds(erlang:universaltime()),
+    %% but significantly faster.
+    %% Assumes that less than a million now() calls are done per second.
+    {MegaSeconds, Seconds, _} = now(),
```
Vagabond: Wouldn't os:timestamp be even faster, it doesn't have to be locked to make the clock monotonic like now().

eriksoe: @Vagabond: Ah - didn't know about os:timestamp(), and wondered why it (a fast timestamp function without monotonicity guarantee and the possibility of skewedness it entails) wasn't there.
Thank you!

eriksoe: Fwiw, my measurements show that it may be a bit (20-30%) faster - measurable but not dramatical.

Vagabond:
Merged, with some cleanups.

PR #62, opened 2011-07-17
Make riak_core_vnode_sup start options configurable

jbrisbin:
The default start options for the vnode supervisor don't fit every situation. I've added code to make it configurable by adding something like:

{riak_core_vnode_sup, [
  {lifecycle_options, {
    {simple_one_for_one, 10, 10}, 
    [
      {undefined, {riak_core_vnode, start_link, []}, temporary, brutal_kill, worker, dynamic}
    ]
  }]
}

Which allows you to make vnodes fit the situation of your particular application (e.g. without brutal_kill etc...).

### DIFF
--- a/ebin/riak_core.app
+++ b/ebin/riak_core.app
@@ -3,7 +3,7 @@
 {application, riak_core,
 [
   {description, "Riak Core"},
-  {vsn, "0.14.0"},
+  {vsn, "0.14.1"},
   {modules, [
              app_helper,
              bloom,
--- a/ebin/riak_core.app
+++ b/ebin/riak_core.app
@@ -3,7 +3,7 @@
 {application, riak_core,
 [
   {description, "Riak Core"},
-  {vsn, "0.14.1"},
+  {vsn, "0.14.2"},
   {modules, [
              app_helper,
              bloom,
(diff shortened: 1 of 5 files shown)


PharkMillups:
Thanks, Jon. 

This one is added to the review queue. We'll merge/get back to you with comments ASAP. 

Keep the code coming. 

Mark

rzezeski:
Hey, Jon.

Have you used this to change anything other then the shutdown method?

jbrisbin:
I haven't personally, no.

rzezeski:
So what I'm wondering is if we really want a config for the entire child spec.  It seems just a config for the shutdown method is enough.  This could just be some undocumented thing that devs can make use of.  But I also wonder why the shutdown method is `brutal_kill` to begin with.  I wonder if this is something you'd want to be able to change per vnode type, but I don't think there is anyway to achieve that.

Anyways, I'm rambling.  Could you humor me and give me the use case where you needed this to be a timeout?

russelldb:
Hey Jon,

It's been a while. Do you still need this feature? I agree with Ryan that being able to configure the entire child spec like this is overkill. Given how much riak_core has moved on, and how small the change is, would you mind re-submitting if you still value the change?

PR #63, opened 2011-07-19
Az475 integrate lager

Vagabond:


jonmeredith:
I had problems building - riak_core expects 0.9, lager is now 1.0 so it blows up.

Other than that, +1 merge.

PR #65, opened 2011-08-03
New join/claim/membership implementation and model

jtuple:
This request includes the initial implementation of the new join/claim approach along with the corresponding EQC model used during development.

Concerning the implementation, the main thing missing is support for rolling upgrades as well as various extensions to the base approach previously discussed. Currently, the implementation has only been tested by hand. Future work will adapt the EQC model to drive a live test cluster for testing, as well as implement the extensions / rolling upgrade support.


src/riak_core.erl
```diff
@@ -52,24 +52,72 @@ join(Node) when is_atom(Node) ->
...
             end;
         pang ->
             {error, not_reachable}
     end.
 
+remove(Node) ->
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    case riak_core_ring:member_status(Ring, Node) of
+        invalid ->
+            io:format("~p isn't a member of the cluster.~n", [Node]),
```
jonmeredith: user output should be handled by the caller - riak_kv_console in this case, should return {error, not_cluster_member} or something instead.

src/riak_core.erl
```diff
@@ -52,24 +52,72 @@ join(Node) when is_atom(Node) ->
...
+            riak_core_ring_manager:set_my_ring(Ring2),
+            case riak_core_ring:random_other_node(Ring2) of
+                no_node ->
+                    ok;
+                RandomNode ->
+                    riak_core_gossip:send_ring(Node, RandomNode),
+                    ok
+            end;
+        invalid ->
+            io:format("~p isn't a member of the cluster.~n", [Node]),
```
jonmeredith: more io:format.

src/riak_core.erl
```diff
@@ -52,24 +52,72 @@ join(Node) when is_atom(Node) ->
...
+                    end;
                 _ -> 
-                    {error, different_ring_sizes}
+                    {error, unable_to_get_join_ring}
             end;
         pang ->
             {error, not_reachable}
     end.
 
+remove(Node) ->
```
jonmeredith, line 74: Maybe add comment to explain difference between remove/leave?

src/riak_core_ring_manager.erl
```diff
@@ -72,6 +73,9 @@ get_my_ring() ->
 refresh_my_ring() ->
     gen_server2:call(?MODULE, refresh_my_ring, infinity).
```
jonmeredith, line 75: What calls refresh_ring/? I can't find it anything, is it generally useful?

src/riak_core_vnode.erl
```diff
@@ -126,8 +126,24 @@ continue(State, NewModState) ->
...
     
 
-vnode_command(Sender, Request, State=#state{mod=Mod, modstate=ModState}) ->
-    case Mod:handle_command(Request, Sender, ModState) of
+vnode_command(Sender, Request, State=#state{index=Index,
+                                            mod=Mod,
+                                            modstate=ModState}) ->
+    %% Check if we should forward
+    Node = node(),
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
```
jonmeredith, line 134: Computing this per-request is too expensive - ideally the logic to decide whether a vnode should shutdown/forward should be outside of the vnode itself. This is the latency critical path.

We may also need to make similar checks with the new handle_coverage code - you may need to discuss with Kelly.

src/riak_core_ring.erl
```diff
@@ -324,22 +664,277 @@ pick_val(M1,M2) ->
     end.                   
 
 %% @private
-% @doc If two states are mutually non-descendant, merge them anyway.
-%      This can cause a bit of churn, but should converge.
-% @spec reconcile(MyNodeName :: term(),
-%                 StateA :: chstate(), StateB :: chstate())
-%              -> chstate()
-reconcile(MyNodeName, StateA, StateB) ->
```
jonmeredith, line 332: All of the reconciliation code seems a bit fiddly and there are no tests.  Some comments explaining the way in which the two divergent rings are reconciled would be helpful.

src/riak_core_ring.erl
```diff
@@ -77,14 +107,19 @@
...
 %% Public API
 %% ===================================================================
 
-%% @doc Produce a list of all nodes that own any partitions.
+%% @doc Produce a list of all nodes that are members of the cluster
 -spec all_members(State :: chstate()) -> [Node :: term()].
-all_members(State) ->
-    chash:members(State#chstate.chring).
+all_members(#chstate{members=Members}) ->
+    get_members(Members, [valid, leaving, exiting]).
```
jonmeredith: Why not call get_members/1 here instead? If new types are added you'd have to include them both places.

jonmeredith:
I've added some line comments above, here are more general ones.

It might make sense to add a riak_core_gossip:to_other_node or riak_core_gossip:trigger() that does the other node selection and gossip.  That pattern appears a lot and in some cases we don't handle no_node returns (even though they are very unlikely).  It also lays the way for changing the gossip mechanism.

I built a 24-node cluster successfully - I had to fiddle with the ring to understand progress, so we need some human-comprehensible status information.  They don't have to be done as part of this p/r, but I thought I'd write up thoughts here.

Add to status - preferably a tuple that we convert to terminal/gui output.
- Who is the claimant
- How many partition transfers need to happen before the ring is settled.
- Which nodes are active/leaving/exiting

And we could benefit from some stats 
- total number of gossip exchanges
- ignored gossip - increment counter
- number of rings in conflict that had to be resolved

We should also add some extra logging
- join cluster message
- remove cluster message
- leave cluster message

It's a big improvement on what we have.  I haven't run the eqc model yet, but and I'll test the changes you're pushing that resolve the issues I had leaving the cluster when I get a chance.

src/riak_core_ring_manager.erl
jonmeredith, line 75: derp. refresh_ring != refresh_my_ring.

jonmeredith:
More playing with my 24-node cluster (code at 3f4d6c2...).  I found two things

1) Leave isn't reliable yet.  I made a 24 node cluster, then on the claimant ran 
  rpc:multicall(OtherNodes, riak_core, leave, []).  There were 5 nodes left with status still valid afterwards

```
(dev1@127.0.0.1)81> riak_core_console:member_status().
================================= Membership ==================================
Status     Ring    Node
-------------------------------------------------------------------------------
valid     16.8%    'dev12@127.0.0.1'
valid     16.8%    'dev13@127.0.0.1'
valid     16.8%    'dev19@127.0.0.1'
valid     16.8%    'dev1@127.0.0.1'
valid     16.4%    'dev20@127.0.0.1'
valid     16.4%    'dev6@127.0.0.1'
-------------------------------------------------------------------------------
Valid:6 / Leaving:0 / Exiting:0
ok
```

Running a second time did the trick - could there be any more set_my_ring calls that should be ring_trans?

```
(dev1@127.0.0.1)82> riak_core_console:member_status().
================================= Membership ==================================
Status     Ring    Node
-------------------------------------------------------------------------------
leaving   16.4%    'dev12@127.0.0.1'
leaving   16.8%    'dev13@127.0.0.1'
leaving   16.0%    'dev19@127.0.0.1'
leaving   15.2%    'dev20@127.0.0.1'
leaving   16.4%    'dev6@127.0.0.1'
valid     19.1%    'dev1@127.0.0.1'
-------------------------------------------------------------------------------
Valid:1 / Leaving:5 / Exiting:0
```

2) The ring may not be being cleared on exit correctly, I've seen this a couple of times on restart

```
15:44:12.383 [info] Application ssl started on node 'dev21@127.0.0.1'
Eshell V5.8.4  (abort with ^G)
(dev21@127.0.0.1)1> 15:44:59.946 [notice] "node removal completed, exiting."
15:44:59.947 [info] Spidermonkey VM (pool: riak_kv_js_hook) host stopping (<0.305.0>)
```

Restarting again keeps the node up.

The good news is the data survived all my tinkerings.

jtuple:
I think I've covered every remaining item still on the list for this pull request. Let me know if there is anything else needed.

My recent commits include consolidation of the random_gossip logic into a reusable function, as well as the addition of the random_recursive_gossip and recursive_gossip functions. The recursive gossip code is fixed gossip approach that decomposed the cluster members list into a binary tree and gossips the ring to a node's left and right children. recursive_gossip gossips to a node's children, while random_recursive_gossip picks a random node and gossips to it's children.

Periodic gossip (gossip interval) continues to use the random_gossip, while recursive gossip replaces random gossip for the "gossip immediately case". Specifically, when a ring is directly changed, random_recursive_gossip is used to initiate a recursive gossip, and recursive_gossip is used to continue the recursive gossip for the "gossip after ring reconciled" case. 

In case it wasn't clear, Riak already has gossip immediately on change/reconciliation semantics. I just changed the gossip procedure in those instances from random to recursive to ensure a deterministic propagation.

This change is a bit more chatty though. On a 8 node cluster with default gossip interval, I noticed 20-30 gossips/minute on average using the previous code, and 50-60 gossips/minute using the new code. Of course, time for ring ready convergence dropped from 4-5 minutes to a fraction of a second, so that's the tradeoff.

jtuple:
Added one more commit that adds pending ring percent (ring% after all pending ownership transfers have completed) to member_status, which should be useful for additional testing. To make testing easier, you should also enable debug lager logging in your test cluster (change console lager from info to debug in app.config) to see real-time messages on ownership changes.

Also, a few things I didn't mention last comment.

While the new gossip approach improves ring convergence time, it does not impact cluster rebalance time much because rebalance time is limited by the vnode inactivity timeout. So, if you check riak_core_stat:get_stats you will see a low max convergence_delay, but a high rebalance_delay. This will be addressed in a future ticket/pull-request.

There are a few issues here that slow-down rebalance. The first is the inactivity timeout, which will be addressed by having a separate timer that triggers handoff for pending ownership transfer even if the vnode has not been inactive. This will allow the cluster to scale under load. The current plan is to put this in a vnode_manager, and throttle it so that only the first N pending transfers (first N elements in seen) are forced at a time (others could still happen due to inactivity). I have a preliminary implementation of this already (except that it's in the vnode, and not a vnode_manager).

The second issue is that handoff waits for ring_ready before starting, yet rings are updated frequently during handoff by individual vnodes. For example, vnode 1, 2, 3 are on node A. The ring changes and is ready. Vnode 1 does an empty handoff to node B, updating the ring. The ring is no longer ready. So, 2 and 3 must now wait for the ring to propagate across the cluster and reconverge. This is no longer a major problem due to the faster gossip, but it is still an issue that should be fixed at some point. The plan would be to have the vnode_manager decide on a handoff plan, so we get one handoff decision per ring update, per node rather than per ring update, per vnode. 

The final issue is that rebalancing happens as soon as possible by the claimant, which is not ideal when adding multiple nodes at once. For example, if you add 4 nodes to the cluster at once, you'll most likely see two rebalance phases. In the first phase, node 1 will handoff 50% of it's partitions to node 2. In the second phase, nodes 1 and 2 will then handoff the appropriate amount of partitions to the remaining 2 nodes.  Handing off 50% of the ring space to node 2, just to reshuffle data around right after is not ideal. We've discussed this in the past. The plan is to make a 'rebalance' and 'add' command. 'Add' adds the node without rebalance, 'Join' is basically 'add', 'rebalance' for compatibility with existing scripts. To add multiple nodes at once 'add' them all, then 'rebalance' rather than 'join' each one.

jonmeredith:
Found two last issues

1) If I create a cluster of dev1/dev2 and another cluster of dev3/dev4 - I can join dev2 (cluster 1) to dev3 (cluster2)

```
(dev1@127.0.0.1)1> riak_core_console:member_status().
================================= Membership ==================================
Status     Ring    Node
-------------------------------------------------------------------------------
valid     50.0%    'dev1@127.0.0.1'
valid     50.0%    'dev2@127.0.0.1'
-------------------------------------------------------------------------------
Valid:2 / Leaving:0 / Exiting:0
ok
```

```
(dev2@127.0.0.1)4>  riak_core_console:member_status().
================================= Membership ==================================
Status     Ring    Node
-------------------------------------------------------------------------------
valid     33.6%    'dev2@127.0.0.1'
valid     33.2%    'dev3@127.0.0.1'
valid     33.2%    'dev4@127.0.0.1'
-------------------------------------------------------------------------------
Valid:3 / Leaving:0 / Exiting:0
ok
```

I thought the cluster name code was supposed to prevent this.

2) If I join a node to the cluster with a 'lower' name than the current claimant, the claimant stays put.  In an early iteration I thought the claimant was supposed to move to the lowest name, but I'm not sure if we changed it.

jtuple:
Concerning #1. We talked about disallowing cluster joins once, but never decided to go that route. The current proposal and implementation are simply designed to make joining a node to cluster affair, rather than a cluster to cluster affair.

Joins are always allowed and are one-way. Joining dev2 to dev3/dev4 is allowed to happen. The code that is in place ensures that joining dev2 to dev3/dev4 does not also make dev1 join the cluster. In fact, dev2 starts to ignore all gossips of dev1 after joining the dev3/dev4 cluster. You never join clusters together. You join a node to a cluster. Of course, I could make it so dev1 realizes dev2 has left, but does the equivalent of 'remove dev2'. On line 220 of riak_core_gossip, I do have a note to add in some code that tells the ignored node to stop gossiping to the current node, which I've considered implementing as a remove on the out-of-date cluster (that's what the model does).

Concerning #2. Only Justin's proposal had the claimant change to the lowest name. To prevent several corner cases that came up in modeling, my approach has been different. The claimant is always defined and never changes, unless the claimant is removed from/leaves the cluster. However, in your case, if you remove the claimant from the cluster, the lowest named node would take over as the claimant on the next ring convergence.

src/riak_core.erl
```diff
@@ -52,24 +52,68 @@ join(Node) when is_atom(Node) ->
     {ok, OurRingSize} = application:get_env(riak_core, ring_creation_size),
```
jonmeredith, line 52: note unrelated to the pull request - ring_creation_size isn't necessarily the current ring size, the env var may have been changed after the ring was created.

src/riak_core.erl
```diff
@@ -52,24 +52,68 @@ join(Node) when is_atom(Node) ->
...
-                {ok, OurRingSize} ->
-                    riak_core_gossip:send_ring(Node, node());
+            case rpc:call(Node, riak_core_ring_manager, get_my_ring, []) of
+                {ok, Ring} ->
+                    case riak_core_ring:num_partitions(Ring) of
+                        OurRingSize ->
+                            Ring2 = riak_core_ring:add_member(node(), Ring,
+                                                              node()),
+                            Ring3 = riak_core_ring:set_owner(Ring2, node()),
+                            riak_core_ring_manager:set_my_ring(Ring3),
```
jonmeredith, line 62: Why not use ring_trans here?

src/riak_core.erl
jtuple, line 52: Yeah, agreed. This was the way riak_core already did things, so I didn't change it. But yeah, may as well fix it at some point while I'm changing everything else anyway.

src/riak_core.erl
jtuple, line 62: For joins, we're not updating the existing ring, we're unconditionally replacing it with a ring retrieved from a node in the cluster we're joining.

jonmeredith:
+1 reviewed.

Future tasks will address:
- rolling upgrades
- only singleton nodes can join clusters (must leave first - no merging clusters)
- make sure node belongs to the cluster for partition ownership.
- mechanism to mark nodes as down so that progress can take place.

PR #66, opened 2011-08-04
Rename eunit target to test in Makefile.

kellymclaughlin:


jaredmorrow:
Added fix for az584

kellymclaughlin:
Thanks for fixing that up Jared!

PR #68, opened 2011-08-19
Add async vnode worker pools

Vagabond:
Initial support for vnode worker pools, using the poolboy library.

A good way to test them is to modify riak_kv_vnode to return a poolspec and then use one of the pool pids printed to the console to do something like this:

[ riak_core_vnode_worker_pool:handle_work(Pid, fun() -> io:format("hello ~p~n", [N]), timer:sleep(5000), 1/(N rem 2) end, {raw, N, self()}) || N <- lists:seq(1, 10)].

This will submit 10 jobs to the worker pool, half of which will crash. If your pool is smaller than 10, you will see the other jobs wait their turn to be worked on. You will also get the reply back to the shell, which you can inspect using flush().

An example modification to the riak_kv_vnode can be seen here: https://gist.github.com/5911fd3dde5b019ad28b

### DIFF
--- a/src/riak_core_vnode_worker.erl
+++ b/src/riak_core_vnode_worker.erl
@@ -58,8 +58,6 @@ handle_call({work, Pool, Work, WorkFrom}, {Pid, _} = From, #state{module = Mod,
             NS;
         {noreply, NS} ->
             NS
-        %{stop, NS} ->
-            %???
     end,
     %% check the worker back into the pool
     poolboy:checkin(Pool, self()),
--- a/src/riak_core_vnode_worker.erl
+++ b/src/riak_core_vnode_worker.erl
@@ -1,0 +1,83 @@
+%%
+%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(riak_core_vnode_worker).
+
+-behaviour(gen_server).
+
+-export([behaviour_info/1]).
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2,
+        code_change/3]).
+-export([start_link/1, handle_work/4]).
+
+-record(state, {
+        module :: atom(),
+        modstate :: any()
+    }).
+
+-spec behaviour_info(atom()) -> 'undefined' | [{atom(), arity()}].
+behaviour_info(callbacks) ->
+    [{init_worker,3},
+     {handle_work,3}];
+behaviour_info(_Other) ->
+    undefined.
+
+start_link(Args) ->
+    WorkerMod = proplists:get_value(worker_callback_mod, Args),
+    [VNodeIndex, WorkerArgs, WorkerProps] = proplists:get_value(worker_args, Args),
+    gen_server:start(?MODULE, [WorkerMod, VNodeIndex, WorkerArgs, WorkerProps], []).
+
+handle_work(Pid, Pool, Work, From) ->
+    gen_server:call(Pid, {work, Pool, Work, From}).
+
+init([Module, VNodeIndex, WorkerArgs, WorkerProps]) ->
+    {ok, WorkerState} = Module:init_worker(VNodeIndex, WorkerArgs, WorkerProps),
+    {ok, #state{module=Module, modstate=WorkerState}}.
+
+handle_call({work, Pool, Work, WorkFrom}, {Pid, _} = From, #state{module = Mod,
+        modstate = ModState} = State) ->
+    gen_server:reply(From, ok), %% unblock the caller
+    NewModState = case Mod:handle_work(Work, WorkFrom, ModState) of
+        {reply, Reply, NS} ->
+            riak_core_vnode:reply(WorkFrom, Reply),
+            NS;
+        {noreply, NS} ->
+            NS
+        %{stop, NS} ->
+            %???
+    end,
+    %% check the worker back into the pool
+    %io:format("returning worker to the pool~n"),
+    poolboy:checkin(Pool, self()),
+    Pid ! checkin,
+    {noreply, State#state{modstate=NewModState}};
+handle_call(_Event, _From, State) ->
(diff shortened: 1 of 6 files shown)


src/riak_core_vnode_worker.erl
```diff
@@ -0,0 +1,83 @@
...
+        {reply, Reply, NS} ->
+            riak_core_vnode:reply(WorkFrom, Reply),
+            NS;
+        {noreply, NS} ->
+            NS
+        %{stop, NS} ->
+            %???
+    end,
+    %% check the worker back into the pool
+    %io:format("returning worker to the pool~n"),
```
kellymclaughlin: Get rid of this.

kellymclaughlin:
Several commented-out `io:format` statements in `riak_core_vnode_worker_pool` that can go unless you have a specific reason for keeping them around in which case I would turn them into lager calls.

Vagabond:
Yeah, those need to go, I forgot to pull them out as I was using them for debugging.

kellymclaughlin:
Any word on getting a license on the code? Also, do you think we should just use our fork or continue to push changes to the original author and use it as possible?

kellymclaughlin:
Last comment was in reference to poolboy, btw.

Vagabond:
Poolboy is under the UNLICENSE http://unlicense.org/ which means its public-domain where a copyright holder is capable of releasing code into the public domain. In jurisdictions where you aren't able to release something into the public domain, the license gives anyone the right to use the code for any purpose anyway.

Also see http://ar.to/2010/01/dissecting-the-unlicense for detailed analysis of the license.

I am working on getting the rest of my changes re-integrated into the upstream repo, the author is reviewing my changes this week.

kellymclaughlin:
A `test` target in the Makefile for poolboy would be nice just for uniformity with other projects.

Vagabond:
Actually, looks like my changes were just merged upstream.

Vagabond:
And just added a test target.

kellymclaughlin:
How about adding `error_logger:tty(false),` to one of the first test cases to get rid of the annoying error reports when the server dies?

kellymclaughlin:
Getting occasional failure from poolboy tests. About 1 in 5 runs gives me this:

<pre><code>

INFO:  Cover compiling /Users/kelly/Applications/basho/poolboy
======================== EUnit ========================
module 'poolboy'
  module 'poolboy_tests'
    poolboy_tests: pool_startup_test...ok
    poolboy_tests: pool_overflow_test...[2.006 s] ok
    poolboy_tests: pool_empty_test...[2.505 s] ok
    poolboy_tests: pool_empty_no_overflow_test...[2.004 s] ok
    poolboy_tests: worker_death_test...[1.504 s] ok
    poolboy_tests: worker_death_while_full_test...*failed*
::error:{assertEqual_failed,
          [{module,poolboy_tests},
           {line,217},
           {expression,
               "length ( gen_fsm : sync_send_all_state_event ( Pid , get_avail_workers ) )"},
           {expected,5},
           {value,4}]}
  in function poolboy_tests:'-worker_death_while_full_test/0-fun-0-'/2
  in call from poolboy_tests:worker_death_while_full_test/0


    poolboy_tests: worker_death_while_full_no_overflow_test...*failed*
::error:{badmatch,{error,{already_started,<0.121.0>}}}
  in function poolboy_tests:worker_death_while_full_no_overflow_test/0


    poolboy_tests: pool_full_nonblocking_no_overflow_test...*failed*
::error:{badmatch,{error,{already_started,<0.121.0>}}}
  in function poolboy_tests:pool_full_nonblocking_no_overflow_test/0


    poolboy_tests: pool_full_nonblocking_test...*failed*
::error:{badmatch,{error,{already_started,<0.121.0>}}}
  in function poolboy_tests:pool_full_nonblocking_test/0


    [done in 8.045 s]
  [done in 8.045 s]
module 'poolboy_sup'
module 'poolboy_test_worker'
=======================================================
  Failed: 4.  Skipped: 0.  Passed: 5.
Cover analysis: /Users/kelly/Applications/basho/poolboy/.eunit/index.html
ERROR: One or more eunit tests failed.
</code></pre>

Vagabond:
Fix for the tests pushed.

src/riak_core_vnode_worker.erl
```diff
@@ -0,0 +1,82 @@
...
+            riak_core_vnode:reply(WorkFrom, Reply),
+            NS;
+        {noreply, NS} ->
+            NS
+        %{stop, NS} ->
+            %???
+    end,
+    %% check the worker back into the pool
+    poolboy:checkin(Pool, self()),
+    Pid ! checkin,
```
kellymclaughlin: How about using `gen_fsm:send_all_state_event` here and changing the pool module to handle `checkin` in `handle_event` rather than `handle_info`?

Vagabond: Changed

src/riak_core_vnode_worker_pool.erl
```diff
@@ -0,0 +1,129 @@
...
+handle_sync_event(_Event, _From, StateName, State) ->
+    {reply, {error, unknown_message}, StateName, State}.
+
+handle_info({'DOWN', _Ref, _, Pid, Info}, StateName, #state{monitors=Monitors} = State) ->
+    %% remove the listing for the dead worker
+    case lists:keyfind(Pid, 1, Monitors) of
+        {Pid, _, From, Work} ->
+            riak_core_vnode:reply(From, {error, {worker_crash, Info, Work}}),
+            NewMonitors = lists:keydelete(Pid, 1, Monitors),
+            self() ! checkin, %% pretend a worker just checked in
```
kellymclaughlin: Need to update this too.

src/riak_core_vnode_worker.erl
```diff
@@ -0,0 +1,82 @@
...
+handle_call({work, Pool, Work, WorkFrom}, {Pid, _} = From, #state{module = Mod,
+        modstate = ModState} = State) ->
+    gen_server:reply(From, ok), %% unblock the caller
+    NewModState = case Mod:handle_work(Work, WorkFrom, ModState) of
+        {reply, Reply, NS} ->
+            riak_core_vnode:reply(WorkFrom, Reply),
+            NS;
+        {noreply, NS} ->
+            NS
+        %{stop, NS} ->
```
kellymclaughlin: What do you think about this?Right now I am inclined to remove it and if as we make the changes on the kv side the need for it becomes apparent we add it in at that time.

kellymclaughlin:
I think everything looks good code-wise. My only other concern is testing. It seems like the test suggestions you included in the initial pull request comments might make a good starting point for some basic tests using a dummy test vnode module. Some eqc tests for poolboy might also be nice at some point. I don't know if now is the best time to do this or if we should circle back to it as a separate task as time permits. We will have some basic functional exercise of the code via the key listing tests in kv, but it would be nice to do some direct testing. What's your thoughts?

Vagabond:
I've added some simple tests to exercise the pool. I'd love to do some EQC stuff, but I'm not sure we have the time right now, especially considering my lack of experience with it.

Regardless, the new tests get decent coverage of the new code.

kellymclaughlin:
+1, merge it!

PR #69, opened 2011-08-22
Az577 riak control, track handoff

russelldb:
Add code to riak_core_vnode_master and riak_core_handoff_manager so that the state of a vnode can be ascertained without asking the vnode (specifically avoiding calls to get_mod_index.)

### DIFF
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -21,13 +21,17 @@
 -export([add_exclusion/2, get_handoff_lock/1, get_exclusions/1]).
 -export([remove_exclusion/2]).
 -export([release_handoff_lock/2]).
--record(state, {excl}).
+-export([add_handoff/3]).
+-export([remove_handoff/2]).
+-export([get_handoff/2]).
+-export([all_handoffs/0]).
+-record(state, {excl, handoffs}).
 
 start_link() ->
     gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
 
 init([]) ->
-    {ok, #state{excl=ordsets:new()}}.
+    {ok, #state{excl=ordsets:new(), handoffs=dict:new()}}.
 
 add_exclusion(Module, Index) ->
     gen_server:cast(?MODULE, {add_exclusion, {Module, Index}}).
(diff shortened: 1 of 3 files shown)


src/riak_core_vnode_master.erl
```diff
@@ -59,6 +59,10 @@ get_vnode_pid(Index, VNodeMod) ->
     RegName = reg_name(VNodeMod),
     gen_server:call(RegName, {Index, get_vnode}, infinity).
 
+is_vnode_pid(Index, VNodeMod) ->
```
beerriot, line 62: It looks like `is_vnode_pid/2` is supposed to be the same as `get_vnode_pid/2`, except it doesn't start up the vnode if it's not running.  I think it might be better to add `get_vnode_pid/3` that takes a boolean `StartIfUnstarted` parameter, with get_vnode_pid/2 defaulting that param to 'true'.  The name `is_vnode_pid` makes me expect that the function is a predicate that accepts a pid.

src/riak_core_vnode.erl
```diff
@@ -214,6 +214,7 @@ active(handoff_complete, State=#state{mod=Mod,
                                       handoff_token=HT}) ->
     %% ?debugFmt("Finished HO: ~p :: ~p -> ~p~n", [Idx, Prev, New]),
     riak_core_handoff_manager:release_handoff_lock({Mod, Idx}, HT),
+    riak_core_handoff_manager:remove_handoff(Mod, Idx),
```
beerriot, line 217: Would it be possible to roll `add_handoff/3` and `remove_handoff/2` into `get_handoff_lock/1` and `release_handoff_lock/2`?  That would reduce the burden on future code having to remember to call _both_ functions every time.

src/riak_core_handoff_manager.erl
```diff
@@ -21,13 +21,17 @@
...
+-export([get_handoff/2]).
+-export([all_handoffs/0]).
+-record(state, {excl, handoffs}).
 
 start_link() ->
     gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
 
 init([]) ->
-    {ok, #state{excl=ordsets:new()}}.
+    {ok, #state{excl=ordsets:new(), handoffs=dict:new()}}.
```
beerriot, line 34: As long as the number of concurrent handoffs is low, `orddict` might be more efficient (and easier to read in a stack trace) than `dict`.

jonmeredith:
I'd guess this code is sadly orphaned and got re-implemented.  Any reason to keep the PR open Russell?

russelldb:
Nope, I'll kill it

PR #70, opened 2011-08-25
Bucket fixup modules

Vagabond:


src/riak_core.erl
```diff
@@ -77,19 +78,38 @@ vnode_modules() ->
...
+bucket_fixups() ->
+     case application:get_env(riak_core, bucket_fixups) of
+        undefined -> [];
+        {ok, Mods} -> Mods
+    end.
+
+%% @doc Register a riak_core application.
+register_application([]) ->
+    ok;
+register_application([{bucket_fixup, FixupMod}|T]) ->
```
rzezeski: I find it odd that this is called `register_application` but the things being registered are either bucket fixups or vnodes.  Why not `register_bucket_fixup` and `register_vnode`?  Or even just keep it one function but call it `register`.

src/riak_core.erl
```diff
@@ -77,19 +78,38 @@ vnode_modules() ->
...
+    register_application(T).
+
 register_vnode_module(VNodeMod) when is_atom(VNodeMod)  ->
+    register_riak_core_application_module(VNodeMod, vnode_modules).
+
+register_riak_core_application_module(Module, Key) when is_atom(Module), is_atom(Key) ->
     {ok, App} = case application:get_application(self()) of
         {ok, AppName} -> {ok, AppName};
-        undefined -> app_for_module(VNodeMod)                        
+        undefined -> app_for_module(Module)
```
rzezeski, line 103: According to the docs for R14B02 you can pass a module to `application:get_application/1`.  Also, `get_application/0` will get the app of `self()`.

src/riak_core_bucket.erl
```diff
@@ -92,6 +100,29 @@ get_bucket(Name, Ring) ->
         {ok, Bucket} -> Bucket
     end.
 
+%% Temporarily duplicate here from riak_core_ring_manager until the default
```
rzezeski: Instead of duplicating why not just export from ring_manager for now (or export from here)?  Otherwise someone else patching this code might not realize it's copied.

src/riak_core.erl
Vagabond: Jon told me to call it register, but I didn't listen. I'll change it

src/riak_core_bucket.erl
Vagabond: Fixed

src/riak_core.erl
Vagabond, line 103: This is legacy code and I'd prefer not to change it in case there's some reason behind the weird way the application lookup is being done.

rzezeski, line 103: Don't show the code your fear :)

Given we're on the verge of a release I guess I can allow it.

rzezeski:
+1 to merge

src/riak_core.erl
```diff
@@ -77,19 +78,38 @@ vnode_modules() ->
...
+    register_riak_core_application_module(FixupMod, bucket_fixups),
+    register(T);
+register([{vnode_module, VNodeMod}|T]) ->
+    register_riak_core_application_module(VNodeMod, vnode_modules),
+    register(T).
+
 register_vnode_module(VNodeMod) when is_atom(VNodeMod)  ->
+    register_riak_core_application_module(VNodeMod, vnode_modules).
+
+register_riak_core_application_module(Module, Key) when is_atom(Module), is_atom(Key) ->
```
jonmeredith: register_mod would have probably been sufficient - and s/Key/Type/g would probably be more descriptive

src/riak_core_ring_manager.erl
```diff
@@ -268,11 +280,64 @@ back(N,X,[H|T]) ->
...
+        Fixups ->
+            Buckets = riak_core_ring:get_buckets(Ring),
+            lists:foldl(
+                fun(Bucket, AccRing) ->
+                        BucketProps = riak_core_bucket:get_bucket(Bucket, Ring),
+                        %% Merge anything in the default properties but not in
+                        %% the bucket's properties. This is to ensure default
+                        %% properties added after the bucket is created are
+                        %% inherited to the bucket.
+                        MergedProps = lists:ukeymerge(1,
```
jonmeredith: Why not export riak_core_bucket:merge_props and reuse that here?

jonmeredith:
+1 merge.

With the search+repl changes fixes the hook issues.

PR #71, opened 2011-08-26
az612: Add joining/down status, fix bugs

jtuple:
Add 'joining' member status to new cluster membership model and
implementation. When a node joins a cluster, it comes in with status
'joining' rather than 'valid'. The claimant then moves a node from 'joining'
to 'valid' after it ensures all cluster members have learned of the new
node joining the cluster. This change guarantees that all 'valid' members
vote on ring ready consensus under various failure scenarios.

Add 'down' member status to new cluster membership model and implementation.
The state is designed to allow a user to mark a down node as 'down' in order
to allow the rest of the cluster to converge. A vote from a 'down' node is
not necessary for ring ready consensus, and therefore a 'down' node's ring
state may become outdated. If a 'down' node gossips to another node that
believes it to be down (such as after coming back online), the other node
tells the 'down' node to rejoin the cluster, therefore making its state
current. Nodes do not gossip to 'down' nodes, and 'down' node ownership is
not changed during a rebalance.

Incorporate minor changes and bug fixes:
-- Fix next merging bug in riak_core_ring:remove_node and model.
-- Fix bug with nodes moving from 'leaving' to 'exiting' while having pending
   indices.
-- Fix negative random seed bug in join/membership model.
-- Change core vnode so that a vnode does not shutdown if a completed handoff
   was to node that is now known to be 'invalid'.
-- Change update_ring to remove tuples from next for all invalid nodes, even
   for completed transfers.
-- Change leave in model to be a local transition like in the implementation.
-- Handle "neither claimant valid" case in reconcile_ring.

PR #73, opened 2011-08-30
Deterministic vclock merge

jonmeredith:
Incorporate changes from Trifork branch to make vclock merging deterministic (thanks Kresten/Erik).

The makes vclock:merge(A, B) == vclock:merge(B, A) when two node ids have the same timestamp.
Required by https://github.com/basho/riak_core/pull/71 to ensure that rings converge.


slfritchie:
I've a style twiddling patch at https://gist.github.com/f765e3331ca134b8aed5 that
you may want to consider.  Otherwise, +1 to merge, thanks!

PR #74, opened 2011-09-01
Az630: Add legacy gossip mode + rolling upgrade between gossip versions

jtuple:
This PR adds a legacy gossip mode to the new gossip/membership code as well as supports rolling upgrades. In-place upgrades, leave/rejoin, and stable hybrid/mixed-version clusters are all supported. Read the commit message for the first commit for more details.

At a minimum please re-test the following scenarios that I've already tested manually:
1. In-place upgrades. Have 0.14.2 cluster, and in-place upgrade a node at a time. Cluster should stay in legacy mode until all nodes are upgraded, and then move into new gossip/membership protocol.
2. Leave/rejoin upgrades. Similar to above.
3. Try setting the {riak_core, legacy_gossip, true} in the app.config and ensure that the node uses legacy gossip.
4. Test that the cluster automatically changes from new to legacy gossip whenever an old or legacy-mode node joins, and back to new if that node leaves/is removed.

A few things. In a mixed cluster environment, 0.14.2 nodes will fail on ringready/transfers due to the RPC call retrieving the new ring. This PR negotiates formats over gossip, but not over RPC. However, transfers/ringready from a new code node will work fine in a mixed cluster.

Another issue is that in a mixed-mode cluster, transfers is somewhat useless. It will always say there is some outstanding transfers because the new nodes are running riak_pipe but the old nodes are not. This is the same on master. It's an issue with handoff waiting for node_watcher:services to claim the target node is running pipe, which it doesn't because 0.14.2 doesn't run pipe. For similar reasons, old nodes may sometimes not shutdown after handoff in a mixed-version cluster.

Note: The riak_core_gossip_legacy.erl file is copy of gossip from master with some minor changes. It's much clear to review if you look at it with respect to master's gossip rather than as a new file:
git diff master:src/riak_core_gossip.erl src/riak_core_gossip_legacy.erl

### DIFF
--- a/src/riak_core_ring.erl
+++ b/src/riak_core_ring.erl
@@ -1518,11 +1518,12 @@ membership_test() ->
     {_, RingA7} = reconcile(RingB2, RingA6),
     ?assertEqual([nodeA, nodeB, nodeC], all_members(RingA7)),
 
-    Priority = [{invalid,1}, {valid,2}, {exiting,3}, {leaving,4}],
+    Priority = [{invalid,1}, {down,2}, {joining,3}, {valid,4}, {exiting,5},
+                {leaving,6}],
     RingX1 = fresh(nodeA),
     RingX2 = add_member(nodeA, RingX1, nodeB),
     RingX3 = add_member(nodeA, RingX2, nodeC),
-    ?assertEqual(valid, member_status(RingX3, nodeC)),
+    ?assertEqual(joining, member_status(RingX3, nodeC)),
 
     %% Parallel/sibling status changes merge based on priority
     [begin
--- a/src/riak_core_ring.erl
+++ b/src/riak_core_ring.erl
@@ -53,9 +53,15 @@
          update_meta/3]).
 
 -export([cluster_name/1,
+         legacy_ring/1,
+         legacy_reconcile/2,
+         upgrade/1,
+         downgrade/2,
          claimant/1,
          member_status/2,
          all_member_status/1,
+         update_member_meta/5,
+         get_member_meta/3,
          add_member/3,
          remove_member/3,
          leave_member/3,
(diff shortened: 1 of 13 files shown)


slfritchie:
Item #1 should be doable with basho_expect's Run-systest-rolling-upgrade
script as-is.  #3 would require only small changes: b_e can set arbitrary 
knobs in app.config.  I'm currently running a rolling upgrade test from 0.14.2
to 0.14.2 (not a typo) to make certain that recent b_e changes haven't broken
anything big.

src/riak_core_ring.erl
```diff
@@ -511,6 +626,77 @@ handoff_complete(State, Idx, Mod) ->
...
 
+%% ===================================================================
+%% Legacy reconciliation
+%% ===================================================================
+
+%% @doc Incorporate another node's state into our view of the Riak world.
+legacy_reconcile(ExternState, MyState) ->
+    case vclock:equal(MyState#chstate.vclock, vclock:fresh()) of
+        true -> 
+            {new_ring, #chstate{nodename=MyState#chstate.nodename,
```
jonmeredith, line 637: Is this deliberate use of #chstate when all the others have been converted to ?CHSTATE

jtuple, line 637: Yes. All the legacy_ routines are called expecting the old chstate. I used to directly call into a copied riak_core_ring_legacy module, but decided to copy out the few functions i needed rather than have an entire file sitting around.

jonmeredith:
Almost +1 merge.  Shame about all the legacy stuff littering the new code but it's hard to avoid - cluster membership code worked great both leave old/rejoin new cases and stop/upgrade/start and the app env prevented teh gossip switchover.

Only thing I've found is eunit test failure

```

  riak_core_ring: membership_test...*failed*
::error:{assertEqual_failed,[{module,riak_core_ring},
                           {line,1525},
                           {expression,"member_status ( RingX3 , nodeC )"},
                           {expected,valid},
                           {value,joining}]}
  in function riak_core_ring:'-membership_test/0-fun-6-'/2
  in call from riak_core_ring:membership_test/0
```

jonmeredith:
+1 merge.

PR #75, opened 2011-09-02
Az642: Preliminary VNode Manager / Rebalance Under Load

jtuple:
This PR adds a preliminary vnode manager that is currently used to 1) trigger ownership handoffs independently from vnode inactivity and 2) help vnodes maintain their forwarding status. The vnode manager will eventually take over most of the vnode pid tracking/spawning logic currently included in vnode_master.

For further information, read the full commit messages.

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -34,7 +34,8 @@
          code_change/4]).
 -export([reply/2]).
 -export([get_mod_index/1,
-         update_forwarding/2]).
+         update_forwarding/2,
+         trigger_handoff/1]).
 
 -spec behaviour_info(atom()) -> 'undefined' | [{atom(), arity()}].
 behaviour_info(callbacks) ->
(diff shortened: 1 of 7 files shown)


src/riak_core_vnode.erl
```diff
@@ -126,21 +138,30 @@ continue(State, NewModState) ->
...
+            riak_core_vnode_manager:set_not_forwarding(self(), true),
+            State#state{forward=undefined}
+    end.
+
+vnode_command(Sender, Request, State=#state{index=Index,
+                                            mod=Mod,
+                                            modstate=ModState,
+                                            forward=Forward}) ->
+    %% Check if we should forward
+    case Forward of
```
jonmeredith, line 177: Need to apply the same forwarding logic to handle_coverage too.

jonmeredith:
Please could you add a comment block to riak_core_vnode explaining how the transition to a new node works now with the handoff_node and forwarding in the new world.

After that and the handle_coverage fix I'm +1 on it.

As we discussed in chat there's more refactoring to do here but it'll have to wait post-1.0.

PR #77, opened 2011-09-02
Minor changes before rebase/merge of new cluster membership code

jtuple:
Changes:
-- Join checks against current ring_size rather than application variable.
-- Join requires joining node to not already be part of a cluster.
-- Down fails if the cluster is in legacy gossip mode.
-- riak_core_ring:ready_members returns nodes guaranteed safe for requests.


jonmeredith:
+1 on visual inspection.

PR #78, opened 2011-09-03
Overhaul cluster membership, ring format, and gossip protocol

jtuple:
Tickets: AZ462 AZ533 AZ630 AZ642
Fixes: BZ1024 BZ878 BZ688 BZ931 BZ869

Change cluster membership/gossip/ring in the following manner:
-- Partition ownership and cluster membership are decoupled concepts.

-- A single node in the cluster makes all partition claim decisions
   in a deterministic manner, avoiding endless churn possible in random
   merging. This single node (the claimant) can be assumed by any node in the
   cluster, and therefore avoids introducing a single point of failure.

-- All data from all vnode modules is guaranteed to be handed off to a new
   owner before the owner takes over ownership for a partition, thus avoiding
   requests prematurely being sent to a node that does not yet have the data.

-- Nodes will always handoff their data before cleaning exiting the cluster,
   even if the node is restarted before finishing handoff.

-- Waiting for ring convergence is now part of the gossip/membership protocol,
   and is used to ensure consistent ring transitions across the cluster.

-- Joining/leaving a cluster is handled more gracefully, and there is no
   need to wait for ring convergence (riak-admin ring-ready) as is previously
   suggested. Adding 20 nodes to a cluster at once should "just work".

-- Handoff related to partition ownership changes can now occur under load,
   therefore allowing a cluster to scale up/down while handling normal
   requests.

Other changes:
-- Support for new commands: member_status, ring_status.
-- Tracking of various new riak_core stats.


jtuple:
See prior incremental pull-requests / reviews:
https://github.com/basho/riak_core/pull/65
https://github.com/basho/riak_core/pull/71
https://github.com/basho/riak_core/pull/74
https://github.com/basho/riak_core/pull/75
https://github.com/basho/riak_core/pull/77

PR #79, opened 2011-09-04
Eunit ring fixes

jonmeredith:
Fixes to a make a couple of EQC tests pass.

Changed riak_core_vnode to consistently use get_my_ring() rather than get_raw_ring() so the ring manager doesn't need to start.

Changed node_watcher_qc to add nodes to the test ring as well as set ownership (actually, probably doesn't need to set ownership any more, but left as it was).


jtuple:
Changes look to be safe. Verified creating 3-node cluster and verified vnode fowarding (code path changed in this PR) still worked as desired.

Ran full riak_core eunit tests 10 times, all passed. Ran core_vnode_eqc and node_watcher_qc an additional 10 times individually to ensure test didn't depend on earlier tests running. All tests passed.

+1 merge.

PR #82, opened 2011-09-07
Add util support for checking node architecture.

massung:
Added support for checking the architecture of the current node.

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -37,7 +37,8 @@
          start_app_deps/1,
          build_tree/3,
          rpc_every_member/4,
-         rpc_every_member_ann/4]).
+         rpc_every_member_ann/4,
+	 is_arch/1]).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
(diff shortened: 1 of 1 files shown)


src/riak_core_util.erl
```diff
@@ -262,6 +263,19 @@ build_tree(N, Nodes, Opts) ->
...
+get_arch () -> string:to_lower(erlang:system_info(system_architecture)).
+
+%% Checks if this node is of a given architecture
+-spec is_arch (atom()) -> boolean().
+is_arch (linux) -> string:str(get_arch(),"linux") > 0;
+is_arch (darwin) -> string:str(get_arch(),"darwin") > 0;
+is_arch (sunos) -> string:str(get_arch(),"sunos") > 0;
+is_arch (osx) -> is_arch(darwin);
+is_arch (solaris) -> is_arch(sunos);
+is_arch (_) -> false.
```
rzezeski: Instead of just returning false I might throw here to make it clear it doesn't understand the passed in arch.

rzezeski:
Jeff, the following is what I use.  You might have to use `untabify` command

```
(setq-default indent-tabs-mode nil)
(setq tab-width 4)
```

rzezeski:
+1 to merge

PR #83, opened 2011-09-09
Bz1178 ring fix on register

jonmeredith:
Makes the bucket property fixups run on register.

Tested by spinning up riak with riak_search {enabled, false} , setting {search, true} bprops, checking the bucket properties to make sure there are no post-commit hooks.
Then restarting riak with riak_search {enabled, true} and checkign the postcommit hook - it should be set now.

### DIFF
--- a/src/riak_core.erl
+++ b/src/riak_core.erl
@@ -265,6 +265,11 @@ register(Props) ->
 
 %% @doc Register a named riak_core application.
 register(_App, []) ->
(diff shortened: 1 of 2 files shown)


Vagabond:
This looks good, updating the older tests that didn't use register because there was no way to specify the app name might be nice, but not essential.

+1 for merge.

jonmeredith:
Thanks for the review - I left the refactor of the other tests for a rainy day.

PR #84, opened 2011-09-09
Bz1178 ring fix on register 1.0

jonmeredith:
Same as #83 on 1.0


Vagabond:
+1 for merge, same as other PR.

PR #85, opened 2011-09-15
Change log level of worker pool start message.

kellymclaughlin:
The message is too verbose to be on for the default configuration. Change it to debug level from notice level.


massung:
+1 to merge

PR #88, opened 2011-09-15
Fix bug in riak_core_coverage_fsm

kellymclaughlin:
This change fixes a bug in `riak_core_coverage_fsm` where the updated state is not passed to the module implementing the behavior in the `finish` call. This can lead to incomplete results for operations that accumulate results in the state and do something with them in the `finish` function.


massung:
+1 to merge

PR #89, opened 2011-09-15
Az708 async core vnode eqc 1.0

jonmeredith:
Extends core_vnode_eqc to test async commands.

Currently has a timeout in prop_simple to make pass until bz://1188 is fixed.


Vagabond:
This is already merged as part of another branch.

PR #90, opened 2011-09-19
Bz1188 worker pool cleanup

Vagabond:


src/riak_core_vnode_sup.erl
```diff
@@ -28,10 +28,10 @@
 -export([start_vnode/2]).
 
 start_vnode(Mod, Index) when is_integer(Index) -> 
-    supervisor:start_child(?MODULE, [Mod, Index]).
+    supervisor_pre_r14b04:start_child(?MODULE, [Mod, Index]).
```
jonmeredith, line 31: Can you add code comment about why we did this and when we expect it to be safe to deprecate.

jonmeredith:
+1 merge after adding comment.

PR #92, opened 2011-09-22
Add helper functions for config values from proplist or env.

reiddraper:
AZ669

### DIFF
--- a/src/app_helper.erl
+++ b/src/app_helper.erl
@@ -24,6 +24,10 @@
 
 -export([get_env/1, get_env/2, get_env/3, get_prop_or_env/3, get_prop_or_env/4]).
 
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
 %% ===================================================================
 %% Public API
 %% ===================================================================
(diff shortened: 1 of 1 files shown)


src/app_helper.erl
```diff
@@ -22,7 +22,7 @@
 
 -module(app_helper).
 
--export([get_env/1, get_env/2, get_env/3]).
+-export([get_env/1, get_env/2, get_env/3, get_prop_or_env/3, get_prop_or_env/4]).
```
kellymclaughlin: This is strictly my opinion, but when the number of exports gets beyond 2 or 3 and when there is more than one different function name I like to add a line break after each exported function to make a nice columnar list. It makes it easy to read and easy to edit.

src/app_helper.erl
```diff
@@ -49,3 +49,20 @@ get_env(App, Key, Default) ->
...
+%% @doc Retrieve value for Key from Properties if it exists, otherwise
+%%      return from the application's env.
+-spec get_prop_or_env(atom(), [{atom(), term()}], atom()) -> term().
+get_prop_or_env(Key, Properties, App) ->
+    get_prop_or_env(Key, Properties, App, undefined).
+
+%% @doc Return the value for Key in Properties if it exists, otherwise return
+%%      the value from the application's env, or Default.
+-spec get_prop_or_env(atom(), [{atom(), term()}], atom(), term()) -> term().
+get_prop_or_env(Key, Properties, App, Default) ->
```
kellymclaughlin, line 70: For consistency with the `get_env` functions in this module, I would say reorder the parameters to be `App, Key, Props, Default` and `App, Key, Props` for `get_prop_or_env/3`.

kellymclaughlin:
I think this presents a good opportunity for you to work with eunit. Can you create a set of basic eunit tests that exercise this function. The tests in `riak_core_config` might be some good examples to follow.

src/app_helper.erl
```diff
@@ -49,3 +57,65 @@ get_env(App, Key, Default) ->
...
+       fun get_prop_or_env_default_value_test_case/0,
+       fun get_prop_or_env_undefined_value_test_case/0,
+       fun get_prop_or_env_from_env_test_case/0,
+       fun get_prop_or_env_from_prop_test_case/0,
+       fun get_prop_or_env_from_prop_with_default_test_case/0
+      ]
+    }.
+
+setup() ->
+    application:load(riak_core),
```
kellymclaughlin: I expected you would use `riak_core` as the App, but since you use `bogus_app` you can eliminate this call to '`load` and the call to `stop` in the `cleanup` function. Just have `cleanup` return `ok`.

src/app_helper.erl
```diff
@@ -49,3 +57,65 @@ get_env(App, Key, Default) ->
...
+    application:set_env(bogus_app, envkeyone, value),
+    application:set_env(bogus_app, envkeytwo, valuetwo).
+
+cleanup(_Ctx) ->
+    application:stop(riak_core).
+
+get_prop_or_env_default_value_test_case() ->
+    ?assert(get_prop_or_env(key, [], bogus, default) =:= default).
```
kellymclaughlin: Just as a point of information, there are several other assert macros that eunit provides that are useful. One is `?assertEqual` where the first parameter is the expected value and the second is an expression. The other is `?assertMatch` which is like `?assertEqual`, but can selectively pattern match on the result of the expression evaluation.

kellymclaughlin:
+1 to merge.

reiddraper:
merged in 6f4fb7eb98a8eda79d84408d1405ebf7ecf9cb45

PR #93, opened 2011-09-23
BZ 1220: Fix ringready in 1.0 to show all nodes in a mixed-version cluster

jtuple:
In a mixed node cluster, 'riak-admin ringready' will fail on 0.14.2 nodes, and only list 1.0 nodes on a 1.0 cluster. This fix makes it so that ringready from a 1.0 node will list all nodes when part of a mixed cluster.

https://issues.basho.com/show_bug.cgi?id=1220


Vagabond:
+1 to merge.

PR #94, opened 2011-09-23
Make vnode forward messages until vnode master removes it

jonmeredith:
With new cluster membership logic the shutdown on handoff completion happens much closer to the node processing live traffic.  Ensure that all commands passed into the vnode are forwarded to the new node before shutting down.


jtuple:
Jon, it looks like this is a 1.0 change given that you've branched off 1.0 initially. Could you redo this as a pull-request into 1.0 rather than into master? Lots of divergent changes showing up here that make it hard to see your changes.

jonmeredith:
thx Joe - will resubmit.

PR #95, opened 2011-09-25
Make vnode forward messages until vnode master removes it

jonmeredith:
With new cluster membership logic the shutdown on handoff completion happens much closer to the node processing live traffic. Ensure that all commands passed into the vnode are forwarded to the new node before shutting down.

### DIFF
--- a/src/riak_core_vnode_master.erl
+++ b/src/riak_core_vnode_master.erl
@@ -32,7 +32,7 @@
          sync_command/4,
          sync_spawn_command/3, make_request/3,
          make_coverage_request/4,
-         unregister_vnode/2,
+         unregister_vnode/2, unregister_vnode/3,
          all_nodes/1, reg_name/1, all_index_pid/1]).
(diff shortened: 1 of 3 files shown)


src/riak_core_vnode_master.erl
```diff
@@ -136,8 +136,12 @@ make_coverage_request(Request, KeySpaces, Sender, Index) ->
                           sender=Sender,
                           request=Request}.
 
-unregister_vnode(Index, VMaster) ->
-    gen_server:cast(VMaster, {unregister, Index}).
+unregister_vnode(Index, VNodeMod) ->
```
Vagabond, line 139: Why not just remove unregister_vnode/2?

jonmeredith, line 139: Could do.  Minimal-change-gone-mad.

Vagabond:
I don't really like the style of this change, but I understand the expediency for getting it into 1.0.

I think I'll give it a +1.

PR #96, opened 2011-09-27
Sync at handoff complete

jonmeredith:
The new cluster membership code switched to forwarding
once handoff is complete.  Without this change the vnode
starts forwarding while the new owner is still processing
buffered TCP data.


jtuple:
+1 merge

PR #97, opened 2011-09-27
AZ 761: Prevent nodes from prematurely leaving

jtuple:
Change ring_ready to also consider exiting nodes in it's decision, fixing a corner case that led to notfounds. This was the previous behavior during most of the development of the new clustering code (and was heavily tested), but was changed last minute to address a bug where the current claimant would shutdown after leaving but before other nodes saw it transition from exiting to leaving, and therefore the ring would never converge and a new claimant would never be selected.

Since the exiting change was reverted to old behavior, the above claimant issue was addressed by modifying the claimant so that it does not transition itself from exiting to invalid, and therefore does not shutdown. After transitioning to exiting, the ring will converge, a new claimant will be selected (since only valid and leaving nodes can be the claimant), and the new claimant will transition the previous claimant from exiting to invalid and start the shutdown process.

Updated and verified in the membership EQC model as well.

### DIFF
--- a/src/riak_core_ring.erl
+++ b/src/riak_core_ring.erl
@@ -597,7 +597,7 @@ ring_ready(State0) ->
     Owner = owner_node(State0),
     State = update_seen(Owner, State0),
     Seen = State?CHSTATE.seen,
-    Members = get_members(State?CHSTATE.members, [valid, leaving]),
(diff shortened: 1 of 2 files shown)


jonmeredith:
+1 merge

PR #98, opened 2011-09-29
Fix bug that prevents down nodes from rejoining

jtuple:
https://issues.basho.com/show_bug.cgi?id=1229


jtuple:
This change is entirely isolated and manually verified in all possible call paths (there's exactly two in all of riak) that lead into the modified code. There is zero chance this can cause new bugs.

jonmeredith:
+1 merge - did multiple leaves/joins with nodes down and restarting downed nodes always recovered.

PR #99, opened 2011-09-29
Add check for deleted modstate in handle_info and maybe_handoff.

jonmeredith:
Check for the modstate being wrapped with {deleted, ModState} while the vnode is unregistering,

Fixes: bz://1227


rustyio:
+1 merge

rustyio:
If/when we refactor this, might consider creating a separate variable in the #state record rather than wrapping the ModState in a {deleted, ...} tuple. 

Also, this is an FSM but only has one explicit state (active), but many implicit states (normal, handoff, forwarding, ...). Would be good to make distinct FSM states for each of those implicit states.

PR #101, opened 2011-10-07
Fix bz1235: vnodes crashing when receiving unexpected handoff

jtuple:
Sync messages sent to vnodes when they are exiting cause them to crash:
https://issues.basho.com/show_bug.cgi?id=1235

After talking to Jon, the decision for now is to prevent vnodes from crashing, and pass an error up to the handoff receiver. The handoff receiver will check for the error, and exit. This will cause the handoff sender to crash, but this an acceptable crash (compared to crashing vnodes). The handoff will eventually be restarted in the future against the proper target node, and complete successfully.


jonmeredith:
Code looks good, will try and execute it later.

jonmeredith:
+1 merge.

jonmeredith:
Just noticed a compile warning on src/riak_core_handoff_receiver.erl:109: Warning: variable 'Err' is unused.  Would be good to clean up before merge.

PR #102, opened 2011-10-11
Fix BZ1242: Stalled handoff when ring is fixed-up

jtuple:
This pull-request fixes BZ1242 by changing riak_core_vnode:should_handoff to use get_raw_ring rather than get_my_ring. This is necessary because get_my_ring returns the fixed-up version of the ring, which has a locally modified vector clock that will never satisfy the ring_ready check used in should_handoff.

In order to prevent other misuses of the fixed-up ring, this pull-request also adds new tainted ring logic that identifies misuses of the fixed-up ring. 

Testing with the tainted logic revealed several additional misuses of the fixed-up ring which are also fixed in this pull-request. 

Manual testing revealed no more tainted error messages after all committed changes.

### DIFF
--- a/src/riak_core_ring.erl
+++ b/src/riak_core_ring.erl
@@ -397,8 +397,12 @@ random_other_active_node(State) ->
 -spec reconcile(ExternState :: chstate(), MyState :: chstate()) ->
         {no_change, chstate()} | {new_ring, chstate()}.
 reconcile(ExternState, MyState) ->
-    check_tainted(ExternState, "Error: reconciling a tainted ring"),
-    check_tainted(MyState, "Error: reconciling a tainted ring"),
+    check_tainted(ExternState,
+                  "Error: riak_core_ring/reconcile :: "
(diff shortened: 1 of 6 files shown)


src/riak_core_ring.erl
```diff
@@ -364,6 +385,8 @@ random_other_active_node(State) ->
 -spec reconcile(ExternState :: chstate(), MyState :: chstate()) ->
         {no_change, chstate()} | {new_ring, chstate()}.
 reconcile(ExternState, MyState) ->
+    check_tainted(ExternState, "Error: reconciling a tainted ring"),
```
slfritchie: Joe, for each of the `check_tainted()` cases here and following ... is it useful to
include in the string: 1. the caller, e.g. "reconcile", and 2. (when applicable) which
ring is tainted (e.g. `reconcile()`'s `ExternState` or `MyState`)?

slfritchie:
+1 to merge, thanks!

PR #103, opened 2011-10-18
Fix a couple of specs

rzezeski:
Fix two specs with incorrect return types.  These specs are giving me problems
upstream while trying to spec out riak_search.


rzezeski:
I know it's not in the PR but @argv0 did +1 this.

PR #105, opened 2011-10-31
Initial Riak Control GUI commits.

massung:


src/riak_core_handoff_manager.erl
```diff
@@ -21,13 +21,26 @@
 -export([add_exclusion/2, get_handoff_lock/1, get_exclusions/1]).
 -export([remove_exclusion/2]).
 -export([release_handoff_lock/2]).
--record(state, {excl}).
+-export([add_handoff/3,remove_handoff/2,all_handoffs/0,get_handoffs/1]).
+-record(state, {excl,handoffs}).
```
Vagabond, line 25: typespecs for the record elements would be nice.

massung, line 25: I don't know what the type spec for the exclusions are. I can put a comment or one for the handoffs, though.

Vagabond, line 25: To the source!:

https://github.com/erlang/otp/blob/master/lib/stdlib/src/ordsets.erl#L29

looks like ordsets:ordsets() is the type you want.

src/riak_core_handoff_manager.erl
```diff
@@ -57,8 +70,27 @@ release_handoff_lock(LockId, Token) ->
     
 handle_call({get_exclusions, Module}, _From, State=#state{excl=Excl}) ->
     Reply =  [I || {M, I} <- ordsets:to_list(Excl), M =:= Module],
-    {reply, {ok, Reply}, State}.
+    {reply, {ok, Reply}, State};
+handle_call(all_handoffs, _From, State=#state{handoffs=Hoffs}) ->
+    {reply, {ok, Hoffs}, State};
+handle_call({get_handoffs, Idx}, _From, State=#state{handoffs=Hoffs}) ->
+    All=[{Mod,dict:find(Idx,ModHoffs)} || {Mod,ModHoffs} <- Hoffs],
```
Vagabond: I'm not thrilled with a LC followed by a filter, why not do it all together with a lists:foldl? something like

``` erlang
lists:foldl(fun({Mod, MofHoffs}, Acc) ->
    case dict:find(Idx, ModHoffs) of
       {ok, Value} ->
          [Value|Acc];
      _ ->
         Acc
    end, [], Hoffs)
```

Or something, rather than iterating the list 3 separate times?

massung: Yeah, thought about that. Wanted to be explicit about what was going on (and I think people have a hard time wrapping their heads around fold returns). I was assuming the list would never get that big where performance would be an issue.  But I'm up for changing it.

Vagabond: I actually find the fold cleaner to read, and its certainly faster, and generates less intermediate data for the GC to scoop up later.

jtuple:
There are issues with the caching of the handoff information. The current design has vnodes add handoff information to the handoff_manager when a handoff starts, and removes the information when the handoff completes or there is a detected handoff error. Unfortunately, if a handoff_sender crashes or a vnode crashes, the cached handoff information is not removed. This is easy to reproduce. Start 2 nodes, write a bunch of data to one node (to make handoff take some time), then join the second node to the first to start handoffs. While handoff is occuring, kill all the vnodes on the first node:

``` erlang
Pids = [Pid || {_,Pid,_,_} <- supervisor:which_children({riak_core_vnode_sup, node()})].
[exit(Pid, kill) || Pid <- Pids].
```

Then, stop the second Riak node so that new handoff doesn't start. The output of `riak_core_handoff_manager:all_handoffs()` will remain non-empty for eternity.

Here's the good news. The handoff_sender is linked to the vnode. So, either one will cause the other to die as well. I think the cleanest option would be to pass the vnode pid as part of the information cached in the handoff_manager, and to have the handoff_manager filter out information associated with a pid that is no longer alive (`is_pid(Pid) andalso is_process_alive(Pid)`). You could probably use a monitor as well, but that's more code. Or, could rely upon catching the process exit in the vnode module, but that won't work for kills.

src/riak_core_handoff_manager.erl
massung: Done.

jonmeredith:
+1 merge.

The transfer code is a bit fragile at the moment around vnode deaths, but that's slated to be fixed when we add a supervision tree around the handoff processes.

PR #107, opened 2011-11-02
Fix default histogram max.

rustyio:
Revert histogram max to 5M (was typo'd to 500k in commit a6b679d5.


jonmeredith:
+1 merge - resolves failing eunit test for me.

PR #109, opened 2011-11-04
Replace incorrect {noreply, State} with continue(State) in riak_core_vnod

jonmeredith:
Replace incorrect {noreply, State} with continue(State) in riak_core_vnode.

Corrects the handle_info return when the vnode is forwarding and has deleted the state.

Fixes: bz://1227


kellymclaughlin:
+1 to merge.

PR #110, opened 2011-11-12
Add ability to throttle gossip rate

rzezeski:


rzezeski:
I verified that gossip is throttled with this patch in place.  I joined two devrel nodes with ring size of 1024 and monitored/graphed various stats.
- gossip received each minute
- transfers pending
- partitions moved per ring event
- as well as the vnode_manager, ring_events, and gossip mailbox sizes

I found that while there are higher spikes for the ring_events and gossip mailboxes the gossip msgs and vnode_manager mailbox were properly throttled.

rzezeski:
Here are the graphs.  The d1-b/a stands for dev1 before/after.  So for each set of graphs I first run without the patch (before) and then with the patch (after), back to back (except in 45-per-10 I accidentally ran with the patch first).  In between each run I recompiled and reset the ring.  Each title is `<# of gossip> per <seconds>`.

## 45 per 10

45 per 10

## 10 per 10

10 per 10

## 1 per 10

1 per 10

PR #111, opened 2011-11-15
`make docs` should now work

matehat:
Just a few fixes that make the `make docs` command work.


rzezeski:
Thanks for the fix @matehat.

PR #112, opened 2011-11-22
Fixed the foldl call that got missed in review and testing.

massung:
Was missing the initial accumulator argument.


Vagabond:
+1

PR #113, opened 2011-11-29
Handoff caching is now in an a-list instead of a dictionary.

massung:
Also added a suite of eunit tests.


russelldb:
Tests pass, coverage shows the right thing is tested, riak_control runs. All looks good to me +1 to merge.

PR #114, opened 2011-12-02
Az950 new default claim

rzezeski:
Change the default claim to version 2.


jonmeredith:
I think we should make the new_ring_claim code call into riak_core_claim for customers that have set their config to use the new claim stuff.

rzezeski:
This PR has been made obsolete by https://github.com/basho/riak_core/pull/126

PR #116, opened 2011-12-07
Register bucket_validator mods with riak_core

russelldb:
Call validation functions when bucket props are set


rustyio:
+1 merge

PR #117, opened 2011-12-08
Start of handoff manager rework: supervision tree

massung:
Created the following supervision tree for handoffs:

```
riak_core_handoff_sup (one_for_all)
 |
 + --> riak_core_handoff_manager (worker)
 |
 + --> riak_core_handoff_listener_sup (one_for_one)
 |      |
 |      + --> riak_core_handoff_listener (worker)
 |
 + --> riak_core_handoff_receiver_sup (simple_one_for_one)
 |      |
 |      + --> riak_core_handoff_receiver (worker)
 |
 + --> riak_core_handoff_sender_sup (simple_one_for_one)
        |
        + --> riak_core_handoff_sender (worker)
```

Updated riak_core_vnode.erl to start a sender via the sender supervisor and riak_core_handoff_listener to start a receiver via the receiver supervisor.


src/riak_core_handoff_listener_sup.erl
```diff
@@ -0,0 +1,39 @@
...
+
+%% beahvior functions
+-export([start_link/0,
+         init/1
+        ]).
+
+-define(CHILD(I,Type), {I,{I,start_link,[]},permanent,brutal_kill,Type,[I]}).
+
+%% begins the supervisor, init/1 will be called
+start_link () ->
```
rzezeski, line 32: What's up with the space before the args?  Might as well use no space to be consistent with rest of code.

rzezeski:
Overall the patch looks fine but I want to state for the record that using `one_for_all` on the handoff supervisor means that the handoff manager must not carelessly fail as it causes the entire handoff process to come down with it.

rzezeski:
+1 to merge

I tested both joining a node and restarting a node and in both cases handoff still worked.

PR #118, opened 2011-12-09
Added summary stats for vnode queue length.

jonmeredith:
Adds summary statistics on vnodes running and their queue lengths for each application registered with riak_core.


reiddraper:
+1 to merge

PR #120, opened 2011-12-13
Refactor riak_core vnode management (part 2)

jtuple:
Move forwarding and handoff decisions from individual vnode processes
and into the vnode manager. The vnode manager makes handoff and
forwarding decisions whenever the ring changes, and triggers vnode
state changes as appropriate.

Rewrite the logic by which per vnode handoff is marked as complete in
the ring. In particular, move the logic from riak_core_gossip and into
riak_core_vnode. The underlying ring changes are still serialized by
riak_core_ring_manager through the ring_trans function.

Change gossip throttling logic to trigger gossip whenever gossip tokens
reset, replacing the gossip interval approach.

Perform various tuning and additional minor changes that improve cluster
operation during a heavy gossip spike.

The app.config settings and notes from part 1 of the vnode manager rewrite still apply here:
https://github.com/basho/riak_core/pull/106

Changes tested manually along with basho_bench and basho_expect. Need to work on comprehensive integration tests for both part 1 and part 2 of this work. Will work on that alongside code review.

### DIFF
--- a/src/riak_core_ring_manager.erl
+++ b/src/riak_core_ring_manager.erl
@@ -247,6 +247,9 @@ handle_call({ring_trans, Fun, Args}, _From, State=#state{raw_ring=Ring}) ->
             prune_write_notify_ring(NewRing),
             riak_core_gossip:random_recursive_gossip(NewRing),
             {reply, {ok, NewRing}, State#state{raw_ring=NewRing}};
+        {set_only, NewRing} ->
+            prune_write_ring(NewRing),
+            {reply, {ok, NewRing}, State#state{raw_ring=NewRing}};
         {reconciled_ring, NewRing} ->
             prune_write_notify_ring(NewRing),
             riak_core_gossip:recursive_gossip(NewRing),
@@ -382,6 +385,12 @@ prune_write_notify_ring(Ring) ->
     do_write_ringfile(Ring),
     set_ring_global(Ring),
     riak_core_ring_events:ring_update(Ring).
+
+prune_write_ring(Ring) ->
+    riak_core_ring:check_tainted(Ring, "Error: Persisting tainted ring"),
+    riak_core_ring_manager:prune_ringfiles(),
+    do_write_ringfile(Ring),
+    set_ring_global(Ring).
 
 %% ===================================================================
 %% Unit tests
--- a/src/riak_core_ring_manager.erl
+++ b/src/riak_core_ring_manager.erl
@@ -380,10 +380,7 @@ set_ring_global(Ring) ->
 
 %% Persist a new ring file, set the global value and notify any listeners
 prune_write_notify_ring(Ring) ->
-    riak_core_ring:check_tainted(Ring, "Error: Persisting tainted ring"),
-    riak_core_ring_manager:prune_ringfiles(),
-    do_write_ringfile(Ring),
-    set_ring_global(Ring),
+    prune_write_ring(Ring),
     riak_core_ring_events:ring_update(Ring).
 
 prune_write_ring(Ring) ->
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -385,6 +385,11 @@ finish_handoff(State=#state{mod=Mod,
                                  forward=HN})
     end.
 
+handle_event({set_forwarding, undefined}, _StateName,
+             State=#state{modstate={deleted, _ModState}}) ->
+    %% The vnode must forward requests when in the deleted state, therefore
+    %% ignore requests to stop forwarding.
+    continue(State);
 handle_event({set_forwarding, ForwardTo}, _StateName, State) ->
     lager:debug("vnode fwd :: ~p/~p :: ~p -> ~p~n",
                 [State#state.mod, State#state.index, State#state.forward, ForwardTo]),
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -90,7 +90,7 @@ behaviour_info(_Other) ->
           handoff_node=none :: none | node(),
           handoff_pid :: pid(),
           pool_pid :: pid() | undefined,
-          event_timer :: reference(),
+          manager_event_timer :: reference(),
           inactivity_timeout}).
 
 start_link(Mod, Index, Forward) ->
@@ -114,7 +114,6 @@ send_command_after(Time, Request) ->
 
 
 init([Mod, Index, InitialInactivityTimeout, Forward]) ->
-    %%TODO: Should init args really be an array if it just gets Init?
     process_flag(trap_exit, true),
     {ModState, Props} = case Mod:init([Index]) of
         {ok, MS} -> {MS, []};
@@ -280,16 +279,16 @@ active(handoff_complete, State=#state{mod=Mod,
                                       index=Idx, 
                                       handoff_token=HT}) ->
     riak_core_handoff_manager:release_handoff_lock({Mod, Idx}, HT),
-    State2 = start_event_timer(handoff_complete, State),
+    State2 = start_manager_event_timer(handoff_complete, State),
     continue(State2);
 active({handoff_error, _Err, _Reason}, State=#state{mod=Mod, 
                                                     index=Idx, 
                                                     handoff_token=HT}) ->
     riak_core_handoff_manager:release_handoff_lock({Mod, Idx}, HT),
-    State2 = start_event_timer(handoff_error, State),
+    State2 = start_manager_event_timer(handoff_error, State),
     continue(State2);
-active({send_event, Event}, State) ->
-    State2 = start_event_timer(Event, State),
+active({send_manager_event, Event}, State) ->
+    State2 = start_manager_event_timer(Event, State),
     continue(State2);
 active({trigger_handoff, TargetNode}, State) ->
      maybe_handoff(State, TargetNode);
@@ -307,6 +306,12 @@ active(_Event, _From, State) ->
     Reply = ok,
     {reply, Reply, active, State, State#state.inactivity_timeout}.
 
+%% This code lives in riak_core_vnode rather than riak_core_vnode_manager
+%% because the ring_trans call is a synchronous call to the ring manager,
+%% and it is better to block an individual vnode rather than the vnode
+%% manager. Blocking the manager can impact all vnodes. This code is safe
+%% to execute on multiple parallel vnodes because of the synchronization
+%% afforded by having all ring changes go through the single ring manager.
 mark_handoff_complete(Idx, Prev, New, Mod) ->
     Result = riak_core_ring_manager:ring_trans(
       fun(Ring, _) ->
@@ -397,7 +402,7 @@ handle_event({set_forwarding, ForwardTo}, _StateName, State) ->
 handle_event(finish_handoff, _StateName, State=#state{mod=Mod,
                                                       modstate=ModState,
                                                       handoff_node=HN}) ->
-    stop_event_timer(State),
+    stop_manager_event_timer(State),
     case HN of
         none ->
             continue(State);
@@ -409,7 +414,7 @@ handle_event(cancel_handoff, _StateName, State=#state{mod=Mod,
                                                       modstate=ModState}) ->
     %% it would be nice to pass {Err, Reason} to the vnode but the 
     %% API doesn't currently allow for that.
-    stop_event_timer(State),
+    stop_manager_event_timer(State),
     case State#state.handoff_node of
         none ->
             continue(State);
@@ -522,8 +527,16 @@ code_change(_OldVsn, StateName, State, _Extra) ->
 maybe_handoff(State=#state{modstate={deleted, _}}, _TargetNode) ->
     %% Modstate has been deleted, waiting for unregistered.  No handoff.
     continue(State);
-maybe_handoff(State=#state{handoff_node=HN}, _TargetNode) when HN /= none ->
+maybe_handoff(State=#state{index=Idx, mod=Mod, handoff_node=HN},
+              TargetNode) when HN /= none ->
     %% Already handing off
+    case HN of
+        TargetNode ->
+            ok;
+        _ ->
+            lager:info("~s/~b: ignoring handoff request to ~p, already "
+                       "handing off to ~p", [Mod, Idx, TargetNode, HN])
+    end,
     continue(State);
 maybe_handoff(State=#state{mod=Mod, modstate=ModState}, TargetNode) ->
     case Mod:handoff_starting(TargetNode, ModState) of
@@ -583,13 +596,13 @@ reply(ignore, _Reply) ->
 %% messages sent by the vnode. Therefore, the vnode periodically resends event
 %% messages until an appropriate message is received back from the vnode
 %% manager. The event timer functions below implement this logic.
-start_event_timer(Event, State=#state{mod=Mod, index=Idx}) ->
+start_manager_event_timer(Event, State=#state{mod=Mod, index=Idx}) ->
     riak_core_vnode_manager:vnode_event(Mod, Idx, self(), Event),
-    stop_event_timer(State),
-    T2 = gen_fsm:send_event_after(30000, {send_event, Event}),
-    State#state{event_timer=T2}.
+    stop_manager_event_timer(State),
+    T2 = gen_fsm:send_event_after(30000, {send_manager_event, Event}),
+    State#state{manager_event_timer=T2}.
 
-stop_event_timer(#state{event_timer=undefined}) ->
+stop_manager_event_timer(#state{manager_event_timer=undefined}) ->
     ok;
-stop_event_timer(#state{event_timer=T}) ->
+stop_manager_event_timer(#state{manager_event_timer=T}) ->
     gen_fsm:cancel_timer(T).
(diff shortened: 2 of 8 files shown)


jonmeredith:
Although it didn't change in this p/r - handoff_manager:L114 should check for is_pid(Pid) - I've seen supervisors return undefined for the pid when things are in flux, which would result in calls to get_mod_index and monitors.
We3 should also probably log a warning the get_mod_idx call fails otherwise that'll just silently ignore possibly locked up vnodes.  I wonder if it would be better to kill them if that was the case so we were in a known state.

jonmeredith:
Some more trickle feedback

If the vnode manager is restarted after a crash, what mechanism makes it start doing primary ownership handoffs if there are no gossips?

Vnode state in a couple of places, lots of digging to find out what's going on
- should a vnode be able to forward and handoff at the same time?
- state is shared between two states lists and an ETS table.

Should all_nodes really be called all_vnodes?

If debugging an issue, I'd like to know the status of the running vnodes - which are
running, whether they are primary/fallback nodes, if they are in the process of
transferring ownership.  It isn't obvious to me how to request that information.

init/1 - update_forwarding relies on the ets table built in find_vnodes, should the
         order be different?

handle_cast({unregister}) 
- is it guaranteed the vnode is not forwarding or handing off?  What stops the forwarding/handoff list being stale?

handle_cast({ring_changed, Ring}) ->
- what is forced_ownership_handoff throttling and why is that needed on top of handoff concurrency?  why not start transferring all vnodes?

src/riak_core_ring_manager.erl
```diff
@@ -383,6 +386,12 @@ prune_write_notify_ring(Ring) ->
     set_ring_global(Ring),
     riak_core_ring_events:ring_update(Ring).
 
+prune_write_ring(Ring) ->
```
rzezeski, line 386: Change `prune_write_ring_notify` to use `prune_with_ring` to avoid duplication.

src/riak_core_vnode.erl
```diff
@@ -111,7 +113,7 @@ send_command_after(Time, Request) ->
     gen_fsm:send_event_after(Time, ?VNODE_REQ{request=Request}).
 
 
-init([Mod, Index, InitialInactivityTimeout]) ->
+init([Mod, Index, InitialInactivityTimeout, Forward]) ->
     %%TODO: Should init args really be an array if it just gets Init?
```
rzezeski: This TODO no longer applies.

src/riak_core_vnode_manager.erl
```diff
@@ -258,10 +269,142 @@ add_vnode_rec(I,  _State=#state{idxtab=T}) -> ets:insert(T,I).
 get_vnode(Idx, Mod, State) ->
     case idx2vnode(Idx, Mod, State) of
         no_match ->
-            {ok, Pid} = riak_core_vnode_sup:start_vnode(Mod, Idx),
+            ForwardTo = get_forward(Mod, Idx, State),
+            {ok, Pid} = riak_core_vnode_sup:start_vnode(Mod, Idx, ForwardTo),
```
rzezeski, line 285: Should this be starting a proxy vnode instead?

src/riak_core_vnode_manager.erl
```diff
@@ -258,10 +269,142 @@ add_vnode_rec(I,  _State=#state{idxtab=T}) -> ets:insert(T,I).
...
+    case orddict:find({Mod, Idx}, VNodes) of
+        error ->
+            ok;
+        {ok, Pid} ->
+            %% Changing the forwarding state of a vnode FSM that is
+            %% concurrently shutting down may result in an error,
+            %% which we can safely ignore. Same with a noproc error
+            %% occuring if the Pid for this Mod/Index is now stale.
+            try
+                riak_core_vnode:set_forwarding(Pid, ForwardTo),
```
rzezeski: There is no reason for a try/catch here.  set_forwarding uses `gen_fsm:send_all_state_event` which is async and returns `ok` regardless if pid is alive or not.

src/riak_core_vnode_manager.erl
```diff
@@ -258,10 +269,142 @@ add_vnode_rec(I,  _State=#state{idxtab=T}) -> ets:insert(T,I).
...
+                         Owner;
+                     {_, _, _} ->
+                         Me
+                 end,
+    case TargetNode of
+        Me ->
+            false;
+        _ ->
+            case app_for_vnode_module(Mod) of
+                undefined -> false;
```
rzezeski: Should the undefined case really be an exception/crash?  When do we have vnodes that don't live under an application? (I can think of one, testing).

src/riak_core_vnode_manager.erl
```diff
@@ -258,10 +269,142 @@ add_vnode_rec(I,  _State=#state{idxtab=T}) -> ets:insert(T,I).
...
+    Pid = get_vnode(Idx, Mod, State),
+    maybe_trigger_handoff(Mod, Idx, Pid, State).
+
+maybe_trigger_handoff(Mod, Idx, Pid, _State=#state{handoff=HO}) ->
+    case orddict:find({Mod, Idx}, HO) of
+        {ok, TargetNode} ->
+            try
+                riak_core_vnode:trigger_handoff(Pid, TargetNode),
+                ok
+            catch
```
rzezeski: No need for try/catch.  This call always returns `ok`.

src/riak_core_vnode_manager.erl
```diff
@@ -27,10 +27,9 @@
 -export([start_link/0, stop/0]).
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
--export([all_vnodes/0, all_vnodes/1, ring_changed/1, set_not_forwarding/2,
-         force_handoffs/0]).
+-export([all_vnodes/0, all_vnodes/1, ring_changed/1, force_handoffs/0]).
 -export([all_nodes/1, all_index_pid/1, get_vnode_pid/2, start_vnode/2,
```
rzezeski: Please get rid of `all_nodes` as it's the same as `all_vnodes`.  The only dep that uses it is riak_pipe and it's a simple one line change.  At least deprecate it via `@deprecate` or something.

jonmeredith:
More incremental review notes

I'd like to break the target node case statement in should_handoff into a function so we can trace with redbug.

core_vnode:L520 - can we add a lager:info message to say request handoff to node X but handoff already 
                      active to Y, iff X /= Y

start_event_timer - could we rename something better it is really vnode_mgr_event.

Is it worth checking for an event ref when receiving send_event messages, could have them in the mailbox after cancel.  are they truly idempotent?

Why does the vnode mark the handoff complete, not the vnode manager?  That is the only place the vnode now deals with the ring.

src/riak_core_vnode_manager.erl
```diff
@@ -258,10 +269,142 @@ add_vnode_rec(I,  _State=#state{idxtab=T}) -> ets:insert(T,I).
...
+    %% Inform vnodes that have changed forwarding status
+    Diff = NewForwarding -- Forwarding,
+    [change_forward(VNodes, Mod, Idx, ForwardTo)
+     || {{Mod, Idx}, ForwardTo} <- Diff],
+
+    State#state{forwarding=NewForwarding}.
+
+change_forward(VNodes, Mod, Idx, ForwardTo) ->
+    case orddict:find({Mod, Idx}, VNodes) of
+        error ->
```
rzezeski, line 335: This check is unnecessary given the `VNodes` passed in is the same list used to build the `Diff`.  It might even be nice to just include the `Pid` in the forward diff so you don't have to perform the lookup.

src/riak_core_vnode_manager.erl
```diff
@@ -258,10 +269,142 @@ add_vnode_rec(I,  _State=#state{idxtab=T}) -> ets:insert(T,I).
...
+check_forward(Ring, Mod, Index) ->
+    Node = node(),
+    case riak_core_ring:next_owner(Ring, Index, Mod) of
+        {Node, NextOwner, complete} ->
+            {{Mod, Index}, NextOwner};
+        _ ->
+            {{Mod, Index}, undefined}
+    end.
+
+update_forwarding(State) ->
```
rzezeski: It would be nice if funs like `update_forwarding` avoiding using functions like `get_raw_ring` and `change_forward` so that they are pure and can be more easily tested in isolation.

src/riak_core_vnode_manager.erl
jtuple, line 285: Nope. Proxies are always running and are started under a completely separated supervision tree. The vnode manager spawns and manages real vnode proceses. The proxy processes talk to the vnode manager to know who to forward requests to, the vnode manager never talks through proxies.

src/riak_core_vnode_manager.erl
jtuple: Yeah, good catch. Same for the other lines. Earlier code used sync calls in many places in the vnode manager, which needed the try/catch. Later refactoring changed nearly everything to be async to prevent one vnode from blocking the manager and therefore affecting other vnodes. I'll go through and remove all unnecessary try/catch.

src/riak_core_vnode_manager.erl
jtuple: Good question. This code snippet was around prior to 1.0, survived into 1.0, and is here now (code is just moving between modules). I can't imagine a case this is supposed to be undefined, but who knows. I like the idea of an exception, but am unsure if we should change the semantics this close to a code freeze. There may be a race between when a vnode is spawned and when the application modules list is updated or something, who knows. Probably something to clean-up post 1.1 when we have more time to test.

src/riak_core_vnode.erl
```diff
@@ -326,12 +307,58 @@ active(_Event, _From, State) ->
     Reply = ok,
     {reply, Reply, active, State, State#state.inactivity_timeout}.
 
+mark_handoff_complete(Idx, Prev, New, Mod) ->
+    Result = riak_core_ring_manager:ring_trans(
+      fun(Ring, _) ->
```
rzezeski, line 317: The new code to mark handoff complete on the ring no longer relies on the status on the new owner (i.e. `NewStatus`) being `valid`.  I.e. a ring transition will occur even if the new owner's status is `invalid`, `leaving` or `exiting`.  Not sure if it breaks the system as a whole but wanted to point it out.

jonmeredith:
Issues found joining 18 nodes concurrently

```
14:35:33.184 [error] Supervisor riak_core_vnode_proxy_sup had child {riak_kv_vnode,182687704666362864775460604089535377456991567872} started with riak_core_vnode_proxy:start_link(riak_kv_vnode, 182687704666362864775460604089535377456991567872) at <0.418.0> exit with reason reached_max_restart_intensity in context shutdown
14:35:33.140 [error] gen_server riak_core_vnode_manager terminated with reason: no match of right hand value {'EXIT',timeout} in riak_core_vnode_proxy:call/2
```

and 

```
14:38:28.144 [error] Supervisor riak_core_vnode_sup had child undefined started with {riak_core_vnode,start_link,undefined} at <0.10651.0> exit with reason no function clause matching riak_kv_vnode:handle_command({backend_callback,#Ref<18006.0.0.50690>,merge_check}, ignore, {deleted,{state,22835963083295358096932575511191922182123945984,false,riak_kv_bitcask_backend,{state,...},...}}) in context child_terminated
```

src/riak_core_vnode_proxy.erl
```diff
@@ -52,6 +52,10 @@ call(Name, Msg) ->
     {ok,Res} = (catch gen:call(Name, '$vnode_proxy_call', Msg)),
     Res.
 
+call(Name, Msg, Timeout) ->
+    {ok,Res} = (catch gen:call(Name, '$vnode_proxy_call', Msg, Timeout)),
```
rzezeski, line 56: What is the purpose of the catch?  Even if there is a problem you'll end up with a `badmatch`.  Why not just get rid of the catch?

rzezeski:
Why is vnode_proxy not just a gen_server?

src/riak_core_vnode_proxy.erl
Vagabond, line 56: Because gen:call() can exit(noproc), at the very least: https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen.erl#L158

rzezeski, line 56: Okay, that doesn't change the fact that it will then return something that is NOT `{ok, Res}`.  But I think a bigger issue is why not just use gen_server for the vnode proxy.

jtuple:
Still working on finishing up code changes based on the review feedback and answering remaining questions 1/1. But, to prevent back and forth discussion on the 'gen:call' point I'll chime in really quick.

So, first off. Why not gen_server? I already answered this during the part 1 review that added the proxy module, but that was on a Skype call rather than in writing so I'll repeat here to have a record.

As written, the proxy module is clear in it's intentions. It enters a receive loop, and forwards all messages other than the system messages and the explicitly tagged vnode_proxy message on to the vnode pid. The default is to forward, with only specific messages not being forwarded. Thus, `gen_fsm:send_event(vnode_proxy_pid)` works just as as well as `gen_fsm:send_event(vnode_pid)`.

If you take the time to look at `gen_server` in the OTP source, you realize you could safely emulate similar behavior by having the proxy be a `gen_server` and forwarding all messages that show up in `handle_info` on to the vnode pid, since `handle_info` will get the non-gen_server calls. But, doing so is a bit hackish in my opinion. You're also at the liberty of the gen_server implementation and hope it doesn't muck with any messages you really want forwarded to the vnode pid. It's also less clear. Having a gen_sever that we send gen_fsm events to will look a bit odd. The real problem is that gen_server also only works if we're forwarding onto a gen_fsm. If vnodes are ever changed to, say, a gen_server, then a gen_server-based proxy would no longer work since it would handle the calls rather than forward them on. The proxy as implemented could forward just as easily to a gen_server, gen_fsm, etc.

As far as the 'catch gen:call', Andrew has the right idea. All OTP behaviors that use 'gen' underneath (such as server, fsm, etc) will catch the gen:call. In gen_server, it catches the call, and will re-issue an exit if catches an EXIT message. I decided to keep the catch, but just fail on bad-match for the failure case. This turns things into a error with a stack trace rooted in the proxy module, rather than just an unexpected exit or error rooted in the gen module that really should be treated as opaque.

rzezeski:
Ah yea, now I remember us discussing this before when I first reviewed proxies.  Thanks for clarifying for a second time, Joe.

src/riak_core_vnode_manager.erl
jtuple: So, `all_nodes` only returns Pids while `all_vnodes` returns a `{Mod, Idx, Pid}` tuple. After looking, I realized that pipe actually doesn't even call `manager:all_nodes` it calls `master:all_nodes`. Of course, I removed `master:all_nodes` in part 2 of this rewrite, so I technically broke pipe on master when that merged in. Anyway, I went ahead and removed `all_nodes` from manager, and re-added it back to master -- however, the version in master now just calls `manager:all_vnodes` and extracts the pids. The master fun is also commented and marked as deprecated.

src/riak_core_vnode_manager.erl
jtuple, line 335: Actually, `VNodes` and `AllVNodes` aren't used at all in the calculation of the `Diff`. The newly refactored code makes that a little clearer. The `forwarding` list and `Diff` are generated from the ring state, and don't take into account running vnodes at all. So, it's entirely possible we have new forwarding information for a vnode we don't have a pid for in `VNodes`.

src/riak_core_vnode_manager.erl
jtuple: `update_forwarding/1` was just a wrapper around `update_forwarding/3` that grabbed the ring first. I've refactored things a bit so that `update_forwarding/1` is no longer used and was removed, and `update_forwarding/3` has always taken the ring as an argument. I also extracted the pure portion of `update_forwarding` out into `compute_forwarding` for testing purposes.

jtuple:
Pushed out various code changes and fixes. Hopefully this addresses all concerns noted. I fixed some bugs Jon found, fixed vnode manager initialization to be correct, and performed various code clean-up and refactoring.

I still need to work on exposing an information function that returns vnode status: mod/index/pid, fowarding to, scheduled to handoff to, etc.

Also, to answer a question.

Jon: "what is forced_ownership_handoff throttling and why is that needed on top of handoff concurrency? why not start transferring all vnodes?"

Prior to 1.0, we only triggered handoffs when inactive. In 1.0, we added the ability to handoff primary ownership under load, but it was throttled by the knob. This allows people who wanted to prevent handoff under load (or to tune it) to be able to play with this knob. Handoff concurrency doesn't distinguish between under load and idle. It's up to you if we care about this anymore, or if we just want to force handoff under load regardless and have no knob.

rzezeski:
In regards to throttling handoff I think we should do it both during quiescent states and under load.  Especially under load.  Furthermore I think we should have **ONE** config that determines that throttling, not two.  I vote `handoff_concurrency` is used for both.

rzezeski:
Turns out I misunderstood your last statement.  For anyone else reading, the forced_ownership_handoff is to throttle handoff **during load**.

+1 to merge.

jonmeredith:
+1 merge, reviewed code.  Ran 18 node cluster join/leaves under load (using the az896-vnode-manager-rewrite-part2-merge branch)

Have seen a couple of times where the handoff sender has exited with {error, enotcon}.
One run saw some notfounds during join but have been unable to reproduce.

This code will be tested heavily after code-freeze and has no known issues that will affect integration testing.

jtuple:
Rebased and merged:
https://github.com/basho/riak_core/commit/b30f83152090c668a6f5aa606c4554151363ddb0

PR #121, opened 2011-12-14
Synchronous VNode Startup (AZ1011)

rustyio:
Fixes regression where a service was declared "up" before all vnodes were started.

This pull request modifies `riak_core_ring_handler:ensure_vnodes_started/N` to add a longer wait for the application to start up, errors if the application fails, starts the vnodes for the application, and then declares the service up.

Depends upon the following pull requests:
- https://github.com/basho/riak_kv/pull/261
- https://github.com/basho/riak_pipe/pull/35
- https://github.com/basho/riak_search/pull/95

p.s. - Yes, this is a dirty, dirty hack. In my opinion, there are issues of complexity and unclear responsibilities in `riak_core_node_watcher`, `riak_core_node_watcher_events`, `riak_core_ring_events`, and `riak_core_ring_handler`. This is due for refactoring.

### DIFF
--- a/src/riak_core_ring_handler.erl
+++ b/src/riak_core_ring_handler.erl
@@ -120,7 +120,7 @@ ensure_vnodes_started({App,Mod}, Ring) ->
                        end,
 
                        %% Let the app finish starting...
-                       case wait_for_app(App, 1000, 100) of
+                       case riak_core:wait_for_application(App) of
                            ok ->
                                %% Start the vnodes.
                                [Mod:start_vnode(I) || I <- Startable],
@@ -157,15 +157,4 @@ startable_vnodes(Mod, Ring) ->
                 RO ->
                     [RO | riak_core_ring:my_indices(Ring)]
             end
-    end.
-
-wait_for_app(App, 0, _) ->
-    {error, {timeout, App}};
-wait_for_app(App, Count, Sleep) ->
-    case lists:keymember(App, 1, application:which_applications()) of
-        true ->
-            ok;
-        false ->
-            timer:sleep(Sleep),
-            wait_for_app(App, Count - 1, Sleep)
     end.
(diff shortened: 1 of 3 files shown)


jonmeredith:
Thanks Rusty.  I'd like to think through consequences of where we mark services up before we merge - I'll look when I can.

jonmeredith:
After Rusty humoring me over IM, a couple of items to resolve
1) Move the riak_core_ring_events:force_sync_update() call into riak_core:register so apps using core don't need to see its dirty laundry
2) Convert kv/pipe to use the newer riak_core:register() function instead of register_vnode_mod.

This change does couple the names of services to the name of the application which is undesirable, however pragmatism wins out until we can refactor.

src/riak_core_ring_handler.erl
```diff
@@ -114,16 +114,30 @@ ensure_vnodes_started({App,Mod}, Ring) ->
...
+
+                       %% Let the app finish starting...
+                       case riak_core:wait_for_application(App) of
+                           ok ->
+                               %% Start the vnodes.
+                               [Mod:start_vnode(I) || I <- Startable],
+
+                               %% Mark the service as up.
+                               SupName = list_to_atom(atom_to_list(App) ++ "_sup"),
+                               SupPid = erlang:whereis(SupName),
```
beerriot, line 130: I think the `_sup` suffix standard is a good practice, but another option is to use two undocumented functions:

```
{SupPid, _AppMod} = application_master:get_child(application_controller:get_master(riak_kv)).
```

Probably better to stick with how you have it already, but I found it too odd that OTP didn't expose something as simple as `application:whereis(AppName)` to resist digging.

jonmeredith:
+1 merge.  Tested under load and discovered issues with WM routes being registered before sup started.   Will submit a separate pull request for that.

PR #122, opened 2011-12-14
Handoff manager rework part 2

massung:
Make the handoff manager responsible for starting handoffs via a queue, limiting the number of active handoffs poassible at one time.

### DIFF
--- a/include/riak_core_handoff.hrl
+++ b/include/riak_core_handoff.hrl
@@ -1,3 +1,14 @@
+
+-record(handoff, { module        :: atom(),
+                   index         :: integer(),
+                   target_node   :: atom(),
+                   vnode_pid     :: pid(),
+                   sender_pid    :: pid(),
+                   restarts      :: integer(),
+                   timestamp     :: tuple(),
+                   status        :: any()
+                 }).
+
 -define(PT_MSG_INIT, 0).
 -define(PT_MSG_OBJ, 1).
 -define(PT_MSG_OLDSYNC, 2).
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -19,7 +19,7 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 -export([add_exclusion/2, get_exclusions/1, remove_exclusion/2]).
--export([add_outbound/3,clear_queue/0,all_handoffs/0]).
+-export([add_outbound/3, clear_queue/0, all_handoffs/0]).
 -record(state, {excl,handoffs,handoff_queue}).
 
 -include_lib("eunit/include/eunit.hrl").
@@ -33,8 +33,7 @@ init([]) ->
 
 %% enqueue a handoffs to happen at a later point
 add_outbound(Module, Idx, Node) ->
-    ok=gen_server:call(?MODULE,{add_outbound,Module,Idx,Node}),
-    gen_server:cast(?MODULE,process_handoff_queue).
+    gen_server:cast(?MODULE,{add_outbound,Module,Idx,Node}).
 
 %% cancel all pending handoffs
 clear_queue() ->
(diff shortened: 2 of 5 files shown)


massung:
There's (in the future when the vnode manager starts the handoff process instead of the vnode) a potential race condition if the vnode manager tells the handoff manager to start a handoff that's already in progress. If it's merely queued, that will get caught, but not if it's actively running.

jonmeredith:
L36/L40 - Calls into handoff manager for new code should be sync unless there is a good reason not to (ignore exclusions - hopefully that can go)

L34 - Add more comments about semantics of add_outbound, what happens in cases where transfer is non-existant, pending or active.

# TODO
- set_concurrency call not stubbed or implemented yet.  It should cancel newest handoffs down to the concurrency limit or start new ones up to the new concurrency limit and set the appenv in case of restart.
- add progress reporting callback for caller of add_outbound
- switch retrieval of vnode pid to use progress function provided by the add_outbound.  The vnode may have been restarted since handoff was added.  If that isn't ready, look the pid up from the vnode manager (but that will make EQC harder later).
- clear_queue() / cancel_pending_handoffs - currently just empties the queue, it does not cancel the active handoffs.  The original email called for a single clear function.  Using the same atom for the gen_server message is a good idea - less mental translation when looking at source code/debug traces when things are going wrong.

```
  %% Cancel all active and pending handoffs (calling their progress function) and 
  %% empty the list of handoffs
  riak_core_handoff_manager:clear() ->

```
- the code needs to check the Reason from 'DOWN' messages - I don't see what re-adds the handoff to the queue if it failed
- what prevents multiple handoffs to the same partition being started if handoff_concurrency > 1  getting restarted.
- the monitor code is fine - I was wrong.  You'll just get an immediate 'DOWN' message if the pid is dead.
- no status reporting stubbed or implemented in this module or riak_core:handoff_status()

# General structure

The way the handoff state is tracked is going to make doing status updates hard.  Here is what needs to be reported on status (slightly tweaked from the email I originally sent, in pigeon spec)

```
    %% Return the status of all handoffs.  Each handoff is either active or pending
    %% and provides a proplist that we can extend in the future.
    riak_core_handoff_manager:status() -> [{handoff(), direction(), handoff_status()}]

    handoff() -> {vnodemod(), index(), type(), node()}
    direction() -> inbound | outbound
    type() -> primary | fallback
    handoff_status() -> {active, [{start, now()}},
                                  {vnode_pid, pid()}
                                  {total_objects, NumObjects}, %% Perhaps processed objects is better
                                  {transport, transport()}} | %% Transfer is active
                        {pending, [{backoff_until, {utcdate(),utctime()}}, 
                                   {retries, int()}, 
                                   {last_failure, any()}]}
```

The way things are, you would have to add and remove entries from the queue structure and track them in the #state.handoffs, there also needs to be somewhere to store backoff intervals and number of retries.  I think it would be a much cleaner design if you just used a list of records to represent each handoff, with the ordering representing the state of the current queue.

```
  -record(handoff, {modidx :: {vnodemod(), index()}, node :: node(), type :: type(), direction :: direction(),
                                 retries :: integer(), backoff_until :: integer(),
                                 vnode_pid :: pid(), socket_pid:: pid(), status :: [{atom(), any()}]})
```

With this arrangement, you can do things like find make a list of pending outbound handoffs

```
     [H || H=#handoff{direction = outbound, socket_pid = undefined} <- Handoffs]
```

Find a handoff record

```
     get_handoff(ModIdx, Handoffs) ->
       lists:keysearch(ModIdx, #handoff.modidx, Handoffs). % returns {value, #handoff{}} or false
```

Update a handoff record 

```
     update_handoff(NewH, Handoffs) ->
        lists:keyreplace(NewH#handoff.modidx, #handoff.modidx, Handoffs, NewH)
```

Status becomes a list comprehension or fold over the state to reformat that information into the (hopefully future-proof) proplist format returned by status.  The internal structure of the handoff manager is clear to change.

Doing it this way is slightly less efficient as far as garbage/reductions, but I think it will clarify a lot of the code and remove risks of having state in one place, not the other.  You no longer need to call the supervisors to count types of active transfers and will have a consistent view of all handoffs.  With the handoff sender/receivers reporting status to the manager you will be able to get status updates without having to bother all of the (potentially very busy) handoff processes.

# Testability

This code is going to be heavily quickchecked.  For test setup/convenience/performance testing within a single VM is desirable.  This means being very conscious of registered names in the design.  There aren't too many in here, just for starting the sender.  Keeping the sender_sup pid in the state and using that when launching senders will make it easier to have multiple instances of the handoff manager running cleanly.  We'll have to do something similar for the handoff listener so it can spawn receivers from the appropriate listener.

PR #123, opened 2011-12-14
Setting Monitors on Vnode "Sender" Arguments (AZ996)

beerriot:
Vnodes are passed a `Sender` argument that is of the form `{fsm|server|raw,
X, Y}`. This `monitor/1` function unwraps that structure appropriately and
sets up a monitor on the pid that it names.

This is used by code in the `az996-listkeys-backpressure` branch of riak_kv.


kellymclaughlin:
+1 to merge

PR #125, opened 2011-12-15
Set default handoff_concurrency to 1

rzezeski:
We've found that in extreme load situations our default handoff concurrency,
paired with the fact that, currently, no incoming throttling is done, can cause
the node to become overloaded and the latency to spike.  Rather than have
a potentially harmful default Riak should err on side of safety.


jonmeredith:
+1 merge

PR #126, opened 2011-12-16
Az950 new default claim on 1.0

rzezeski:
Change the default claim to version 2.


src/riak_core_claim.erl
```diff
@@ -105,12 +114,30 @@ default_wants_claim(Ring0, Node) ->
...
     end.
 
-%% @spec default_choose_claim(riak_core_ring()) -> riak_core_ring()
-%% @doc Choose a partition at random.
-default_choose_claim(Ring) ->
-    default_choose_claim(Ring, node()).
+wants_claim_v2(Ring) ->
+    wants_claim_v2(Ring, node()).
+
+wants_claim_v2(Ring, Node) ->
```
Vagabond, line 120: What exactly is going on in this function, is it finding out if the node has less partitions than the average is expected to be? Can we add a comment?

rzezeski, line 120: Yes, this function literally returns how many partitions the given `Node` wants.  Yes, I'll add a comment.

Vagabond:
This all seems to work fine, +1 for this change.

PR #127, opened 2011-12-19
Have handoff sender monitor vnode and use helper process

rzezeski:
Now that the handoff sender and vnode are no longer linked it is unsafe
to use the master's `sync_command` call.  If the vnode fails the sync_command
call will never return and the sender will be stuck indefinitely unless it is
manually killed at the console.  Adding a timeout could fix this but would cause
other complications if you set it too low.

This patch gets around the issue by using a helper process to perform the fold.
The sender process then monitors the vnode and spawn_links the helper.  If
either the vnode or helper gets into trouble the sender will know about it.
In the happy path the sender will get the fold result back and exit cleanly.


jonmeredith:
+1 merge - tested in conjunction with pull request #120

PR #128, opened 2011-12-19
Handoff Manager now in charge of senders and receivers.

massung:
- Vnode and Listener go through the handoff manager to spawn senders and receivers.
- Handoff manager uses handoff_concurrency to determine if a new sender or receiver can spawn.
- Handoff lock has been removed. Just die and let vnode or listener try again.
- Ability to hard-kill handoffs.
- Handoffs have (unused) status information and know when they were started to track time.


src/riak_core_vnode.erl
```diff
@@ -519,18 +500,13 @@ start_handoff(State=#state{index=Idx, mod=Mod, modstate=ModState}, TargetNode) -
...
+                {ok,_Pid} ->
                     NewState = State#state{modstate=NewModState,
-                                           handoff_token=HandoffToken,
                                            handoff_node=TargetNode},
-                    {ok, HandoffPid} = riak_core_handoff_sender_sup:start_sender(TargetNode, Mod, Idx),
-                    riak_core_handoff_manager:add_handoff(Mod, Idx, TargetNode),
-                    continue(NewState#state{handoff_pid=HandoffPid})
+                    continue(NewState);
+                {error,_Reason} ->
+                    continue(State#state{modstate=NewModState})
```
massung, line 553: I need this verified. It appears to work just fine, but I'm unsure if this is the proper thing to do here.

src/riak_core_handoff_listener.erl
```diff
@@ -67,8 +67,12 @@ terminate(_Reason, _State) -> ok.
...
-    gen_tcp:controlling_process(Socket, Pid),
-    ok = riak_core_handoff_receiver:set_socket(Pid, Socket),
-    {ok, State}.
+    case riak_core_handoff_manager:add_inbound(SslOpts) of
+        {ok, Pid} ->
+            gen_tcp:controlling_process(Socket, Pid),
+            ok = riak_core_handoff_receiver:set_socket(Pid, Socket),
+            {ok, State};
+        {error, Reason} ->
+            {stop, Reason, State}
```
massung: Want to ensure this is the right message to return. Looks like this is called from gen_nb_server.erl, but want to be sure.

jonmeredith: No, that'll stop the listener.   This should still return {ok, State}.

Instead you should call gen_tcp:close(Socket) and probably let riak_core_stats know you've denied an incoming handoff by incrementing a counter.

jonmeredith: Calling the handoff sender to get the SSL opts also feels a bit dirty, it doesn't happen very often and could easily be refactored into a riak_core_handoff_util module and called directly by sender and receiver.

src/riak_core_handoff_manager.erl
```diff
@@ -53,38 +78,34 @@ remove_exclusion(Module, Index) ->
...
+    case receive_handoff(SSLOpts) of
+        {ok,Handoff} ->
+            {reply,{ok,Handoff#handoff.pid},State#state{handoffs=[Handoff|HS]}};
+        Error ->
+            {reply,Error,State}
+    end;
+handle_call(handoff_status,_From,State=#state{handoffs=HS}) ->
+    Senders=[H || H=#handoff{sender=Sender} <- HS, Sender =/= undefined],
+    Receivers=[H || H=#handoff{receiver=Recv} <- HS, Recv =/= undefined],
+    {reply, {ok, [{senders,Senders},{receivers,Receivers}]}, State};
```
jonmeredith: Is this handoff senders/receivers format different to the one from the email to better support riak control?  Here's how I thought we were going to do it - using a common format to make it easy for consumers of handoff information.  Splitting it may be ok too as there are only going to be inbound or outbound handoff.  I'm not sure if it's more common to have to combine the two lists in the tuple or filter them in a list comprehension.

Making it the same list seems like it would make it more likely to stay the same format as handoff manager is improved/extended.

```
    %% Return the status of all handoffs.  Each handoff is either active or pending
    %% and provides a proplist that we can extend in the future.
    riak_core_handoff_manager:status() -> [{handoff(), direction(), handoff_status()}]

    handoff() -> {vnodemod(), index(), type(), node()}
    direction() -> inbound | outbound
    type() -> primary | fallback
    handoff_status() -> {active, [{start, now()}},
                                  {vnode_pid, pid()}
                                  {total_objects, NumObjects}, %% Perhaps processed objects is better
                                  {transport, transport()}} | %% Transfer is active
                        {pending, [{backoff_until, {utcdate(),utctime()}}, 
                                   {retries, int()}, 
                                   {last_failure, any()}]}
```

src/riak_core_handoff_manager.erl
```diff
@@ -53,38 +78,34 @@ remove_exclusion(Module, Index) ->
...
-handle_cast({add_handoff, {Mod, Idx, Node}}, State=#state{handoffs=Hoffs}) ->
-    NewHoffs=[{{Mod,Idx},Node}|Hoffs],
-    {noreply, State#state{handoffs=NewHoffs}};
-handle_cast({remove_handoff,Hoff={_Mod,_Idx}}, State=#state{handoffs=Hoffs}) ->
-    NewHoffs=lists:keydelete(Hoff,1,Hoffs),
-    {noreply, State#state{handoffs=NewHoffs}};
+handle_call({add_outbound,Mod,Idx,Node,Pid},_From,State=#state{handoffs=HS}) ->
+    case send_handoff(Mod,Idx,Node,Pid) of
+        {ok,Handoff} ->
+            {reply,{ok,Handoff#handoff.pid},State#state{handoffs=[Handoff|HS]}};
```
jonmeredith: Updating the state this way would keep the handoff list most-recently started first.  I thought we'd do it the other way, but either works as long as it's well commented near the state record.  I'm not sure if we're more interested in youngest or oldest entries.

include/riak_core_handoff.hrl
```diff
@@ -3,3 +3,26 @@
 -define(PT_MSG_OLDSYNC, 2).
 -define(PT_MSG_SYNC, 3).
 -define(PT_MSG_CONFIGURE, 4).
+
+%% information about a sender process
```
jonmeredith: Unless the record is needed outside of handoff manager (can't think of anywhere it would be) I think it's better to include the record with the.erl module, so that if we ever have to patch the module it's only one file that's affected, not two and it doesn't create an unnecessary dependency between a source file and a header.

src/riak_core_handoff_manager.erl
massung: Each one has a timestamp for when it began, so I don't think ordering matters much as we can get the order at any time by sorting (again, if needed). If we end up having odd access patterns in the future where we are always getting the oldest ones to do something on them (e.g. show me all handoffs > 10 minutes old), then youngest-last makes sense. But currently, I'm unaware of any such access patterns.

src/riak_core_handoff_manager.erl
massung: So, currently the handoff() type is somewhat meaningless since receivers don't have that information. Hence splitting out senders into their own type and receivers into theirs. The sender record looks very similar to handoff(), except it has a little more information. The handoff() record I put together has timestamps and the sender/receiver pid. 

Active / pending makes no sense in this context since we aren't queueing anything up. If we want to save failed handoffs (any attempt that returned {error, max_concurrency}) in pending as showing things that will potentially (should?) happen in the future, I'd be cool with that, but right now since senders and receivers are different structures, I figured it made more sense to return those as opposed to active vs. pending, which, again, has no meaning currently.

massung: Some of the other things in that spec are implicit in the current setup (e.g. direction = inbound if receiver =/= undefined), but I can add them in anyway. type() I don't grok, though. Is that saying the receiver is a primary or fallback? the sender? something else? Again, though, not all the information is available for all the handoffs (inbound vs. outbound).

src/riak_core_handoff_listener.erl
massung: "Calling the handoff sender to get the SSL opts also feels a bit dirty"

I'm not following that. Where is that happening?

include/riak_core_handoff.hrl
massung: Right now I'm going with the assumption that it'll be needed by riak_control. I think this gets a lot cleaner, though, once the receiver has information about itself and the sender/receiver records can be unified.

src/riak_core_handoff_manager.erl
jonmeredith: Agreed, it isn't totally complete and some things may not make sense.  My hope was to make the format close to where we want to be to minimize changes down the road.

The type() is supposed to give an indication of the type of handoff - whether it's a handoff due to a change of ownership or if it's a handoff returning data from a fallback to the primary owner.  That information would have to be provided by the vnode mgr and it may not be complete enough to provide it yet.

src/riak_core_handoff_manager.erl
jonmeredith: Fair enough.  Maintaining the ordering was just an attempt to make it easy to pick next handoffs.  It probably makes the design fragile to maintenance, although with an update function that knows to preserve things it would be probably be fine.

include/riak_core_handoff.hrl
jonmeredith: I think it would be a bad idea for riak_control to directly use the record - the usage I was proposing was for tracking internal state of the handoff manager.  That's why the status() function format uses a proplist for most information.  The point you made above about type() being unknown at the moment was fair and it should probably have been in the proplist.  Maybe an improved version would have

```
riak_core_handoff_manager:status() -> [{handoff(), direction(), active | pending, Props}]
```

rather than the nested handoff_status tuple.

It would be good to be able to change how the handoff manager works with how it interfaces outside of riak_core.

jonmeredith:
Here's the output I got from status. 

```
{ok,[{{handoff,undefined,undefined,undefined,undefined}, inbound,active,undefined}, 
     {{handoff,undefined,undefined,undefined,undefined}, inbound,active,undefined},
     {{handoff,riak_kv_vnode, 205523667749658222872393179600727299639115513856, 'dev2@127.0.0.1',undefined}, outbound,active,undefined},
     {{handoff,riak_kv_vnode, 137015778499772148581595453067151533092743675904, 'dev3@127.0.0.1',undefined}, outbound,active,undefined},
     {{handoff,riak_kv_vnode, 68507889249886074290797726533575766546371837952, 'dev4@127.0.0.1',undefined}, outbound,active,undefined}]}
```

 I think it would be good if we can get the proplist in place for future extensibility as mentioned above.  Also I don't think the handoff atom is necessary in the handoff() element.  That was a pseudo '-spec' entry rather than a record definition.

```
handoff() -> {vnodemod(), index(), type(), node()}
riak_core_handoff_manager:status() -> [{handoff(), direction(), active | pending, Props}]
```

I noticed you also wrapped the result in an {ok, List} tuple.  I was originally just going to return a list to make it easier for people to fiddle around on the shell with, for example to get a list of indices being in outbound handoffs

```
[Idx || {{_VMod, Idx, _Type, _Node}, outbound, _State, _Props} <- riak_core_handoff_manager:status()].
```

Looking at that output, the type() probably doesn't belong in the handoff() and would be better in the Props list. It's better to return an empty list than undefined so that when we add more properties, code dealing with old nodes won't have to special case undefined.

```
(dev1@127.0.0.1) riak_core_handoff_manager:status().

[{{undefined,undefined,undefined}, inbound,active,[]}, 
 {{undefined,undefined,undefined}, inbound,active,[]},
 {{riak_kv_vnode, 205523667749658222872393179600727299639115513856, 'dev2@127.0.0.1'}, outbound,active, []},
 {{riak_kv_vnode, 137015778499772148581595453067151533092743675904, 'dev3@127.0.0.1'}, outbound,active, []},
 {{riak_kv_vnode, 68507889249886074290797726533575766546371837952, 'dev4@127.0.0.1'}, outbound,active, []}]
```

src/riak_core_handoff_manager.erl
```diff
@@ -15,34 +15,76 @@
...
+    gen_server:call(?MODULE,{add_inbound,SSLOpts}).
 
-all_handoffs() ->
-    gen_server:call(?MODULE, all_handoffs).
+handoff_status() ->
+    gen_server:call(?MODULE,handoff_status).
 
-get_handoffs(Idx) ->
-    gen_server:call(?MODULE, {get_handoffs, Idx}).
+set_concurrency(Limit) ->
```
jonmeredith, line 84: This should really go through the gen_server to avoid a race of and add_inbound/add_outbound executing on the server while the appenv setting runs on the caller thread.

src/riak_core_handoff_manager.erl
```diff
@@ -53,38 +95,43 @@ remove_exclusion(Module, Index) ->
...
+    case receive_handoff(SSLOpts) of
+        {ok,Handoff=#handoff_status{transport_pid=Receiver}} ->
+            {reply,{ok,Receiver},State#state{handoffs=[Handoff|HS]}};
+        Error ->
+            {reply,Error,State}
+    end;
+handle_call(handoff_status,_From,State=#state{handoffs=HS}) ->
+    Handoffs=[{H,D,active,S} || #handoff_status{ handoff=H,direction=D,status=S } <- HS],
+    {reply, {ok, Handoffs}, State};
+handle_call(kill_handoffs,_From,State=#state{handoffs=HS}) ->
```
jonmeredith: handoffs will immediately begin again after 10s unless the handoff concurrency is also reset.  The original idea was to implement through set_concurrency(0) which would prevent that problem.  I understand the concern you had about people not correctly resetting the value to the correct default.  We _could_ implement an enable/disable but you'd have to start tracking the state of that too, and it may confuse people who are used to being able to just set the handoff concurrency to control handoff.

src/riak_core_handoff_manager.erl
```diff
@@ -53,38 +95,43 @@ remove_exclusion(Module, Index) ->
...
+    [erlang:exit(Pid,kill) || #handoff_status{transport_pid=Pid} <- HS],
+    {reply, ok, State};
+handle_call(kill_old_handoffs,_From,State=#state{handoffs=HS}) ->
```
jonmeredith: Limiting handoff concurrency is supposed to remove the most recent handoffs.  It looks to me that you're keeping the 'Limit' most recent ones the way handoffs are added to state.

If you don't want an implicit ordering in the handoff state list, I'm fine with using a timestamp, but the list should be appropriately sorted before applying the limit.

src/riak_core_handoff_manager.erl
```diff
@@ -92,16 +139,81 @@ handle_cast({add_exclusion, {Mod, Idx}}, State=#state{excl=Excl}) ->
     riak_core_ring_events:ring_update(Ring),
     {noreply, State#state{excl=ordsets:add_element({Mod, Idx}, Excl)}}.
 
+
+handle_info({'DOWN',_Ref,process,Pid,_Reason},State=#state{handoffs=HS}) ->
+    NewHS=lists:keydelete(Pid,#handoff_status.transport_pid,HS),
```
jonmeredith: for non-normal reasons we should wrote a log:error() message to say the process died along with details about the transfer (direction, index, remote node etc).  It will be written to the crash log too, but it may make debugging easier.

The sender will have to be taught to gracefully exit if the handoff is dropped on them on connection to avoid spamming the logs.

src/riak_core_handoff_manager.erl
massung: I tried making the change suggested instead of using a #handoff{} record, and I agree that I like it more. Keeping.

src/riak_core_handoff_manager.erl
massung: Appending now. Still keeping the timestamp as it will be extremely useful for riak_control down the road.

src/riak_core_handoff_manager.erl
massung, line 84: Done.

src/riak_core_handoff_manager.erl
massung: Done. kill_handoffs/0 just sets concurrency to 0, which will kill them (and prevent restarting). I'm still partial to a flag instead of resetting to zero, but that can come later.

src/riak_core_handoff_manager.erl
massung: Done.

src/riak_core_handoff_manager.erl
```diff
@@ -92,16 +142,100 @@ handle_cast({add_exclusion, {Mod, Idx}}, State=#state{excl=Excl}) ->
     riak_core_ring_events:ring_update(Ring),
     {noreply, State#state{excl=ordsets:add_element({Mod, Idx}, Excl)}}.
 
+
+handle_info({'DOWN',_Ref,process,Pid,Reason},State=#state{handoffs=HS}) ->
+    case lists:keytake(Pid,#handoff_status.transport_pid,HS) of
+        {value,H=#handoff_status{handoff={Mod,Index,_},direction=Dir},NewHS} ->
+            if
```
jonmeredith: Handofffs cancelled due to max_concurrency limit are logged as an error, we did it to ourselves and it's expected so it should really log as an info message instead.

IMO it would be a bit more idiomatic to write it this way (as if statements make me furrow my brow with their fall-through true case)

```
case Reason of
  normal ->
     ok;
   max_concurrency ->
    lager:info(blah did what I expected),
    remove_handoff_and_error()
  _ ->
    lager:error("ohnoes it died", ~p\n reason)
    remove_handoff_and_error()
end
```

src/riak_core_vnode.erl
```diff
@@ -477,9 +502,9 @@ maybe_handoff(State=#state{mod=Mod, modstate=ModState}) ->
             continue(State)
     end.
 
-should_handoff(#state{handoff_node=HN}) when HN /= none ->
-    %% Already handing off
-    false;
+%should_handoff(#state{handoff_node=HN}) when HN /= none ->
```
rzezeski, line 505: If we don't need this code can we please remove it.

jonmeredith:
+1 merge.

tested adding/removing nodes under load.
tested killing handoff sender/receiver/manager

good to merge.

PR #130, opened 2011-12-24
Avoid use of ets:match() in hot code paths

slfritchie:
Profiling with DTrace on Solaris suggests that ets:match() is causing
a much higher than necessary lock contention rate within the Erlang
R14B03 and R14B04 VMs.  Eliminating ets:match() in the riak_core and
riak_kv applications is substantial when using Pthread locking instead
of Ericsson's hand-tuned locking (which is the default).  The results
are more ambiguous when the default locking strategy is used.

### DIFF
--- a/src/riak_core_node_watcher.erl
+++ b/src/riak_core_node_watcher.erl
@@ -65,13 +65,13 @@ node_down() ->
     gen_server:call(?MODULE, {node_status, down}, infinity).
 
 services() ->
-    ordsets:from_list([Service || [Service] <- ets:match(?MODULE, {{'_', '$1'}, '_'})]).
+    gen_server:call(?MODULE, services, infinity).
 
 services(Node) ->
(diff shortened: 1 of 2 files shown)


rzezeski:
Scott, not sure if you already know about this but a few releases ago there were concurrency options added to ETS: `read_concurrency` and `write_concurrency`.  The `read_concurrency` one showed me favorable gains on a highly concurrent (running on nodes with 16 cores), read heavy load of a public ETS table on an old project I did.  Might be worth a look for an additional boost if needed.

rzezeski:
Before I review this it needs to be rebased against master.  A lot has changed since Nov 17th.

slfritchie:
Ryan, the patch is destined both for master and 1.0 branches.
The code as-is is for the 1.0 branch, IIRC.  For master, I'll create
another pull request -- the diffs will be against a different module,
but otherwise the diffs will likely be unchanged.

rzezeski:
This PR has be made obsolete by https://github.com/basho/riak_core/pull/132

PR #131, opened 2011-12-30
AZ1047: Increase default gossip limit for 1.0.3

jtuple:
Increase the default gossip limit for Riak 1.0.3 to emulate having no limit at all (as was the case in 1.0.2 and earlier). This keeps behavior the same, but gives users the possibility to establish a lower limit if they run into gossip overload conditions. New default allows 1 million gossip messages per minute.


jonmeredith:
+1 merge

PR #132, opened 2012-01-03
Merge of slf-core-ets-match-replacement + hand-edits onto master.

slfritchie:
Avoid use of ets:match() in hot code paths

Profiling with DTrace on Solaris suggests that ets:match() is causing
a much higher than necessary lock contention rate within the Erlang
R14B03 and R14B04 VMs.  Eliminating ets:match() in the riak_core and
riak_kv applications is substantial when using Pthread locking instead
of Ericsson's hand-tuned locking (which is the default).  The results
are more ambiguous when the default locking strategy is used.

Hand-edits were required due to changes in master after the 1.0 branch
was created.

### DIFF
--- a/src/riak_core_node_watcher.erl
+++ b/src/riak_core_node_watcher.erl
@@ -68,10 +68,10 @@ services() ->
     gen_server:call(?MODULE, services, infinity).
 
 services(Node) ->
-    pubtab_get_services(Node).
+    internal_get_services(Node).
 
 nodes(Service) ->
-    pubtab_get_nodes(Service).
+    internal_get_nodes(Service).
 
 
 %% ===================================================================
@@ -99,7 +99,7 @@ init([]) ->
     net_kernel:monitor_nodes(true),
 
     %% Setup ETS table to track node status
-    ets:new(?MODULE, [protected, named_table]),
+    ets:new(?MODULE, [protected, read_concurrency, named_table]),
 
     {ok, schedule_broadcast(#state{})}.
 
@@ -317,8 +317,8 @@ node_down(Node, State) ->
 
 
 node_delete(Node) ->
-    Services = pubtab_get_services(Node),
-    [pubtab_delete(Node, Service) || Service <- Services],
+    Services = internal_get_services(Node),
+    [internal_delete(Node, Service) || Service <- Services],
     ets:delete(?MODULE, Node),
     Services.
 
@@ -327,15 +327,15 @@ node_update(Node, Services) ->
     %% know and determine what's changed (if anything).
     Now = riak_core_util:moment(),
     NewStatus = ordsets:from_list(Services),
-    OldStatus = ordsets:from_list(services(Node)),
+    OldStatus = ordsets:from_list(internal_get_services(Node)),
 
     Added     = ordsets:subtract(NewStatus, OldStatus),
     Deleted   = ordsets:subtract(OldStatus, NewStatus),
 
     %% Update ets table with changes; make sure to touch unchanged
     %% service with latest timestamp
-    [pubtab_delete(Node, Ss) || Ss <- Deleted],
-    [pubtab_insert(Node, Ss) || Ss <- Added],
+    [internal_delete(Node, Ss) || Ss <- Deleted],
+    [internal_insert(Node, Ss) || Ss <- Added],
 
     %% Keep track of the last time we recv'd data from a node
     ets:insert(?MODULE, {Node, Now}),
@@ -395,20 +395,20 @@ peers_update(NewPeers, State) ->
     %% Broadcast our current status to new peers
     broadcast(Added, State#state { peers = NewPeers }).
 
-pubtab_delete(Node, Service) ->
-    Svcs = pubtab_get_services(Node),
+internal_delete(Node, Service) ->
+    Svcs = internal_get_services(Node),
     ets:insert(?MODULE, {{by_node, Node}, Svcs -- [Service]}),
-    Nds = pubtab_get_nodes(Service),
+    Nds = internal_get_nodes(Service),
     ets:insert(?MODULE, {{by_service, Service}, Nds -- [Node]}).
 
-pubtab_insert(Node, Service) ->
+internal_insert(Node, Service) ->
     %% Remove Service & node before adding: avoid accidental duplicates
-    Svcs = pubtab_get_services(Node) -- [Service],
+    Svcs = internal_get_services(Node) -- [Service],
     ets:insert(?MODULE, {{by_node, Node}, [Service|Svcs]}),
-    Nds = pubtab_get_nodes(Service) -- [Node],
+    Nds = internal_get_nodes(Service) -- [Node],
     ets:insert(?MODULE, {{by_service, Service}, [Node|Nds]}).
 
-pubtab_get_services(Node) ->
+internal_get_services(Node) ->
     case ets:lookup(?MODULE, {by_node, Node}) of
         [{{by_node, Node}, Ss}] ->
             Ss;
@@ -416,7 +416,7 @@ pubtab_get_services(Node) ->
             []
     end.
 
-pubtab_get_nodes(Service) ->
+internal_get_nodes(Service) ->
     case ets:lookup(?MODULE, {by_service, Service}) of
         [{{by_service, Service}, Ns}] ->
             Ns;
--- a/src/riak_core_node_watcher.erl
+++ b/src/riak_core_node_watcher.erl
@@ -65,13 +65,13 @@ node_down() ->
     gen_server:call(?MODULE, {node_status, down}, infinity).
 
 services() ->
-    ordsets:from_list([Service || [Service] <- ets:match(?MODULE, {{'_', '$1'}, '_'})]).
+    gen_server:call(?MODULE, services, infinity).
 
 services(Node) ->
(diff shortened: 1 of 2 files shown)


src/riak_core_node_watcher.erl
```diff
@@ -65,13 +65,13 @@ node_down() ->
     gen_server:call(?MODULE, {node_status, down}, infinity).
 
 services() ->
-    ordsets:from_list([Service || [Service] <- ets:match(?MODULE, {{'_', '$1'}, '_'})]).
```
rzezeski, line 68: Why does this go thru the gen_server when the `services/1` does not?  Seems either they should both go thru the gen_server or directly access the table.

src/riak_core_node_watcher.erl
```diff
@@ -392,3 +395,31 @@ peers_update(NewPeers, State) ->
     %% Broadcast our current status to new peers
     broadcast(Added, State#state { peers = NewPeers }).
 
+pubtab_delete(Node, Service) ->
```
rzezeski: Can we please remove the "pubtab_" prefixes?  They don't make sense as the table is not public.  Even if it was this is an implementation detail and doesn't convey any useful information.

src/riak_core_node_watcher.erl
```diff
@@ -151,7 +151,11 @@ handle_call({node_status, Status}, _From, State) ->
              {Status, Status} -> %% noop
                  State
     end,
-    {reply, ok, update_avsn(S2)}.
+    {reply, ok, update_avsn(S2)};
+handle_call(services, _From, State) ->
+    Res = [Service || {{by_service, Service}, Nds} <- ets:tab2list(?MODULE),
```
rzezeski, line 156: Given this change is an optimization based on data seen "in the wild" then perhaps we should cache this info in the gen_server state?  From what I understand it is rarely updated but read often, so it seems to be a good candidate for caching.

src/riak_core_node_watcher.erl
```diff
@@ -392,3 +395,31 @@ peers_update(NewPeers, State) ->
...
+
+pubtab_insert(Node, Service) ->
+    %% Remove Service & node before adding: avoid accidental duplicates
+    Svcs = pubtab_get_services(Node) -- [Service],
+    ets:insert(?MODULE, {{by_node, Node}, [Service|Svcs]}),
+    Nds = pubtab_get_nodes(Service) -- [Node],
+    ets:insert(?MODULE, {{by_service, Service}, [Node|Nds]}).
+
+pubtab_get_services(Node) ->
+    case ets:lookup(?MODULE, {by_node, Node}) of
```
rzezeski, line 412: In the `handle_call(services...` callback you use tab2list and list comprehension and here you use `ets:lookup`.  Since this change is related to performance I wanted to point that out in case there is some perf difference between the two.

src/riak_core_node_watcher.erl
```diff
@@ -327,12 +331,11 @@ node_update(Node, Services) ->
```
rzezeski: It's not visible in this diff but I think `OldStatus` should use the private/internal function to get the services for a node rather than the API in case that API ever becomes a gen_server call.  Otherwise you'll have a deadlock.

src/riak_core_node_watcher.erl
```diff
@@ -327,12 +331,11 @@ node_update(Node, Services) ->
 
     Added     = ordsets:subtract(NewStatus, OldStatus),
     Deleted   = ordsets:subtract(OldStatus, NewStatus),
-    Unchanged = ordsets:intersection(NewStatus, OldStatus),
 
     %% Update ets table with changes; make sure to touch unchanged
     %% service with latest timestamp
-    [ets:delete(?MODULE, {Node, Ss}) || Ss <- Deleted],
-    ets:insert(?MODULE, [{{Node, Ss}, Now} || Ss <- Added ++ Unchanged]),
+    [pubtab_delete(Node, Ss) || Ss <- Deleted],
```
rzezeski: This call to pubtab_delete will make a call to get the services for a node but this was already done above and bound to `OldStatus`.  If this is for performance then why not pass that data down rather than access the ETS table again?

src/riak_core_node_watcher.erl
```diff
@@ -392,3 +395,31 @@ peers_update(NewPeers, State) ->
     %% Broadcast our current status to new peers
     broadcast(Added, State#state { peers = NewPeers }).
 
+pubtab_delete(Node, Service) ->
+    Svcs = pubtab_get_services(Node),
+    ets:insert(?MODULE, {{by_node, Node}, Svcs -- [Service]}),
+    Nds = pubtab_get_nodes(Service),
+    ets:insert(?MODULE, {{by_service, Service}, Nds -- [Node]}).
+
+pubtab_insert(Node, Service) ->
```
rzezeski: Given that we are using an ETS type of `set` we could take advantage of that and get rid of the internal delete function.  The `ets:insert` on a set will overwrite whatever the previous value was.

rzezeski:
There is also a `read_concurrency` option for ETS tables which might be useful in this scenario.

src/riak_core_node_watcher.erl
slfritchie, line 68: services/0 is called much less frequently than services/1.  Everyone ought to be
calling the gen_server process, but due to call frequency, services/1 uses a hack
for speed.

src/riak_core_node_watcher.erl
slfritchie, line 156: I don't believe that the call frequency is worth the bother of adding a cache and risking
screwing up the cache sometime later.  freq is _much_ less than 1/sec.

src/riak_core_node_watcher.erl
slfritchie: ok, patching

src/riak_core_node_watcher.erl
slfritchie: This function is only called when a riak_core service goes up or down, which is typically
much less than 1/sec ... much unlike services/1 which can be called many thouands/sec.

src/riak_core_node_watcher.erl
slfritchie: renamed "internal_".

src/riak_core_node_watcher.erl
slfritchie: I was rewriting what was there, not squeezing max efficiency because it's not a
hot-hot-hot code path.

src/riak_core_node_watcher.erl
slfritchie, line 412: `ets:lookup()` is enormously faster than `ets:tab2list()`.  The usage is a matter of
being called thousands/sec vs. 1/minute or less.

slfritchie:
Added `read_concurrency` ... odd that it isn't the default for a protected table.  {sigh}

slfritchie:
I've pooooshed edits to GitHub, please take another look?

src/riak_core_node_watcher.erl
rzezeski, line 156: Interesting, well if this isn't a hot path in the code then why did we need to perform the optimization in the first place?

rzezeski, line 156: Never mind, reading the other comments it seems this is _not_ the hot path.

rzezeski:
Code looks good, thanks Scott.  Tomorrow I'll run some benchmarks.  I doubt I'll be able to reproduce Voxer's load but at least I can sanity check that performance doesn't degrade.

rzezeski:
**+1 to merge.**

After using some custom code to sample `services/1` under join/leave load it appears to have reduced runtime.  Below are histograms generated via the `dbg` module along with the use of our `slide` module.  Below are `{Samples, Mean, {Median, 95th, 99th, Max}}` stats generated from slide for before and after this patch.

## Before

```
trace for {riak_core_node_watcher,services,1} finished: {1035,18,
                                                         {127,241,423,1849}}

trace for {riak_core_node_watcher,services,1} finished: {839,24,
                                                         {127,242,422,1749}}

trace for {riak_core_node_watcher,services,1} finished: {892,49,
                                                         {130,247,1006,3841}}

trace for {riak_core_node_watcher,services,1} finished: {412,11,
                                                         {125,238,248,266}}

trace for {riak_core_node_watcher,services,1} finished: {782,25,
                                                         {127,242,413,1412}}

trace for {riak_core_node_watcher,services,1} finished: {889,53,
                                                         {131,249,1009,4526}}

trace for {riak_core_node_watcher,services,1} finished: {978,17,
                                                     {126,239,249,2684}}
```

## After

```
trace for {riak_core_node_watcher,services,1} finished: {974,12,
                                                         {126,239,249,1401}}

trace for {riak_core_node_watcher,services,1} finished: {372,9,
                                                         {126,239,249,836}}

trace for {riak_core_node_watcher,services,1} finished: {414,6,
                                                         {125,238,248,313}}

trace for {riak_core_node_watcher,services,1} finished: {588,5,
                                                         {125,185,185,185}}

trace for {riak_core_node_watcher,services,1} finished: {385,3,
                                                         {125,127,127,127}}

trace for {riak_core_node_watcher,services,1} finished: {753,5,
                                                         {125,238,248,316}}

trace for {riak_core_node_watcher,services,1} finished: {717,3,
                                                         {125,237,247,276}}

trace for {riak_core_node_watcher,services,1} finished: {925,8,
                                                         {125,238,248,2620}}

trace for {riak_core_node_watcher,services,1} finished: {981,9,
                                                     {126,239,249,787}}
```

PR #134, opened 2012-01-06
AZ1057: Fix claim algorithm v2 to work when in legacy cluster mode

jtuple:
Change wants_claim_v2 and choose_claim_v2 to upgrade the input ring. This allows the algorithms to work properly when in legacy mode, where the old ring format may be passed in rather than the new ring format. This allows for mixed 0.14.2 and 1.1 clusters, such as during a rolling upgrade.


jtuple:
This is the 1.1 variant to the already reviewed and merged 1.0.3 pull-request:
https://github.com/basho/riak_core/pull/133

PR #136, opened 2012-01-10
Change node to use claim_v1 when in legacy mode

jtuple:
The new `claim_v2` algorithms introduced as default for 1.0.3 make assumptions about the ring that do not hold for legacy rings. The claim function may therefore crash, preventing a node from joining a cluster. Therefore, change `default_claim` to use `claim_v1` when given a legacy ring.

This ensure that a cluster can properly perform a rolling upgrade from 0.14.2 to 1.0.3. With this change, the old claim algorithm is still used during the upgrade, and then the new algorithm is used after all nodes have been upgraded to 1.0.3. Fixes observed basho_expect rolling upgrade failure.

### DIFF
--- a/src/riak_core_claim.erl
+++ b/src/riak_core_claim.erl
@@ -81,7 +81,12 @@ default_choose_claim(Ring) ->
     default_choose_claim(Ring, node()).
 
 default_choose_claim(Ring, Node) ->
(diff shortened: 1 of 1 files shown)


jonmeredith:
+1 merge.

PR #139, opened 2012-01-31
Make sure duplicates don't exist in the registered modules

Vagabond:
Allowing duplicates was breaking the pipe and kv unit tests.


jtuple:
+1 merge

PR #140, opened 2012-02-01
List lager as an application riak_core depends on

Vagabond:


Vagabond:
This helps force reltool to start lager earlier in the boot process.

jonmeredith:
+1 merge

PR #141, opened 2012-02-06
Add eunit test for private_dir/0 (and cleanup some whitespace).

seancribbs:
This is a follow-up to AZ1072, adding an eunit test for the private_dir/0 function.

### DIFF
--- a/src/riak_core_coverage_fsm.erl
+++ b/src/riak_core_coverage_fsm.erl
@@ -63,7 +63,7 @@
 -module(riak_core_coverage_fsm).
 -author('Kelly McLaughlin <kelly@basho.com>').
 
--include_lib("riak_core_vnode.hrl").
+-include("riak_core_vnode.hrl").
 
 -behaviour(gen_fsm).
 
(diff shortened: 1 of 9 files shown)


jaredmorrow:
Doesn't compile on builders:  http://buildbot-master.den.basho:8010/builders/test-riak_core/builds/55/steps/shell_3/logs/stdio

To reproduce, clone into a name that isn't the <reponame>:   git clone git@github.com:basho/riak_core builder

seancribbs:
This comes from an `-include_lib` that is broken, see https://gist.github.com/1753157. Needs to be fixed, but not related to this PR.

seancribbs:
@jaredmorrow Fixed all `include_lib` statements of `riak_core` header files so they use `include` instead. Builds properly now on a checkout to a differently-named directory.

jaredmorrow:
+1 compiles cleanly now on buildbot http://buildbot-master.den.basho:8010/builders/test-riak_core/builds/56

Issue #144, opened 2012-02-24
Map-reduce jobs fail during rolling upgrade to 1.1


jtuple:
Map reduce jobs fail against a mixed 1.1/1.0 cluster or 1.1/0.14 cluster. Mixed clusters are common during rolling upgrades, where nodes are upgraded one at a time until the entire cluster is at the new version (1.1 in this case).

In addition to the jobs failing, the Riak 1.1 nodes will print several instances of the following error and then eventually shutdown:

```
2012-02-22 00:56:16.429 [error] <0.1615.0>
gen_server riak_pipe_vnode_master terminated with reason: no function clause matching
riak_core_vnode_master:handle_call(
  {return_vnode,
    {riak_vnode_req_v1,479555224749202520035584085735030365824602865664,
      {raw,#Ref<6584.0.7263.108930>,...},...}},
  {<6604.7292.421>,#Ref<6604.0.6697.250987>},
  {state,undefined,undefined,riak_pipe_vnode,undefined})

2012-02-22 00:56:16.431 [error] <0.1615.0> CRASH REPORT
Process riak_pipe_vnode_master with 0 neighbours crashed with reason:
no function clause matching riak_core_vnode_master:handle_call(
  {return_vnode,
    {riak_vnode_req_v1,479555224749202520035584085735030365824602865664,
      {raw,#Ref<6584.0.7263.108930>,...},...}},
  {<6604.7292.421>,#Ref<6604.0.6697.250987>},
  {state,undefined,undefined,riak_pipe_vnode,undefined})
```

The error arises from Riak 1.1 no longer implementing `riak_core_vnode_master:handle_call({return_vnode, ...)`. Yet, pre-1.1 nodes will still send messages to Riak 1.1 nodes in a mixed cluster that must be handled by this now-missing clause.

jtuple:
There is also a related error. In addition to the above error, it is also possible to see the following errors during a map-reduce query to a mixed 1.1 and pre-1.1 cluster.

```
2012-02-24 14:58:04.501 [error] <0.26014.0> gen_fsm <0.26014.0> in
state wait_for_input terminated with reason: processing_error

2012-02-24 14:58:04.502 [error] <0.26014.0> CRASH REPORT Process
<0.26014.0> with 0 neighbours crashed with reason:
{processing_error,[{gen_fsm,terminate,7},{proc_lib,init_p_do_apply,3}]}

2012-02-24 14:58:04.503 [error] <0.321.0> Supervisor
[...]

PR #145, opened 2012-02-24
Fix map_reduce during rolling upgrade to 1.1.

jtuple:
Resolve issue #144.

Change riak_core_vnode_master to still handle return_vnode messages
given that pre-1.1 nodes may still send those messages to a 1.1 node.

Add legacy routing logic to riak_core_vnode_master:command_return_vnode
in order to properly send messages to pre-1.0 nodes that do not have
vnode proxy processes.

To verify: generate a mixed Riak 1.0 and Riak 1.1 cluster. Without these changes, issuing a map_reduce job against the cluster will fail. With these changes, it will pass.


jonmeredith:
+1 merge

Tested pipe with combinations on 1.0.3 and patched 1.1.0 and it worked with mixed,legacy=true 1.1.0,legacy=true and 1.1.0,legacy=false.

PR #146, opened 2012-03-06
Strawman commit to check that core stats is the right direction

russelldb:
Just for Jon(anyone?) to check that the direction of new stats work is ok.

### DIFF
--- a/ebin/riak_core.app
+++ b/ebin/riak_core.app
@@ -67,7 +67,14 @@
              slide,
              spiraltime,
              supervisor_pre_r14b04,
-             vclock
+             vclock,
+             riak_core_metric,
+             riak_core_metric_sup,
+             riak_core_metric_proc,
+	     riak_core_metric_counter,
+             riak_core_metric_histogram,
+	     riak_core_metric_meter,
+	     riak_core_metric_duration
             ]},
   {registered, []},
   {applications, [
--- a/ebin/riak_core.app
+++ b/ebin/riak_core.app
@@ -71,10 +71,10 @@
              riak_core_metric,
              riak_core_metric_sup,
              riak_core_metric_proc,
-	     riak_core_metric_counter,
+             riak_core_metric_counter,
              riak_core_metric_histogram,
-	     riak_core_metric_meter,
-	     riak_core_metric_duration
+             riak_core_metric_meter,
+             riak_core_metric_duration
             ]},
   {registered, []},
   {applications, [
--- a/src/riak_core_metric_proc.erl
+++ b/src/riak_core_metric_proc.erl
@@ -30,7 +30,7 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
--record(state, {mod, mod_state, display, description}).
+-record(state, {name, mod, mod_state, presentation, description}).
 
 %%%===================================================================
 %%% API
@@ -38,7 +38,7 @@
 
 start_link(App, Name, Args) ->
     RegName = riak_core_metric:regname(App, Name),
-    gen_server:start_link({local, RegName}, ?MODULE, Args, []).
+    gen_server:start_link({local, RegName}, ?MODULE, [{name, Name}|Args], []).
 
 update(App, Name, Args) ->
     RegName = riak_core_metric:regname(App, Name),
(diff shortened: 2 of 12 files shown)


russelldb:
I speel gud

src/riak_core_metric_proc.erl
```diff
@@ -0,0 +1,85 @@
...
+terminate(_Reason, _State) ->
+    ok.
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+mod_from_shortname(ShortName) ->
```
russelldb, line 95: Guess this is no good, should use the binary technique?

ebin/riak_core.app
```diff
@@ -67,7 +67,14 @@
              slide,
              spiraltime,
              supervisor_pre_r14b04,
-             vclock
+             vclock,
+             riak_core_metric,
+             riak_core_metric_sup,
+             riak_core_metric_proc,
+	     riak_core_metric_counter,
```
jonmeredith: is there a space/tabs problem here?

src/riak_core_metric_proc.erl
jonmeredith, line 95: If it is only called on init then performance doesn't matter - clean code wins.

jonmeredith:
Looks like a nice incremental start to me.  

I see what you mean about the duration stats - I'm not aware of any others like that outside of riak_core, it was probably just a convenience to avoid tracking convergence delay in the gossip code.

There are some things to ponder going forward, some stats and some the way that we want to organize things in riak_core in general.

The general thing is that stats are registered when an application starts up, but there is no way to unregister them or change them as an app goes down (or is restarted).  For stats it isn't a particularly big problem, but it is for things like web machine where you have register it register resources with riak_core_web that should be taken down when the app is shutting down.  As we work on the other refactors for riak_core for command line support etc it may be worth collecting the things that we're stuffing into the app env and making them first order riak_core processes that can do things like monitor the application that registered them so we can shutdown cleanly.

One thing to think about is how the stats are presented.  The riak_core:get_stats() and riak_kv:get_stats() calls are used to drive the stats webmachine resource, command line status tool, JMX and SNMP stats apps.  I wonder if you could add some presentation information with the stats description that would let you customize what you saw in each view.

For example, given a stats type

```
     {converge_delay, [{type, duration}]},
```

You have to do work to get each of the different stats out of converge delay.

```
      {converge_delay_min,  proplists:get_value(min,  CDelay)},
      {converge_delay_max,  proplists:get_value(max, CDelay)},
      {converge_delay_mean, proplists:get_value(mean, CDelay)},
      {converge_delay_last, proplists:get_value(last, CDelay)},
```

You could also provide presentation information with the stat spec

```
     {converge_delay, [{type, duration},
                       {group, gossip},
                       {presentation,
                              [{legacy, [min, max, mean, last]}]}
```

and use that to dynamically generate the names instead.  It's only a shortcut
if the existing stats are regular enough to build rules based on it.

The advantage I can see for doing that is if we wanted to start providing
richer interfaces for digging through stats so you could easily produce which groups
of stats we have, or we could start tagging what to display based on level of stats
requested.  

I'm hopeful we'll be able to make the JMX/SNMP app be data driven by the stats
specifications.  For example if we know that converge_delay is a 'duration' then we
can define an OID subtree to describe that and just provide something like
{snmp, coreConvergenceDelay} in the stats spec then add to the MIB by just allocating
the root oid for coreGossipDuration

```
coreConvergenceDelay OBJECT-TYPE
    SYNTAX RiakDuration
    ACCESS read-only
    STATUS mandatory
    DESCRIPTION
        "Length in millseconds for ring to converge."
    ::= { riak core 1 }
```

That's probably digging a bit too deep for now, but it would be really nice
if there was some simple information that could be added to the stat specification
that enables rewriting the riak_kv_stat/riak_core_stat:get_stats() as

```
get_stats() ->
   riak_core_stats:produce(legacy).
```

Don't worry about replicating internal functions within riak_core_stat/riak_kv_stat
if they just get combined for the only purpose of generating the output.  Having
a way to group stats is very useful, but it doesn't have to be the ad-hoc way we
do it right now.

Oh, and if you'd like a good test to see if you've lost any stats out of k/v, I
recommend turning on JMX as it is very picky if the stats aren't present.

ebin/riak_core.app
russelldb: yes /me shakes fist at vi

russelldb:
WRT presentation.

Yes. That was next, honest. riak_core_metric:value/2 exists and arg1 is the display information. Likewise there is a 'display' element in the riak_core_metric_proc state. This is to be passed to the metric impl when value is called. I'll make that the next bit of work so you can see what I had in mind and see if it is the same thing that you have in mind.

jonmeredith:
heh, excellent - glad we're on the same page.  I only mentioned it in terms of sequencing - maybe it would save you some effort to get it working with kv/repl.

russelldb:
Yes, good point. I am adding display code to riak_core first, then if your happy I'll move KV to this new way of gathering stats.

jonmeredith:
I'm happy. Have at it.

russelldb:
Added another commit that shows 'display' code working. I'm going to port riak_kv_stat to flesh out the display code now (assuming you're happy thus far.)

I'll add tests and clean up and issue a proper PR after that.

russelldb:
Closing this and tidying up code for a real PR.

PR #149, opened 2012-03-14
Possibly a typo in two locations

fygrave:
riak_core doesn't build without specifying full package name (Erlang R14A (erts-5.8))


seancribbs:
The `erlang` module's functions are auto-imported. It is only a problem if a local function in the module has the same name.

fygrave:
aha.. makes sense.

PR #148, opened 2012-03-23
Refactor riak stat collection and display

russelldb:
The current stats implementation has a number of problems
- structure wise the modules that implement stats calculation are in
  riak_core, and modules that use them are in riak_core_stats, riak_kv_stats and
  riak_repl_stats
- riak_kv_stat process has been a bottleneck in the past due to the amount of
  garbage it creates and how much data goes through it.
- SNMP/JMX apps are tied to riak_kv_stat implementation and a pain to extend.

Refactor stats so that riak_core will be responsible for maintaining all state for the statistics with an erlang process-per-stat


ebin/riak_core.app
```diff
@@ -67,7 +67,14 @@
              slide,
              spiraltime,
              supervisor_pre_r14b04,
-             vclock
```
jonmeredith, line 70: It's pure pedantry - but this is a nicely sorted list at the moment, why not add after riak_core_handoff_sup?

src/riak_core.erl
```diff
@@ -273,6 +274,12 @@ bucket_validators() ->
         {ok, Mods} -> Mods
     end.
 
+stat_specs() ->
+    case application:get_env(riak_core, stat_specs) of
```
jonmeredith: bit of code duplication - could just use app_helper:get_env(riak_core, stat_specs, []).

src/riak_core_stat.erl
```diff
@@ -19,191 +19,88 @@
...
-    setelement(Elt, State, element(Elt, State) + Amount).
-
-%% @doc Add a value to a set, updating the cumulative min/max/mean
-update_cumulative(Elt, Value, State) ->
-    #cuml{count=N, min=Min, max=Max, mean=Mean} = element(Elt, State),
-    Min2 = erlang:min(Min, Value),
-    Max2 = erlang:max(Max, Value),
-    Mean2 = ((N * Mean) + Value) div (N+1),
-    Stat2 = #cuml{count=N+1, min=Min2, max=Max2, mean=Mean2, last=Value},
-    setelement(Elt, State, Stat2).
```
jonmeredith, line 173: Why not just directly implement this stuff in update/1 and call slide:moment() where required?  It probably doesn't matter for low frequency gossip stats but for any high frequency calls it means calling os:timestamp() which I think maps to a gettimeofday call to the operating system.

src/riak_core_sup.erl
```diff
@@ -70,6 +69,7 @@ init([]) ->
                   ?CHILD(riak_core_node_watcher, worker),
                   ?CHILD(riak_core_vnode_manager, worker),
                   ?CHILD(riak_core_gossip, worker),
+                  ?CHILD(riak_core_metric_sup, supervisor),
```
jonmeredith: I wonder if there are any interesting races with core stats.  Until the core stats are registered on https://github.com/basho/riak_core/pull/148/files#L2R94 they can't be reported, however gossip is up and running.

I think stats should be started pretty early in the initialization sequence.   

You could add an alternate version of start_link that takes a module name and does something like

```
start_link(App, Mod) ->
  case start_link() of
    {ok, Pid} = Res->
      ok = start_stats(App, Mod),
      Res;
    ER ->
      ER
  end.
```

Then remove the register call at https://github.com/basho/riak_core/pull/148/files#L2R94 and change the riak_core_sup call to pass arguments

```
-define(CHILD(I, Type, Args, Timeout), {I, {I, start_link, Args}, permanent, Timeout, Type, [I]}).

?CHILD(riak_core_metric_sup, supervisor, [riak_core, riak_core_stat]).
```

There may be better ways to structure it, but I think that would resolve the race.

src/riak_core_metric_counter.erl
```diff
@@ -0,0 +1,83 @@
...
+-spec increment(atom(), atom()) ->
+                       ok.
+increment(App, Stat) ->
+    increment(App, Stat, 1).
+
+%% @doc decrement the counter for App named Stat
+%% by Amount.
+-spec decrement(atom(), atom(), integer()) ->
+                       ok.
+decrement(App, Stat, Amount) ->
```
jonmeredith, line 47: Could just do a negative amount, that would save making/decoding a tuple and matching on it.

src/riak_core_sup.erl
russelldb: I get it, but why remove call to register stats mod? If the stats sup crashes and is restarted it won't be able to restart the core stats unless the mod is registered.

russelldb: realised this is a stupid question, since it is the riak_core_sup that will restart the riak_core_metric_sup, args and all :blush:

PR #159, opened 2012-03-29
refactored gen_server2 -> riak_core_gen_server

camshaft:
I am developing an application using riak_core and rabbitmq messaging.  When I generated a release with rebar it produced this message:

```
init terminating in do_boot (Module gen_server2 potentially included by two different applications: rabbit_common and riak_core.)
```

This issue was discussed awhile back [here](http://comments.gmane.org/gmane.comp.db.riak.user/4100).

I changed the name to riak_core_gen_server and ran all tests (all 79 tests passed.)

This issue also came up with the priority_queue (originally from rabbitmq) which I also renamed to riak_core_priority_queue.

IMO this will make it easier to build applications using both of these technologies.  Let me know what you think.

Thanks for your time

### DIFF
--- a/ebin/riak_core.app
+++ b/ebin/riak_core.app
@@ -9,7 +9,7 @@
              bloom,
              chash,
              gen_nb_server,
-             gen_server2,
+             riak_core_gen_server,
              json_pp,
              merkerl,
              priority_queue,
--- a/ebin/riak_core.app
+++ b/ebin/riak_core.app
@@ -12,7 +12,7 @@
              riak_core_gen_server,
              json_pp,
              merkerl,
-             priority_queue,
+             riak_core_priority_queue,
              process_proxy,
              riak_core_gossip_legacy,
              riak_core,
--- a/rebar.config
+++ b/rebar.config
@@ -1,4 +1,4 @@
-{erl_first_files, ["src/gen_nb_server.erl", "src/gen_server2.erl"]}.
+{erl_first_files, ["src/gen_nb_server.erl", "src/riak_core_gen_server.erl"]}.
 {cover_enabled, true}.
 {erl_opts, [{parse_transform, lager_transform}]}.
 {edoc_opts, [{preprocess, true}]}.
--- a/src/riak_core_gen_server.erl
+++ b/src/riak_core_gen_server.erl
@@ -334,7 +334,7 @@ enter_loop(Mod, Options, State, ServerName, Timeout) ->
     Name = get_proc_name(ServerName),
     Parent = get_parent(),
     Debug = debug_options(Name, Options),
-    Queue = priority_queue:new(),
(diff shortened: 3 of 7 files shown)


rzezeski:
At this point I don't think there is any longer a reason to use `gen_server2`.  Would you at all be interested in updating this PR to just replace with `gen_server` and remove those two modules?

camshaft:
Yeah for sure.  What would you like to do about the priority_queue?

rzezeski:
Since that module is only used by `gen_server2` then remove it too.

rzezeski:
@CamShaft I can't review this ATM because master is currently locked-down until our next release settles.  Once that's over with I'll review/test this.  Thanks.

camshaft:
That's fine. Let me know if there's anything else we need before it gets pulled.

horkhe:
Hello Gentlemen, could you please advise in what release you plan to get that fixed? Thank you.

rzezeski:
@horkhe I hope to merge this in after the release of Riak 1.2.  Whatever release comes after that should include the change.

horkhe:
@rzezeski thank you

rzezeski:
So 1.2 is out the door now but I'm not sure we can remove gen_server2.  Basho has found a use for msg priorities.  Supposedly there is a gen_server2 application/dependency out there that could be used.  That would allow you to remove the dependency for you custom build with rabbitmq.

I have two concerns.
1. Are there any differences between the gen_server2 in Riak Core and some other copy of gen_server2?
2. Have bug fixes been made to OTP gen_server that have not made it to gen_server2?

I'd personally like to see gen_server2 removed but some analysis would have to be done to see if it affects Riak in any negative ways.

@CamShaft For now, I propose using one of the gen_server2 repos out there.  Would this solution work for you?

horkhe:
@rzezeski why don't just rename it to <code>riak_core_gen_server</code> as @CamShaft suggested and be done with it?

camshaft:
I think if Basho plans on making any changes, we go with what I originally had.  If there are no planned changes to gen_server2, let's pull in from an external source.

Thoughts?

rzezeski:
@CamShaft Okay, after some discussion with other bashoites I think we should go back to the rename change.  This makes the most sense because our gen_server2 copy is old and swapping the latest may cause issues so it would need a more thorough review.  But that is a separate concern and doesn't need to block this PR.

You should be able to revert the last commit, rebase this branch onto master, and then do a force push.  Sorry about the back-and-forth.

camshaft:
@rzezeski no worries.  I'll do that later today when I get a minute.

rzezeski:
@CamShaft thanks for your patience.

camshaft:
@rzezeski anything I can do to help.  You guys are doing awesome work.  Keep it up

Issue #154, opened 2012-03-21
handoff of fallback data can be delayed or never performed


rzezeski:
## Summary

During startup, a node with fallback data on it's secondary partitions
will not start those secondary partitions.  This means that nodes with
fallback data may delay or even never handoff the fallback data.
Depending on the ring and preflist this means that replicas may exist
on the same node for a longer period of time than required.
### Delayed Handoff

In the case where the home node of the fallback partitions is still
down then read/write load will cause the secondaries to spin back up
and handoff will commence when the home node comes alive.
### Missed Handoff

In the case where the home node of the fallbacks partitions is up when
the node(s) with the fallback partitions come up then the secondaries
will _not_ be started by read/write load since all the primaries are
up.  This means that, depending on the ring and preflist, 2 replicas
could exist on the same node until read repair or a write occurs.
That is, since the secondary partitions with fallback data are never
started handoff is missed and read repair + new writes must account
for _all_ of the fallback data that was written while the home node
was down.
## Steps to Reproduce
1. create devrel - `make devrel`, `for d in dev/dev*; do $d/bin/riak start; done`
2. join nodes - `for d in dev/dev*; do $d/bin/riak-admin join dev1@127.0.0.1; done`
3. wait for everything to settle (members + transfers)
4. stop dev4 - `./dev/dev4/bin/riak stop`
5. insert data - `for i in $(seq 1 100); do curl -X PUT -H 'content-type: text/plain' http://localhost:8091/riak/test/$i -d "hello"; done`
6. check data - `for i in $(seq 1 100); do curl http://localhost:8091/riak/test/$i 2> /dev/null && echo; done | wc -l`
7. stop dev1-dev3 `for d in dev/dev{1,2,3}; do $d/bin/riak stop; done`
8. start dev4 - `./dev/dev4/bin/riak start`
9. verify keys don't exist - `for i in $(seq 1 100); do curl http://localhost:8094/riak/test/$i 2> /dev/null && echo; done | less`
10. start dev1-dev3 - `for d in dev/dev*; do $d/bin/riak start; done`
11. wait a little, check transfers - `./dev/dev1/bin/riak-admin transfers`

After step 11 there will be some transfers listed but after a while
they will complete.  Those transfers are for stuff other than the
fallback data.  To be sure you can wait until no transfers are
reported, stop dev1 - dev3, and then try to read the keys from
dev4--they won't be there.
1. stop dev4 - `./dev/dev4/bin/riak stop`
[...]

Issue #150, opened 2012-03-14
riak_core_vnode_manager blocks on vnode startup


jonmeredith:
riak_core_vnode_manager is unresponsive while vnodes are being started by the vnode_sup (as it blocks in start_child).  This should be done asynchronously as it causes timeouts in other parts of the system - e.g. stats.

jrwest:
this has been addresssed: https://github.com/basho/riak_core/blob/master/src/riak_core_vnode_manager.erl#L570-L605

PR #163, opened 2012-04-05
Have vnode manager periodically start never-been-started vnodes

jtuple:
Change the vnode manager to keep track of which vnodes have been started at least once, and periodically  start nodes that have never been started. This ensures that vnodes responsible for fallback data are eventually started, and therefore perform hinted handoff.

Resolves issue #154

A riak_test for issue #154 has also been pushed into the riak_test repo. Against master, the test should fail. Against this branch, it should pass.

The passing test should look as follows:

```
$ ./riak_test rtdev.config gh_riak_core_154
01:50:50.398 [info] Application lager started on node 'riak_test@127.0.0.1'
01:50:50.401 [info] Riak path: "/tmp/rt"
01:50:50.401 [info] Running: /tmp/rt/dev/dev1/bin/riak stop
01:50:50.401 [info] Running: /tmp/rt/dev/dev2/bin/riak stop
01:50:52.676 [info] Resetting nodes to fresh state
01:50:52.676 [debug] Running: git --git-dir="/tmp/rt/dev/.git" --work-tree="/tmp/rt/dev" reset HEAD --hard
01:50:52.725 [debug] Running: git --git-dir="/tmp/rt/dev/.git" --work-tree="/tmp/rt/dev" clean -fd
01:50:52.766 [info] Running: /tmp/rt/dev/dev1/bin/riak start
01:50:52.766 [info] Running: /tmp/rt/dev/dev2/bin/riak start
01:50:54.741 [debug] Supervisor inet_gethost_native_sup started undefined at pid <0.64.0>
01:50:54.743 [debug] Supervisor kernel_safe_sup started inet_gethost_native:start_link() at pid <0.63.0>
01:50:54.764 [info] Deployed nodes: ['dev1@127.0.0.1','dev2@127.0.0.1']
01:50:54.783 [debug] [join] 'dev2@127.0.0.1' to ('dev1@127.0.0.1'): ok
01:50:55.802 [info] Cluster built: ['dev1@127.0.0.1','dev2@127.0.0.1']
01:50:55.802 [info] Increase handoff concurrency on nodes
01:50:55.802 [info] Running: /tmp/rt/dev/dev1/bin/riak stop
01:50:58.072 [info] Running: /tmp/rt/dev/dev1/bin/riak start
01:51:00.090 [info] Running: /tmp/rt/dev/dev2/bin/riak stop
01:51:02.395 [info] Running: /tmp/rt/dev/dev2/bin/riak start
01:51:04.182 [info] Write data while 'dev2@127.0.0.1' is offline
01:51:04.182 [info] Running: /tmp/rt/dev/dev2/bin/riak stop
01:51:07.080 [info] Verify that 'dev2@127.0.0.1' is missing data
01:51:07.080 [info] Running: /tmp/rt/dev/dev2/bin/riak start
01:51:08.867 [info] Running: /tmp/rt/dev/dev1/bin/riak stop
01:51:11.614 [info] Restart 'dev1@127.0.0.1' and sleep for 5 min, allowing handoff to occur
01:51:11.614 [info] Running: /tmp/rt/dev/dev1/bin/riak start
01:56:13.404 [info] Verify that 'dev2@127.0.0.1' has all data
01:56:13.404 [info] Running: /tmp/rt/dev/dev1/bin/riak stop
01:56:16.497 [info] gh_riak_core_154: passed
```

### DIFF
--- a/src/riak_core_vnode_manager.erl
+++ b/src/riak_core_vnode_manager.erl
@@ -41,7 +41,9 @@
 -record(idxrec, {key, idx, mod, pid, monref}).
 -record(state, {idxtab, 
                 forwarding :: [pid()],
-                handoff :: [{term(), integer(), pid(), node()}]
+                handoff :: [{term(), integer(), pid(), node()}],
+                known_modules :: [term()],
+                never_started :: [{integer(), term()}]
                }).
 
 -define(DEFAULT_OWNERSHIP_TRIGGER, 8).
@@ -107,7 +109,8 @@ get_tab() ->
 init(_State) ->
     {ok, Ring} = riak_core_ring_manager:get_raw_ring(),
     Mods = [Mod || {_, Mod} <- riak_core:vnode_modules()],
-    State = #state{forwarding=[], handoff=[]},
+    State = #state{forwarding=[], handoff=[],
+                   known_modules=[], never_started=[]},
     State2 = find_vnodes(State),
     AllVNodes = get_all_vnodes(Mods, State2),
     State3 = update_forwarding(AllVNodes, Mods, Ring, State2),
@@ -215,7 +218,9 @@ handle_cast(management_tick, State) ->
     AllVNodes = get_all_vnodes(Mods, State),
     State2 = update_handoff(AllVNodes, Ring, State),
     trigger_ownership_handoff(Mods, Ring, State2),
-    {noreply, State2};
+    State3 = update_never_started(Ring, State2),
+    State4 = maybe_start_vnodes(State3),
+    {noreply, State4};
 
 handle_cast(_, State) ->
     {noreply, State}.
@@ -251,7 +256,10 @@ code_change(_OldVsn, State, _Extra) ->
 %% ===================================================================
 
 schedule_management_timer() ->
-    timer:apply_after(30000, gen_server, cast, [?MODULE, management_tick]).
+    ManagementTick = app_helper:get_env(riak_core,
+                                        vnode_management_timer,
+                                        30000),
+    timer:apply_after(ManagementTick, gen_server, cast, [?MODULE, management_tick]).
 
 trigger_ownership_handoff(Mods, Ring, State) ->
     Transfers = riak_core_ring:pending_changes(Ring),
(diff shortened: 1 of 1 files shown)


src/riak_core_vnode_manager.erl
```diff
@@ -474,4 +482,43 @@ get_all_vnodes_status(State=#state{forwarding=Forwarding, handoff=HO}) ->
...
+    Length = length(NeverStarted),
+    Split =
+        case Length < MaxStart of
+            true ->
+                Length;
+            false ->
+                MaxStart
+        end,
+    {Start, NeverStarted2} = lists:split(Split, NeverStarted),
+    [get_vnode(Idx, Mod, State) || {Idx, Mod} <- Start],
```
jonmeredith: Could this block the vnodemgr for an extended period of time for a large vnode?

jonmeredith: Especially as it could be up to 16 at a time.

src/riak_core_vnode_manager.erl
```diff
@@ -474,4 +482,43 @@ get_all_vnodes_status(State=#state{forwarding=Forwarding, handoff=HO}) ->
...
+                                    ordsets:from_list(AlreadyStarted)),
+    NeverStarted2 = [{Idx, Mod} || Idx <- NeverStarted],
+    NeverStarted3 = NeverStarted2 ++ State#state.never_started,
+    KnownModules = [Mod | State#state.known_modules],
+    State#state{known_modules=KnownModules, never_started=NeverStarted3}.
+
+maybe_start_vnodes(State=#state{never_started=NeverStarted}) ->
+    MaxStart = app_helper:get_env(riak_core, vnode_rolling_start, 16),
+    Length = length(NeverStarted),
+    Split =
```
jonmeredith: I'm a sucker for erlang:min(Length, MaxStart)

src/riak_core_vnode_manager.erl
jtuple: Yes. Starting vnodes has always been a synchronous action for vnode manager (formerly master). We can reduce the default value to 1 if we'd like. No idea how long people expect to wait for hinted handoff to kick in after a reboot. Alternatively, I can start multiple vnodes in parallel -- but unclear about disk contention issues. Or, could just sends messages in this management tick; which trigger the actual vnode starts. In all cases, the actual starting of an individual vnode would remain synchronous though, just not the entire block.

We already have issue #150 filed to look into starting vnode asynchronously. I currently worry that we have some assumptions built-in about synchronous vnode starting with respect to handoff, vnode proxies, etc. So, would be a change we would need to reason about and test heavily.

jonmeredith: Agreed #150 needs a bit more thought on how things should work.

What about computing the list of never started secondaries, start one and if the tail is non-empty send an event to recompute. The vnode manager will block while it starts the vnode and any queued messages will have the opportunity to be processed before startign the next fallback.

Not sure if you'd have to factor out the start secondaries from the management_tick code or if it would be safe to run the mgmt code more frequently than every 30s.

rzezeski:
DISCLAIMER: I did not review any of the code for this PR.  Just have general comment.

At startup should we just check for any non-empty secondary-partitions?  That way if there is fallback data it will "immediately" be handed off rather than waiting on periodic start?

src/riak_core_vnode_manager.erl
jtuple: To be clear, are you suggesting we don't even have a start rate (eg. 16), but just continue to re-send the message each iteration if the list is non-empty? I'm somewhat partial to the temporal aspect of a tick-based approach, as it throttles starts overtime. Which is useful if we eventually move all vnodes (even primaries) to a rolling start model. How about going with your suggestion, but having a something like `vnode_start_tokens` that is set to 16 or whatever on management ticks, decremented each forced vnode start, and stops sending the continuation message when there are zero tokens. Until next tick, that is.

jtuple:
The question there is how do you determine non-empty?

For Riak K/V, the logic is in the backend module. But, the backend module is initialized by the vnode itself. Using the backend code directly to check `is_empty` would be an option, except that a backend may need to completely start-up before it can answer that question -- which has the same disk-contention issues we were already trying to avoid. Bitcask, for example, computes `is_empty` by doing a fold. It needs to be fully initialized to do so. Bitcask does it this way because deleted tombstones/expired keys effect the empty decision.

We could hack things using `du` I suppose.

Of course, this is a Riak Core issue. How do you generalize this beyond k/v? Vnode modules export `is_empty` as part of the behavior, but that logic once again typically depends on the vnode already been started and initialized in order to answer.

In a future major release, there will be changes to the vnode manager such that the vnode manager tracks the presence of secondary vnodes based on incoming requests and maintains persistent record of such. Once that's done, we'll have a better idea of who we should start first. Until then, it's not as straightforward.

src/riak_core_vnode_manager.erl
jonmeredith: A token/reset model works for me - at the risk of some extra complexity.

jtuple:
Just committed the reduced blocking version as discussed.

rzezeski:
All great points, and I agree it's not straightforward.  I wonder if in the future we might want to store some sort of persistent vnode state that is always cheap to check at startup so we can cheaply determine things like this.  Anyways, looks like you are on it. :thumbsup:

jonmeredith:
Tests well for me.
Killing vnode manager worked, it restored state successfully.
Starting up search after the KV node was up worked.
gh_riak_core_154 test works for me.

+1 merge.

PR #168, opened 2012-04-13
Raise handoff concurrency to two

jonmeredith:
Now that handoff concurrency limits apply to inbound/outbound, we should set the limit to 2 (conservatively) so that a node can at least handle an inbound and outbound transfer.  1 is too conservative, for a three node cluster only one transfer will ever be active at once.

Users can still set higher using the appenv.


jaredmorrow:
+1

PR #170, opened 2012-04-23
handoff visibility

rzezeski:
## Purpose

Visibility into handoff is really poor.  The typical method used to
discover handoff information is `riak-admin transfers` but that gives
hardly any useful information, as shown below.

```
'dev3@127.0.0.1' waiting to handoff 7 partitions
'dev2@127.0.0.1' waiting to handoff 4 partitions
'dev1@127.0.0.1' waiting to handoff 5 partitions
```

This PR adds visibility transfers/handoff by tracking various stats on
active transfers and displaying this information in a human friendly
way, as shown below.

```
./dev/dev1/bin/riak-admin transfers
'dev2@127.0.0.1' waiting to handoff 1 partitions
'dev1@127.0.0.1' waiting to handoff 2 partitions

Active Transfers:

transfer type: hinted_handoff
vnode type: riak_kv_vnode
partition: 1096126227998177188652763624537212264741949407232
started: 2012-04-23 03:08:27 [2.02 s ago]
last update: 2012-04-23 03:08:29 [16.97 ms ago]
objects transferred: 5720

                       2857 Objs/s                        
  dev2@127.0.0.1 =======================>   dev3@127.0.0.1
                          23.59 MB/s
```

This PR also gets rid of the annoying side effect of resetting the
inactivity timeout when calling `riak-admin transfers`.  This would
often cause users to wonder why handoffs were never occurring.
### vnode folds & status updates

One issue with handoff is that it uses vnode folds to do all it's
work.  This has the one nice benefit that it avoids a local copy of
data (1) but has bad side effect of using uninterruptable fold.  That
is, the vnode fold does the work as fast as it can and doesn't stop
until it's done (2).

In order to get status updates about the handoff the accumulator keeps
some local stats and _approximately_ every 2 seconds sends those stats
to the handoff manager via the `status_update/2` API.  I say the
timing is approximate because expiration of the interval is only
checked during a sender/receiver sync phase (determined by
`ACK_COUNT`).  If the receiver can't keep up or the sender fold is
slow then the status updates could take longer.  Essentially, this
code assumes that `ACK_COUNT` objects can be transferred in less than
2s.  **N.B.** The duration of the status update interval will not
invalidate the stats since they are based on start time and time of
last sync (see `riak_core_handoff_manager:update_stats/2`).

The reason the sender only sends a status update every 2s and only
checks if this interval has expired on sender/receiver sync is because
the vnode fold is a tight loop.  Sending an update for every object
would be too chatty and checking the interval every object could
potentially slow from overhead of getting time and doing math.
### determining transfer type

There are two types of transfer currently, _ownership handoff_ and
_hinted handoff_.  Soon there will be another type, _repair_.  In
order to disambiguate the two types of handoff I have to determine if
the source vnode is primary or not.  In the case of ownership handoff
it is a `primary -> secondary` handoff (where the secondary becomes
primary after handoff completes) and for hinted handoff it's
`secondary -> primary`.
### human friendly formatting

In order to make the stats a little easier to read I added a little
human friendly formatting.  I decided to put the code to support this
in Core rather than KV.  I stole and modified the code from
@seancribbs.
### pull vs. push

One aspect of this PR I'm not wild about is the fact that in order to
get the status a msg must be sent to each handoff manager on each each
node for every time `riak-admin transfers` is called (3).  I'd rather
see a push system where all active status data is collated at a
particular node, like the claimant node in ownership, and the status
call simply reads that.  The pull system is probably fine for now but
could cause trouble on larger clusters, especially if some script
accidentally calls it in a tight loop.
### dict vs. riak_core stats

I'm wondering if I should have make use of the stats API for the
collection of data in the handoff manager rather than a dict?
## Footnotes

1: That is, if the handoff sender process itself was running the
handoff then the vnode data would have to be copied from vnode heap to
sender heap.

2: In the future I think an iterator/cursor based approach to handoff
that is async, interruptable, and rate limited would be good.

3: Which calls `riak_core_status:all_active_transfers` where the RPC
is done.

### DIFF
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -36,6 +36,8 @@
 -export([add_outbound/4,
          add_inbound/1,
          status/0,
+         status/1,
+         status_update/2,
          set_concurrency/1,
          kill_handoffs/0
         ]).
@@ -46,17 +48,20 @@
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
--type mod()   :: atom().
 -type index() :: integer().
+-type modindex() :: {module(), index()}.
 
 -record(handoff_status,
-        { modindex      :: {mod(),index()},
-          node          :: atom(),
+        { modindex      :: modindex(),
+          src_node      :: node(),
+          target_node   :: node(),
           direction     :: inbound | outbound,
           transport_pid :: pid(),
           timestamp     :: tuple(),
           status        :: any(),
-          vnode_pid     :: pid() | undefined
+          stats         :: dict(),
+          vnode_pid     :: pid() | undefined,
+          type          :: ownership | hinted_handoff
         }).
 
 -record(state,
@@ -67,13 +72,16 @@
 %% this can be overridden with riak_core handoff_concurrency
 -define(HANDOFF_CONCURRENCY,2).
 
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
 start_link() ->
     gen_server:start_link({local, ?MODULE}, ?MODULE, [], []).
 
 init([]) ->
     {ok, #state{excl=ordsets:new(), handoffs=[]}}.
-
-%% handoff_manager API
 
 add_outbound(Module,Idx,Node,VnodePid) ->
     gen_server:call(?MODULE,{add_outbound,Module,Idx,Node,VnodePid}).
@@ -82,7 +90,16 @@ add_inbound(SSLOpts) ->
     gen_server:call(?MODULE,{add_inbound,SSLOpts}).
 
 status() ->
-    gen_server:call(?MODULE,status).
+    status(none).
+
+status(Filter) ->
+    gen_server:call(?MODULE, {status, Filter}).
+
+%% @doc Send status updates `Stats' to the handoff manager for a
+%%      particular handoff identified by `ModIdx'.
+-spec status_update(modindex(), proplists:proplist()) -> ok.
+status_update(ModIdx, Stats) ->
+    gen_server:cast(?MODULE, {status_update, ModIdx, Stats}).
 
 set_concurrency(Limit) ->
     gen_server:call(?MODULE,{set_concurrency,Limit}).
@@ -99,7 +116,10 @@ remove_exclusion(Module, Index) ->
 get_exclusions(Module) ->
     gen_server:call(?MODULE, {get_exclusions, Module}, infinity).
 
-%% gen_server API
+
+%%%===================================================================
+%%% Callbacks
+%%%===================================================================
 
 handle_call({get_exclusions, Module}, _From, State=#state{excl=Excl}) ->
     Reply =  [I || {M, I} <- ordsets:to_list(Excl), M =:= Module],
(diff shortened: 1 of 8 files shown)


src/riak_core_handoff_sender.erl
```diff
@@ -252,3 +278,53 @@ get_handoff_receive_timeout() ->
...
     EndFoldTime = now(),
     timer:now_diff(EndFoldTime, StartFoldTime) / 1000000.
+
+%% @private
+%%
+%% @doc Produce the value of `now/0' as if it were called `S' seconds
+%% in the future.
+-spec future_now(pos_integer()) -> erlang:timestamp().
+future_now(S) ->
+    {Megas, Secs, Micros} = now(),
```
Vagabond: os:timestamp() is better, unless this needs to be an always-increasing timestamp for some reason.

rzezeski: Better in what way? Less overhead/work?  Since this isn't in a tight loop I'm not sure how much it matters.

Vagabond: Better as in it doesn't tie up a global VM lock for no good reason.

src/riak_core_handoff_manager.erl
```diff
@@ -201,9 +232,57 @@ terminate(_Reason, _State) ->
...
+            ObjsS = round(Objs / ElapsedS),
+            BytesS = round(Bytes / ElapsedS),
+            [{objs_total, Objs},
+             {objs_per_s, ObjsS},
+             {bytes_per_s, BytesS},
+             {last_update, LastUpdate}]
+    end.
+
+filter(none) ->
+    true;
```
Vagabond: Won't this end up being called as lists:filter(true, ...) and throw a function clause error? Shouldn't you return a fun that always returns true instead?

rzezeski: You're right, that needs to return a fun.

src/riak_core_handoff_sender.erl
rzezeski: Well that is certainly a good reason.  I'll update.

src/riak_core_status.erl
```diff
@@ -64,6 +67,16 @@ transfers() ->
...
 
+%% @doc Produce status for all active transfers in the cluster.
+-spec all_active_transfers() -> {Xfers::list(), Down::list()}.
+all_active_transfers() ->
+    {Xfers, Down} =
+        riak_core_util:rpc_every_member(riak_core_handoff_manager,
+                                        status,
+                                        [{direction, outbound}],
+                                        5000),
+    {Xfers, Down}.
```
Vagabond, line 78: Why not just have the RPC call be the last expression in the function?

src/riak_core_handoff_sender.erl
```diff
@@ -175,39 +181,59 @@ start_fold(TargetNode, Module, Partition, ParentPid, SslOpts) ->
...
-             lager:error("Handoff of partition ~p ~p from ~p to ~p failed ~p:~p",
+             lager:error("Handoff of partition ~p ~p from ~p to ~p failed ~p:~p ~p",
                          [Module, Partition, node(), TargetNode,
-                          Err, Reason]),
+                          Err, Reason, erlang:get_stacktrace()]),
              gen_fsm:send_event(ParentPid, {handoff_error, Err, Reason})
      end.
 
 %% When a tcp error occurs, the ErrStatus argument is set to {error, Reason}.
 %% Since we can't abort the fold, this clause is just a no-op.
```
Vagabond, line 191: I really think making this record would be worth it.

rzezeski, line 191: Since I have to make more changes to the sender to support the upcoming repair command I'll do it then.

PR #173, opened 2012-05-09
Add claim sim

jonmeredith:
Add the claim sim that Joe Blomstedt wrote to core so we can ship it with Riak.

Build with the patch and  riak_core_claim_sim:help() should tell you all you need to know.

### DIFF
--- a/src/riak_core_claim_sim.erl
+++ b/src/riak_core_claim_sim.erl
@@ -1,0 +1,546 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_core: Core Riak Application
+%%
+%% Copyright (c) 2007-2012 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc riak_core_claim_sim models adding/removing nodes from the ring
+%%
+-module(riak_core_claim_sim).
+-compile(export_all).
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-record(simopts,  {wants,
+                   choose,
+                   current = fun(_Ring) -> ok end,
+                   prepared = fun(_Ring, _Prepare) -> ok end,
+                   percmd = fun(_Ring, _Cmd) -> ok end,
+                   rebalance = fun(_R1, _R2) -> ok end,
+                   analysis = fun(_Ring, _Cmds) -> ok end,
+                   return_ring = false}).
+
+help() ->
+    io:format("\n"
+              "Claim Simulator - simulates effect of join/leave commands\n"
+              "\n"
+              "  riak_core_claim_sim:run([Opts]).\n"
+              "\n"
+              "Opts = {claimant, node()} | \n"
+              "       {ringfile, path()} | {appvars, [{Key,Val}]} \n"
+              "       {prepare, finish | cancel}\n"
+              "       {print, false | standard_io | io_device()}\n"
+              "       {cmds, [Commands]}\n"
+              "\n"
+              "Commands = {join, node()} | {join, integer()}} | {leave, node()} | [Commands]\n"
+              "\n"
+              "The simulator uses the options to select the ring to use\n"
+              "  if claimant is set, grab the ring and target n-val from that node\n"
+              "  if ringfile is set, load the ring file from disk\n"
+              "  if neither set, use the current ring from the node\n"
+              "\n"
+              "Any active transfers are handled with the prepare option.  The default is to \n"
+              "finish the transfers, but they can be cancelled with {prepare, cancel}\n"
+              "\n"
+              "The cmds list provides the cluster transitions - commands can either be individual\n"
+              "join/leave commands or a batch to be added all at once.\n"
+              "{cmds, [[{join, 'riak@hosta'},{join, 'riak@hostb'}],{join,'riak@hostc'}]}\n"
+              "would join hosta/hostb and recalculate claim, then join hostc.\n"
+              "\n"
+              "Examples:\n"
+              "  Simulate adding 9 nodes in a single batch to the current ring\n"
+              "     riak_core_claim_sim:run([{cmds, [[{join, 9}]]}]).\n"
+              "\n"
+              "  Simulate cancelling transfers, adding a node then and removing a different node\n"
+              "     riak_core_claim_sim:run([{cmds, [{join, 'riak@newnode'},\n"
+              "                                      {leave, 'riak@oldnode'}]},\n"
+              "                              {prepare, cancel}]).\n"
+              "\n").
+
+run([Config]) when is_list(Config) ->
+    io:format("Loading config: ~s~n", [Config]),
+    case file:consult(Config) of
+        {ok, Terms} ->
+            run(Terms);
+        {error, Reason} ->
+            {cannot_read_config, Reason}
+        end;
+run(Opts) ->
+    Claimant = proplists:get_value(claimant, Opts),
+    RingFile = proplists:get_value(ringfile, Opts),
+    AppVars = proplists:get_value(riak_core, Opts, []),
+    Prepare = proplists:get_value(prepare, Opts, finish),
+    Cmds = proplists:get_value(cmds, Opts, []),
+    IoDev = proplists:get_value(print, Opts, standard_io),
+    Analysis = proplists:get_value(analysis, Opts, []),
+    
+    case {Claimant, RingFile} of
+        {undefined, undefined} ->
+            {ok, Ring} = riak_core_ring_manager:get_raw_ring(),
+            SimOpts = default_simopts(IoDev, Analysis),
+            dryrun(Ring, Prepare, Cmds, SimOpts);
+        {Claimant, undefined} ->
+            load_vars(Claimant),
+            run_claimant(IoDev, Claimant, Prepare, Analysis, Cmds);
+        {undefined, RingFile} ->
+            setup_environment(AppVars),
+            run_ring(IoDev, RingFile, Prepare, Analysis, Cmds);
+        _ ->
+            help()
+    end.           
+
+run_rebalance(Ring, Wants, Choose, Rebalance) ->
+    %% o(IoDev, "~nAfter rebalance~n", []),        
+    Ring2 = riak_core_claim:claim(Ring, Wants, Choose),
+    Rebalance(Ring, Ring2),
+    Ring2.
+
+read_ringfile(RingFile) ->
+    case file:read_file(RingFile) of
+        {ok, Binary} ->
+            binary_to_term(Binary);
+        {error, Reason} ->
+            throw({bad_ring, Reason})
+    end.
+
+setup_environment(Vars) ->
+    default_vars(),
+    [application:set_env(riak_core, Key, Val) || {Key, Val} <- Vars],
+    ok.
+
+load_var(Node, App, Var, _Default) ->
+    Result = rpc:call(Node, application, get_env, [App, Var]),
+    case Result of
+        {ok, Value} ->
+            application:set_env(App, Var, Value);
+        _ ->
+            ok
+    end.
+
+load_vars(Node) ->
+    application:load(riak_core),
+    load_var(Node, riak_core, default_bucket_props, [{n_val, 3}]),
+    %% load_var(Node, riak_core, wants_claim_fun, {riak_core_claim, wants_claim_v2}),
+    application:set_env(riak_core, wants_claim_fun, {riak_core_claim, wants_claim_v2}),
+    load_var(Node, riak_core, choose_claim_fun, {riak_core_claim, choose_claim_v2}),
+    load_var(Node, riak_core, target_n_val, 4).
+
+default_vars() ->
+    application:load(riak_core),
+    application:set_env(riak_core, default_bucket_props, [{n_val, 3}]),
+    application:set_env(riak_core, wants_claim_fun, {riak_core_claim, wants_claim_v2}),
+    application:set_env(riak_core, choose_claim_fun, {riak_core_claim, choose_claim_v2}),
+    application:set_env(riak_core, target_n_val, 4).
+
+run_ring(IoDev, RingFile, Prepare, Analysis, CmdsL) ->
+    Ring = read_ringfile(RingFile),
+    SimOpts = default_simopts(IoDev, Analysis),
+    dryrun(Ring, Prepare, CmdsL, SimOpts).
+
+run_claimant(IoDev, Claimant, Prepare, Analysis, CmdsL) ->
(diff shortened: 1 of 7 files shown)


src/riak_core_claim_sim.erl
```diff
@@ -0,0 +1,575 @@
...
+                 {CMod, CFun} ->
+                     [];
+                 {CMod, CFun, Params0} ->
+                     Params0
+             end,
+    Params1 = [{target_n_val, TN} | proplists:delete(target_n_val, Params)],
+    {CMod, CFun, Params1}.
+
+run_rebalance(Ring, Wants, Choose, Rebalance) ->
+    %% o(IoDev, "~nAfter rebalance~n", []),
```
jtuple: Move this into `make_rebalance` and uncomment it. Unlike original sim, we no longer print "After rebalance" after different sets -- just dump rins without any context. Also, we don't print new lines between different ring outputs. Makes it hard to read. Would be a good idea to add newlines back.

src/riak_core_claim_sim.erl
```diff
@@ -0,0 +1,575 @@
...
+                            RingAcc3
+                    end, RingAcc1, Cmds),
+              {_, NewRing2} = riak_core_ring:reassign_indices(NewRing),
+              Pending = riak_core_ring:pending_changes(NewRing2),
+              case Pending of
+                  [] ->
+                      NewRing3 = NewRing2;
+                  _ ->
+                      NewRing3 = riak_core_ring:finish_transfers(NewRing2)
+                      %result(IoDev, NewRing2, NewRing3)
```
jtuple: Original claim sim would print results after completing the transfers but before the next rebalance cycle. This version does not. Are we losing any interesting information?

src/riak_core_claim_sim.erl
```diff
@@ -0,0 +1,575 @@
...
+                          {choose,{riak_core_claim,choose_claim_v1}},
+                          {cmds, [[{join,3}]]},
+                          {analysis, [{failures, 2},{n_val, 3}]},
+                          {print,Fh},
+                          {return_ring, false}])),
+    file:close(Fh).
+                         
+                     
+
+commission_test_() ->
```
jtuple: Useful function to have for analysis, but I don't think we should run it every time we invoke our eunit suite.

src/riak_core_ring.erl
```diff
@@ -747,6 +750,44 @@ pretty_print(Ring, Opts) ->
...
+    {RingChanged or NextChanged, CState3}.
+
+%% @doc Return a ring with all transfers completed - for claim sim
+finish_transfers(Ring) ->
+    Pending = riak_core_ring:pending_changes(Ring),
+    Owners1 = riak_core_ring:all_owners(Ring),
+    Owners2 = [{Idx,NOwner} || {Idx,_,NOwner,_,_} <- Pending],
+    Owners3 = lists:ukeysort(1, Owners2++Owners1),
+    CH1 = Ring?CHSTATE.chring,
+    CH2 = {element(1,CH1), Owners3},
```
jtuple, line 799: I know this was copied verbatim from my original sim code, but should we take the time to add a function to `chash.erl` to alter a chash so that we don't need to assume knowledge of an opaque tuple here?

jtuple, line 799: Then again, we could probably replace this `finish_transfers` function with `future_ring` provided by the staged clustering code after merging and integrating both branches. That does something similar:
https://github.com/basho/riak_core/blob/gh177-staged-clustering/src/riak_core_ring.erl#L758

jtuple:
In general, I'm worried about the simulation code becoming out-of-sync with potential changes in the claimant logic in the future. This is fine for now given the code freeze deadline. But, in the near future we should change this simulator to use `riak_core_claimant:compute_all_next_rings` provided by the new staged clustering code.

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+%% Return a list of failures records, one for each case examined
+node_sensitivity(R, NVal, Depth) ->
+    Members = riak_core_ring:all_members(R),
+    DownCombos = down_combos(Depth, Members),
+    LoadCombos = [{Down, node_load(R, NVal, Down)} || Down <- DownCombos],
+    [analyze_load(Down, Load) || {Down, Load} <- LoadCombos].
+
+%% For a given ring, with a list of downed nodes, compute
+%% all preference lists then count up the number that
+%% each node partipates in.
```
jtuple: `partipates` -> `participates`

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+           norm_fb = norm_fb(NumFBs, TotFBs)} || 
+        {N, NumPris, NumFBs} <- VL].
+
+vnode_load(R, NVal, DownNodes) ->
+    UpNodes = riak_core_ring:all_members(R) -- DownNodes,
+    Keys = [<<(I+1):160/integer>> ||
+               {I,_Owner} <- riak_core_ring:all_owners(R)],
+    %% NValParts = Nval * riak_core_ring:num_partitions(R),
+    AllPLs = [riak_core_apl:get_apl_ann(Key, NVal, R, UpNodes) || Key <- Keys],
+    FlatPLs = lists:flatten(AllPLs),
```
jtuple, line 136: Could also use `lists:flatmap` instead of comprehension + `lists:flatten`

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+              [FBMin, toint(FBMean), toint(FBStdDev), toint(FB10), toint(FB90), FBMax, Down, Worst]),
+    print_analysis1(IoDev, Rest).
+
+%% @private round to nearest int
+toint(F) when is_number(F) ->
+    round(F+0.5);
+toint(X) ->
+    X.
+
+%% Order failurest by number of nodes down ascending, then fbmax, then down list
```
jtuple: `failurest`? Is this a typo for `failure set`?

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+%% -------------------------------------------------------------------
+
+%% @doc Adjacency measures are useful for preflist diversity. Three different
+%% forms of measurement are used.
+%%
+%% An adjacency list is a list of node pairs and the distance between them
+%% by pointwise measuring the distance to each of the other nodes
+%% taking into account wraparounds (forward distance only).
+%%
+%% The ownership list a,b,c,b,a,c,b,c would compute the following adjaceny
```
jtuple: `adjaceny` -> `adjacency`

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+    end.
+
+%% For each pair, get the count of distances < NVal
+score_am([], _NVal) ->
+    undefined;
+score_am(AM, NVal) ->
+    Cs = lists:flatten(
+           [begin
+                [C || {D,C} <- count(Ds, NVal), D < NVal]
+            end || {_Pair,Ds} <- AM]),
```
jtuple, line 374: Could also use `lists:flatmap` instead of comprehension + `lists:flatten`

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+                             riak_core_ring:add_member(hd(Nodes), R, N)
+                     end, R0, Nodes),
+    lists:foldl(fun({I,N}, R) ->
+                        riak_core_ring:transfer_node(I, N, R)
+                end, R1, NewOwners).
+
+
+
+%% Generate a completion test function that makes sure all required
+%% dbs are created
```
jtuple: What are `dbs`? Also, this is the first time you use `RequiredDs` and `DAM`. It's not immediately obvious what these mean. After looking over later code, I'm guessing "required distances" and "dictionary adjacency matrix" but a comment would be helpful here.

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+%% Generate until a fixed length has been hit
+gen_complete_len(Len) ->
+    fun(Owners, _AM) ->
+            {length(Owners) == Len, Len}
+    end.
+
+%% M = list of node names
+%% NVal = target nval
+%% Required = list of required indices - should replace
+
```
jtuple: Is this comment associated with the next function: `construct`. If so, there is nothing named `Required` anywhere.

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+construct(Complete, M, NVal) ->
+    DAM1 = lists:foldl(fun(Pair,AM0) ->
+                               ?DICT:append_list(Pair, [], AM0)
+                       end, ?DICT:new(), [{F,T} || F <- M, T <- M, F /= T]),
```
jtuple: Had to take a minute to parse the list comprehension to determine that `DAM1` was an "empty dictionary adjacency matrix; mapping all possible pairs to the empty list". Might make sense to add a comment here, and possibly even move this into a separate `empty_adjacency_matrix(Members)` function.

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+construct(Complete, M, Owners, DAM, NVal) ->
+    %% Work out which pairs do not have the requiredDs
+    case Complete(Owners, DAM) of
+        {true, _DesiredLen}->
+            {ok, Owners, DAM};
+        {false, DesiredLen} ->
+            %% Easy ones - restrict the eligible list to not include the N-1 
+            %% previous nodes.  If within NVal-1 of possibly closing the ring
+            %% then restrict in that direction as well.
+            Eligible0 = M -- lists:sublist(Owners, NVal - 1),
```
jtuple, line 443: It is not clear to me why we also restrict eligibility in the reverse direction when we have less than `NVal` remaining.

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+                    lager:debug("construct -- unable to construct without violating NVal"),
+                    {Owners1, DAM1} = prepend_next_owner(M, M, Owners, DAM, NVal),
+                    construct(Complete, M, Owners1, DAM1, NVal);
+                _ ->
+                    {Owners1, DAM1} = prepend_next_owner(M, Eligible, Owners, DAM, NVal),
+                    construct(Complete, M, Owners1, DAM1, NVal)
+            end
+    end.
+
+met_required(Owners, DAM, RequiredDs) ->
```
jtuple, line 464: This returns true only when we have met all required distances across all possible pairs in the adjacency matrix? A comment would be very helpful here, even if it just reinforces what a reader determines by looking over the code.

src/riak_core_claim_util.erl
```diff
@@ -0,0 +1,651 @@
...
+%% Generate all permutations of list L
+perm_gen([E]) ->
+    [[E]];
+perm_gen(L) ->
+    lists:append([ begin
+                       [ [X | Y] || Y <- perm_gen(lists:delete(X, L))]
+                   end || X <- L]).
+  
+
+%% Pick all combinations of Depth nodes from the Mmebers list
```
jtuple: `Mmebers` -> `Members`

jtuple:
The `riak_core_claim_util` additions are rather dense. It took me awhile to parse a lot of the code, and I've had several past conversations to hear your intentions to give me a head-start. Better comments and documentation are going to be essential for anyone returning to this code in 6 months and trying to fix a bug or tweak things.

I've given some suggestions for better comments, but that may just be a good starting point. Not sure how much time we have to fiddle with documentation right now. My only concern is that the `riak_core_claim_util` additions are themselves more lines of code then both our `claim_v1` and `claim_v2` algorithms; and that's not even counting the actual `claim_v3` code.

This is to be expected, as adding more constraints will lead to more complex code. We just need to figure out how to best manage this complexity so claim doesn't become something no one really understands.

On another note, it's unclear why some of this code is in `riak_core_claim_util` despite being a core component of `claim_v3` and of no use to anything else. Adjacency matrix stuff is useful for analysis and makes sense, but things like `construct, prepend_next_owner, build` and whatnot seem like shadows of `claim_v3` itself. No big deal for now, but something to consider refactoring down the line.

In any case, I'm +1 for merging after addressing the specific comments I've made and after some attempt at a bit more documentation. Additional documentation, refactoring, and code improvement can all be future work after this release cycle.

This code will conflict heavily with the claimant logic being removed from `riak_core_ring` in my staged clustering work. But, I think that merging this in, then rebasing my staged clustering pull-request, will be the cleanest approach. We shall see.

src/riak_core_claim_sim.erl
jonmeredith: Some of those print removals were an attempt to reduce the amount of output for sims.  As it's most likely to be us running it I'll add it back in.

src/riak_core_ring.erl
jonmeredith, line 799: I think we should fix up post-merge.

src/riak_core_claim_util.erl
jonmeredith, line 136: Agreed - as we're so close to the wire I'd rather wait until we have the opportunity to refactor the sim to use more of the staged claim.  Would be nice if there was a flattening list comprehension.

src/riak_core_claim_util.erl
jonmeredith: typo for failures - I  wanted the list ordered by all single failures first, then multiple failures.

src/riak_core_claim_sim.erl
jonmeredith: I didn't think we were at the time I removed it, but I was probably looking at something that consumed the final ring. It should be there.

src/riak_core_claim_sim.erl
jonmeredith: I was torn - it does make sure the functions run, but without checking the results it gives false security.

src/riak_core_claim_util.erl
jonmeredith: Muscle memory problem - it was supposed to be d's - but changed to distances.  During earl development work for the diversity work I was generating rings and making sure there was one of each distance.  this code could be culled.  I've kept the early work on https://github.com/jonmeredith/ringgen/blob/master/ringgen.erl

src/riak_core_claim_util.erl
jonmeredith: Should have been removed when it was refactored.

src/riak_core_claim_util.erl
jonmeredith: agreed - too much voodoo.

src/riak_core_claim_util.erl
jonmeredith, line 443: It's trying to make sure that the rings it generates do not have wraparound violations.  While tinkering with the algorithm I was trying to answer questions like how large does a ring need to be to satisfy diversity constraints, so was allowing the ring to grow until it was met, but as rings had to be a power of 2 I needed a mechanism to generate good rings when closing the loop.

A better option would be to speculatively close the ring  and if that fails, continue on without the constraint, but the function had served it's purpose by then.

jonmeredith, line 443: In ClaimV3 this code path is only really used if the incoming ring is really out of whack and we're going to 'rediagaonalize' we have a go at generating a diverse ring and if that fails fall back to the (improved for the wraparound case) traditional diagonalize

src/riak_core_claim_util.erl
jonmeredith, line 464: Commented.

src/riak_core_claim_util.erl
jonmeredith: Fixed.

src/riak_core_claim_sim.erl
jonmeredith: I added it back in, but the code path doesn't get hit. I think it's because all of the claim sim deals with directly with calling riak_core_claim:claim/3 rather than calling riak_core_ring:rebalance_ring() which calls riak_core_claim/1 and grabs the wants/claims function from the environment.

jonmeredith:
rebased against master and forced

PR #174, opened 2012-05-14
Slf dtrace take1

slfritchie:
Add basic infrastructure Erlang-triggerable probes to Riak Core apps.

In an ideal world, this module would live in a repo that would be
easily sharable across multiple Basho projects.  The tricky bit for
this would be trying to make generic the
`application:get_env(riak_core, dtrace_support)' call that's
currently in the`riak_kv_dtrace:dtrace/1' function.  But we'll
wait for another day, I think.

The purpose of this module is to reduce the overhead of DTrace (and
SystemTap) probes when those probes are: 1. not supported by the VM,
or 2. disabled by application configuration.  #1 is the bigger
problem: a single call to the code loader can take several
milliseconds.  #2 is useful in the case that we want to try to reduce
the overhead of adding these probes even further by avoiding the NIF
call entirely.

SLF's MacBook Pro tests

without cover, with R14B04 + DTrace:

```
timeit_naive                 average  2236.306 usec/call over     500.0 calls
timeit_mochiglobal           average     0.509 usec/call over  225000.0 calls
timeit_best OFF (fastest)    average     0.051 usec/call over  225000.0 calls
timeit_best ON -init         average     1.027 usec/call over  225000.0 calls
timeit_best ON +init         average     0.202 usec/call over  225000.0 calls
```

with cover, with R14B04 + DTrace:

```
timeit_naive                 average  2286.202 usec/call over     500.0 calls
timeit_mochiglobal           average     1.255 usec/call over  225000.0 calls
timeit_best OFF (fastest)    average     1.162 usec/call over  225000.0 calls
timeit_best ON -init         average     2.207 usec/call over  225000.0 calls
timeit_best ON +init         average     1.303 usec/call over  225000.0 calls
```

### DIFF
--- a/src/riak_core_dtrace.erl
+++ b/src/riak_core_dtrace.erl
@@ -58,12 +58,16 @@ dtrace(ArgList) ->
                         {5.8, _} ->
                             %% R14B04
                             put(?MAGIC, dtrace),
-                            dtrace(ArgList);
+                            if ArgList == enabled_check -> ok;
+                               true                     -> dtrace(ArgList)
+                            end;
                         {Num, _} when Num > 5.8 ->
                             %% R15B or higher, though dyntrace option
                             %% was first available in R15B01.
                             put(?MAGIC, dyntrace),
-                            dtrace(ArgList);
+                            if ArgList == enabled_check -> ok;
+                               true                     -> dtrace(ArgList)
+                            end;
                         _ ->
                             put(?MAGIC, unsupported),
                             false
@@ -73,9 +77,13 @@ dtrace(ArgList) ->
                     false
             end;
         dyntrace ->
-            erlang:apply(dyntrace, p, ArgList);
+            if ArgList == enabled_check -> ok;
+               true                     -> erlang:apply(dyntrace, p, ArgList)
+            end;
         dtrace ->
-            erlang:apply(dtrace, p, ArgList);
+            if ArgList == enabled_check -> ok;
+               true                     -> erlang:apply(dtrace, p, ArgList)
+            end;
         _ ->
             false
     end.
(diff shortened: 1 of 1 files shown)


beerriot:
The best time for me on OS X 10.6.8, R14B04+Dtrace without cover was "timeit_best ON +init":

```
timeit_naive                 average  6441.864 usec/call over     500.0 calls
timeit_mochiglobal           average     2.110 usec/call over  225000.0 calls
timeit_best OFF (fastest)    average     1.244 usec/call over  225000.0 calls
timeit_best ON -init         average     1.185 usec/call over  225000.0 calls
timeit_best ON +init         average     0.229 usec/call over  225000.0 calls
```

Same with cover (though I also don't see the difference your results showed when enabling cover, so maybe I have these confused):

```
timeit_naive                 average  6412.970 usec/call over     500.0 calls
timeit_mochiglobal           average     2.048 usec/call over  225000.0 calls
timeit_best OFF (fastest)    average     1.225 usec/call over  225000.0 calls
timeit_best ON -init         average     1.212 usec/call over  225000.0 calls
timeit_best ON +init         average     0.231 usec/call over  225000.0 calls
```

In case this helps:

```
Erlang R14B04 (erts-5.8.5) [source] [64-bit] [smp:4:4] [rq:4] [async-threads:64] [hipe] [kernel-poll:true] [dtrace]
```

beerriot:
Ooops. Nope, ignore that.  The `timeit…on` tests explicitly set `{dtrace_support, true}` but `timeit…off` doesn't explicitly set `{dtrace_support, false}`.  I had set `true` in the app.config, so the test was invalid.  My numbers now mirror yours.

slfritchie:
The ordering of the runtime based on the env settings, etc. isn't ideal.  Sorry!

src/riak_core_dtrace.erl
```diff
@@ -0,0 +1,286 @@
...
+        unsupported ->
+            false;
+        _ ->
+            S0 = if is_atom(String0) -> erlang:atom_to_binary(String0, latin1);
+                    true             -> String0
+                 end,
+            dtrace([Int0, Int1] ++ Ints ++ [S0, String1] ++ Strings)
+    end.
+
+enabled() ->
```
beerriot, line 124: `enabled/0` looks _very_ similar to `dtrace/1`. Is there any chance the redundancies could be merged for maintainability?

beerriot:
Other than maybe DRYing up `enabled/0`, I am +1 on this PR. Great work, Scott & Ryan!

slfritchie:
Bryan, did you want to re-examine commit e70bc2c?  It looks like my fear of extra overhead isn't justified, looking at the `rebar eunit skip_deps=true suite=riak_core_dtrace` timing report before & after that commit.

beerriot:
I see no appreciable change in the timings before an after e70bc2c … _but_ `enabled/0` is returning `ok` instead of `true`, which means that `put_tag/1' doesn't work.`s/ok;/true;/' and it's all set.

PR #175, opened 2012-05-17
Repair Command

rzezeski:
Add repair ability

Add the framework needed for services on top of core to provide a
"repair" mechanism.  At this point repair was built with the sole
purpose of repairing data for KV and Search.

The key insight behind repair is that since Riak is a replicated data
store one partition can be rebuilt from the replicas on other
partitions.  Specifically, the adjacent (i.e. before and after)
partitions on the ring, together, contain all the replicas that are
stored on the partition to be repaired.  The rub is that adjacent
partitions also contain replicas that are _not_ meant to be on the
partition under repair.  This means a filter function must be used
while folding over the source partitions to transfer only the data
that belongs.  This is done as efficiently as possible by generating a
hash range for all the buckets and thus avoiding a preflist
calculation for each key.  Only a hash of each key is done, it's range
determined from a bucket->range map, and then the hash is checked
against the range.  The services under repair (i.e. Search or KV) must
provide callback functions to generate these ranges since it is
specific to the service.

In order to not repeat code and capitalize on concurrency control the
repair mechanism is currently an extension of handoff.  This creates
some awkwardness as repair is _not_ handoff.  Some of the differences
include:
1. It needs to filter data during the fold
2. All vnodes involved are primary and thus repair _cannot_ block
3. Repair involves 3 distinct partitions and 3 distinct vnodes
4. Repair does not imply a change of responsibility from one vnode to
   another, but is more a sharing of data

In the future the handoff subsystem should be rewritten around the
notion of "transfers" in which repair, handoff, and ownership are all
different logical operations but use the transfer mechanism
underneath.  Especially now that repair is in it should be more clear
what that system should look like to meet the goals of all 3.

The idea of repair lives in the vnode manager as it is very much a
vnode semantic.  This is important because other things like handoff
and ownership also affect vnodes and thus affect repair.  The vnode
manager controls the logical repair, but the handoff/transfer
mechanism controls the physical movement of replicas from source to
target.

For now, it seemed easiest and smartest to _not_ allow ownership
change and repair to run concurrently.  In the event that an ownership
change is detected all repairs will be hard killed, regardless of
status.

In the case were the low-level repair transfer is killed because of
`handoff_concurrency` a msg is _not_ sent back to the vnode mgr to
indicate a failure for that transfer.  Instead, the vnode mgr has a
period tick.  During each tick the vnode mgr checks the status of all
it's repairs and retries any transfers that have since died for a
reason other than completion.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
-
+APPS = kernel stdlib sasl erts ssl tools os_mon runtime_tools crypto inets \
+	public_key mnesia syntax_tools compiler
+COMBO_PLT = $(HOME)/.riak_core_combo_dialyzer_plt
 
 .PHONY: deps test
 
@@ -13,7 +15,7 @@ deps:
 clean:
 	./rebar clean
 
-distclean: clean 
+distclean: clean
 	./rebar delete-deps
 
 test: all
@@ -22,7 +24,19 @@ test: all
 docs: deps
 	./rebar skip_deps=true doc
 
+build_plt: compile
+	dialyzer --build_plt --output_plt $(COMBO_PLT) --apps $(APPS) \
+		deps/*/ebin apps/*/ebin
+
+check_plt: compile
+	dialyzer --check_plt --plt $(COMBO_PLT) --apps $(APPS) \
+		deps/*/ebin apps/*/ebin
+
 dialyzer: compile
-	@dialyzer -Wno_return -c apps/riak_core/ebin
+	@echo
+	@echo Use "'make check_plt'" to check PLT prior to using this target.
+	@echo Use "'make build_plt'" to build PLT prior to using this target.
+	@echo
+	dialyzer --plt $(COMBO_PLT) ebin
 
 
--- a/Makefile
+++ b/Makefile
@@ -26,11 +26,11 @@ docs: deps
 
 build_plt: compile
 	dialyzer --build_plt --output_plt $(COMBO_PLT) --apps $(APPS) \
-		deps/*/ebin apps/*/ebin
+		deps/*/ebin
 
 check_plt: compile
 	dialyzer --check_plt --plt $(COMBO_PLT) --apps $(APPS) \
-		deps/*/ebin apps/*/ebin
+		deps/*/ebin
 
 dialyzer: compile
 	@echo
--- a/include/riak_core_handoff.hrl
+++ b/include/riak_core_handoff.hrl
@@ -13,3 +13,26 @@
         }).
 
 -type ho_stats() :: #ho_stats{}.
+-type ho_type() :: ownership_handoff | hinted_handoff.
+-type predicate() :: fun((any()) -> boolean()).
+
+-type index() :: integer().
+-type mod_src_tgt() :: {module(), index(), index()}.
+-type mod_partition() :: {module(), index()}.
+
+-record(handoff_status,
+        { mod_src_tgt           :: mod_src_tgt(),
+          src_node              :: node(),
+          target_node           :: node(),
+          direction             :: inbound | outbound,
+          transport_pid         :: pid(),
+          timestamp             :: tuple(),
+          status                :: any(),
+          stats                 :: dict(),
+          vnode_pid             :: pid() | undefined,
+          type                  :: ownership | hinted_handoff | repair,
+          req_origin            :: node(),
+          filter_mod_fun        :: {module(), atom()}
+        }).
+-type handoff() :: #handoff_status{}.
+-type handoffs() :: [handoff()].
--- a/include/riak_core_handoff.hrl
+++ b/include/riak_core_handoff.hrl
@@ -26,10 +26,12 @@
           target_node           :: node(),
           direction             :: inbound | outbound,
           transport_pid         :: pid(),
+          transport_mon         :: reference(),
           timestamp             :: tuple(),
           status                :: any(),
           stats                 :: dict(),
           vnode_pid             :: pid() | undefined,
+          vnode_mon             :: reference(),
           type                  :: ownership | hinted_handoff | repair,
           req_origin            :: node(),
           filter_mod_fun        :: {module(), atom()}
--- a/include/riak_core_handoff.hrl
+++ b/include/riak_core_handoff.hrl
@@ -13,7 +13,7 @@
         }).
 
 -type ho_stats() :: #ho_stats{}.
--type ho_type() :: ownership_handoff | hinted_handoff.
+-type ho_type() :: ownership_handoff | hinted_handoff | repair.
 -type predicate() :: fun((any()) -> boolean()).
 
 -type index() :: integer().
@@ -32,9 +32,8 @@
           stats                 :: dict(),
           vnode_pid             :: pid() | undefined,
           vnode_mon             :: reference(),
-          type                  :: ownership | hinted_handoff | repair,
+          type                  :: ho_type(),
           req_origin            :: node(),
           filter_mod_fun        :: {module(), atom()}
         }).
--type handoff() :: #handoff_status{}.
--type handoffs() :: [handoff()].
+-type handoff_status() :: #handoff_status{}.
--- a/src/riak_core_bucket.erl
+++ b/src/riak_core_bucket.erl
@@ -29,7 +29,10 @@
          set_bucket/2,
          get_bucket/1,
          get_bucket/2,
-         merge_props/2]).
+         get_buckets/1,
+         merge_props/2,
+         name/1,
+         n_val/1]).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
@@ -107,6 +110,14 @@ get_bucket(Name, Ring) ->
         {ok, Bucket} -> Bucket
     end.
 
+%% @doc Get bucket properties `Props' for all the buckets in the given
+%%      `Ring'
+-spec get_buckets(riak_core_ring:riak_core_ring()) ->
+                         Props::list().
+get_buckets(Ring) ->
+    Names = riak_core_ring:get_buckets(Ring),
+    [get_bucket(Name, Ring) || Name <- Names].
+
 %% @private
 -spec validate_props(BucketProps::list({PropName::atom(), Value::any()}),
                      Validators::list(module()),
@@ -120,6 +131,12 @@ validate_props(_, [], Errors) ->
 validate_props(BucketProps0, [{_App, Validator}|T], Errors0) ->
     {BucketProps, Errors} = Validator:validate(BucketProps0),
     validate_props(BucketProps, T, lists:flatten([Errors|Errors0])).
+
+name(BProps) ->
+    proplists:get_value(name, BProps).
+
+n_val(BProps) ->
+    proplists:get_value(n_val, BProps).
 
 %% ===================================================================
 %% EUnit tests
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -98,7 +98,7 @@ status(Filter) ->
 
 %% @doc Send status updates `Stats' to the handoff manager for a
 %%      particular handoff identified by `ModIdx'.
--spec status_update(modindex(), proplists:proplist()) -> ok.
+-spec status_update(modindex(), ho_stats()) -> ok.
 status_update(ModIdx, Stats) ->
     gen_server:cast(?MODULE, {status_update, ModIdx, Stats}).
 
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -48,7 +48,6 @@
         ]).
 
 -include("riak_core_handoff.hrl").
--define(DEFAULT_TIMEOUT, 60000).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
@@ -154,24 +153,24 @@ get_exclusions(Module) ->
 
 handle_call({get_exclusions, Module}, _From, State=#state{excl=Excl}) ->
     Reply =  [I || {M, I} <- ordsets:to_list(Excl), M =:= Module],
-    {reply, {ok, Reply}, State, timeout()};
+    {reply, {ok, Reply}, State};
 handle_call({add_outbound,Mod,Idx,Node,Pid},_From,State=#state{handoffs=HS}) ->
     case send_handoff(Mod,Idx,Node,Pid,HS) of
         {ok,Handoff=#handoff_status{transport_pid=Sender}} ->
             HS2 = HS ++ [Handoff],
-            {reply, {ok,Sender}, State#state{handoffs=HS2}, timeout()};
+            {reply, {ok,Sender}, State#state{handoffs=HS2}};
         {false,_ExistingHandoff=#handoff_status{transport_pid=Sender}} ->
-            {reply, {ok,Sender}, State, timeout()};
+            {reply, {ok,Sender}, State};
         Error ->
-            {reply, Error, State, timeout()}
+            {reply, Error, State}
     end;
 handle_call({add_inbound,SSLOpts},_From,State=#state{handoffs=HS}) ->
     case receive_handoff(SSLOpts) of
         {ok,Handoff=#handoff_status{transport_pid=Receiver}} ->
             HS2 = HS ++ [Handoff],
-            {reply, {ok,Receiver}, State#state{handoffs=HS2}, timeout()};
+            {reply, {ok,Receiver}, State#state{handoffs=HS2}};
         Error ->
-            {reply, Error, State, timeout()}
+            {reply, Error, State}
     end;
 
 handle_call({xfer_status, Xfer}, _From, State=#state{handoffs=HS}) ->
@@ -186,7 +185,7 @@ handle_call({kill_xfer, Xfer, Reason}, _From, State) ->
     HS = State#state.handoffs,
     case lists:keytake(TP, #handoff_status.transport_pid, HS) of
         false ->
-            {reply, ok, State, timeout()};
+            {reply, ok, State};
         {value, Xfer, HS2} ->
             #handoff_status{mod_src_tgt={Mod, SrcPartition, TargetPartition},
                             type=Type,
@@ -197,7 +196,7 @@ handle_call({kill_xfer, Xfer, Reason}, _From, State) ->
             lager:info(Msg, [Type, Mod, SrcNode, SrcPartition,
                              TargetNode, TargetPartition, Reason]),
             exit(TP, {kill_xfer, Reason}),
-            {reply, ok, State#state{handoffs=HS2}, timeout()}
+            {reply, ok, State#state{handoffs=HS2}}
     end;
 
 handle_call({send_handoff, Mod, {Src, Target}, ReqOrigin, {FilterMod, FilterFun}=FMF},
@@ -208,14 +207,14 @@ handle_call({send_handoff, Mod, {Src, Target}, ReqOrigin, {FilterMod, FilterFun}
     case send_handoff({Mod, Src, Target}, ReqOrigin, SrcPid, HS, {Filter, FMF}, ReqOrigin) of
         {ok, Handoff} ->
             HS2 = HS ++ [Handoff],
-            {reply, {ok, Handoff}, State#state{handoffs=HS2}, timeout()};
+            {reply, {ok, Handoff}, State#state{handoffs=HS2}};
         {false, Handoff} ->
-            {reply, {ok, Handoff}, State, timeout()}
+            {reply, {ok, Handoff}, State}
     end;
 
 handle_call({status, Filter}, _From, State=#state{handoffs=HS}) ->
     Status = lists:filter(filter(Filter), [build_status(HO) || HO <- HS]),
-    {reply, Status, State, timeout()};
+    {reply, Status, State};
 
 handle_call({set_concurrency,Limit},_From,State=#state{handoffs=HS}) ->
     application:set_env(riak_core,handoff_concurrency,Limit),
@@ -228,9 +227,9 @@ handle_call({set_concurrency,Limit},_From,State=#state{handoffs=HS}) ->
             {_Keep,Discard}=lists:split(Limit,HS),
             [erlang:exit(Pid,max_concurrency) ||
                 #handoff_status{transport_pid=Pid} <- Discard],
-            {reply, ok, State, timeout()};
+            {reply, ok, State};
         false ->
-            {reply, ok, State, timeout()}
+            {reply, ok, State}
     end;
 
 handle_call(get_concurrency, _From, State) ->
@@ -239,7 +238,7 @@ handle_call(get_concurrency, _From, State) ->
 
 handle_cast({del_exclusion, {Mod, Idx}}, State=#state{excl=Excl}) ->
     Excl2 = ordsets:del_element({Mod, Idx}, Excl),
-    {noreply, State#state{excl=Excl2}, timeout()};
+    {noreply, State#state{excl=Excl2}};
 
 handle_cast({add_exclusion, {Mod, Idx}}, State=#state{excl=Excl}) ->
     {ok, Ring} = riak_core_ring_manager:get_raw_ring(),
@@ -251,28 +250,24 @@ handle_cast({add_exclusion, {Mod, Idx}}, State=#state{excl=Excl}) ->
             ok
     end,
     Excl2 = ordsets:add_element({Mod, Idx}, Excl),
-    {noreply, State#state{excl=Excl2}, timeout()};
+    {noreply, State#state{excl=Excl2}};
 
 handle_cast({status_update, ModSrcTgt, StatsUpdate}, State=#state{handoffs=HS}) ->
     case lists:keyfind(ModSrcTgt, #handoff_status.mod_src_tgt, HS) of
         false ->
             lager:error("status_update for non-existing handoff ~p", [ModSrcTgt]),
-            {noreply, State, timeout()};
+            {noreply, State};
         HO ->
             Stats2 = update_stats(StatsUpdate, HO#handoff_status.stats),
             HO2 = HO#handoff_status{stats=Stats2},
             HS2 = lists:keyreplace(ModSrcTgt, #handoff_status.mod_src_tgt, HS, HO2),
-            {noreply, State#state{handoffs=HS2}, timeout()}
+            {noreply, State#state{handoffs=HS2}}
     end.
-
-handle_info(timeout, State) ->
-    %% do nothing with timeout now but maybe leave it in here?
-    {noreply, State, timeout()};
 
 handle_info({handoff_finished, {_Mod, _Target}, Handoff, _Reason}, State) ->
     %% NOTE: this msg will only come in for repair handoff
     riak_core_vnode_manager:xfer_complete(Handoff),
-    {noreply, State, timeout()};
+    {noreply, State};
 
 handle_info({'DOWN',_Ref,process,Pid,Reason},State=#state{handoffs=HS}) ->
     case lists:keytake(Pid,#handoff_status.transport_pid,HS) of
@@ -310,9 +305,9 @@ handle_info({'DOWN',_Ref,process,Pid,Reason},State=#state{handoffs=HS}) ->
             end,
 
             %% removed the handoff from the list of active handoffs
-            {noreply, State#state{handoffs=NewHS}, timeout()};
+            {noreply, State#state{handoffs=NewHS}};
         false ->
-            {noreply, State, timeout()}
+            {noreply, State}
     end.
 
 
@@ -504,11 +499,6 @@ update_stats(StatsUpdate, Stats) ->
     Stats2 = dict:update_counter(objs, Objs, Stats),
     Stats3 = dict:update_counter(bytes, Bytes, Stats2),
     dict:store(last_update, LU, Stats3).
-
-%% @private
--spec timeout() -> Timeout::pos_integer().
-timeout() ->
-    app_helper:get_env(riak_core, handoff_manager_timeout, ?DEFAULT_TIMEOUT).
 
 %%%===================================================================
 %%% Tests
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -12,7 +12,7 @@
 %% specific language governing permissions and limitations
 %% under the License.
 
-%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2012 Basho Technologies, Inc.  All Rights Reserved.
 -module(riak_core_handoff_manager).
 -behaviour(gen_server).
 
@@ -35,6 +35,10 @@
 %% handoff api
 -export([add_outbound/4,
          add_inbound/1,
+         xfer/3,
+         retry_xfer/1,
+         xfer_status/1,
+         kill_xfer/2,
          status/0,
          status/1,
          status_update/2,
@@ -44,35 +48,22 @@
         ]).
 
 -include("riak_core_handoff.hrl").
+-define(DEFAULT_TIMEOUT, 60000).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
--type index() :: integer().
--type modindex() :: {module(), index()}.
-
--record(handoff_status,
-        { modindex      :: modindex(),
-          src_node      :: node(),
-          target_node   :: node(),
-          direction     :: inbound | outbound,
-          transport_pid :: pid(),
-          timestamp     :: tuple(),
-          status        :: any(),
-          stats         :: dict(),
-          vnode_pid     :: pid() | undefined,
-          type          :: ownership | hinted_handoff
-        }).
-
 -record(state,
         { excl,
-          handoffs :: [#handoff_status{}]
+          handoffs=[] :: handoffs()
         }).
 
 %% this can be overridden with riak_core handoff_concurrency
 -define(HANDOFF_CONCURRENCY,2).
-
+-define(HO_EQ(HOA, HOB),
+        HOA#handoff_status.mod_src_tgt == HOB#handoff_status.mod_src_tgt
+        andalso HOA#handoff_status.timestamp == HOB#handoff_status.timestamp).
 
 %%%===================================================================
 %%% API
@@ -90,6 +81,36 @@ add_outbound(Module,Idx,Node,VnodePid) ->
 add_inbound(SSLOpts) ->
     gen_server:call(?MODULE,{add_inbound,SSLOpts}).
 
+%% @doc Initiate a transfer from `SrcPartition' to `TargetPartition'
+%%      for the given `Module' using the `FilterModFun' filter.
+-spec xfer({index(), node()}, mod_partition(), {module(), atom()}) ->
+                  handoff().
+xfer({SrcPartition, SrcOwner}, {Module, TargetPartition}, FilterModFun) ->
+    %% NOTE: This will not work with old nodes
+    ReqOrigin = node(),
+    {ok, Xfer} = gen_server:call({?MODULE, SrcOwner},
+                                 {send_handoff, Module,
+                                  {SrcPartition, TargetPartition},
+                                  ReqOrigin, FilterModFun}),
+    Xfer.
+
+%% @doc Retry the given `Xfer'.
+-spec retry_xfer(handoff()) -> NewXfer::handoff().
+retry_xfer(Xfer) ->
+    #handoff_status{mod_src_tgt={Module, SrcPartition, TargetPartition},
+                    src_node=SrcOwner,
+                    filter_mod_fun=FilterModFun} = Xfer,
+    xfer({SrcPartition, SrcOwner}, {Module, TargetPartition}, FilterModFun).
+
+-spec xfer_status(handoff()) -> complete | in_progress | not_found.
+xfer_status(HS) ->
+    case HS#handoff_status.status of
+        complete -> complete;
+        _ ->
+            SrcNode = HS#handoff_status.src_node,
+            gen_server:call({?MODULE, SrcNode}, {xfer_status, HS})
+    end.
+
 status() ->
     status(none).
 
@@ -98,7 +119,7 @@ status(Filter) ->
 
 %% @doc Send status updates `Stats' to the handoff manager for a
 %%      particular handoff identified by `ModIdx'.
--spec status_update(modindex(), ho_stats()) -> ok.
+-spec status_update(mod_src_tgt(), ho_stats()) -> ok.
 status_update(ModIdx, Stats) ->
     gen_server:cast(?MODULE, {status_update, ModIdx, Stats}).
 
@@ -107,6 +128,12 @@ set_concurrency(Limit) ->
 
 get_concurrency() ->
     gen_server:call(?MODULE, get_concurrency).
+
+%% @doc Kill the transfer `Xfer' with `Reason'.
+-spec kill_xfer(handoff(), any()) -> ok.
+kill_xfer(Xfer, Reason) ->
+    SrcNode = Xfer#handoff_status.src_node,
+    ok = gen_server:call({?MODULE, SrcNode}, {kill_xfer, Xfer, Reason}).
 
 kill_handoffs() ->
     set_concurrency(0).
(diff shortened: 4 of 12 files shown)


src/riak_core_handoff_manager.erl
```diff
@@ -123,27 +150,68 @@ get_exclusions(Module) ->
...
+                             TargetNode, TargetPartition, Reason]),
+            exit(TP, {kill_xfer, Reason}),
+            {reply, ok, State#state{handoffs=HS2}, timeout()}
+    end;
+
+handle_call({send_handoff, Mod, {Src, Target}, ReqOrigin, {FilterMod, FilterFun}=FMF},
+            _From,
+            State=#state{handoffs=HS}) ->
+    Filter = FilterMod:FilterFun(Target),
+    {ok, SrcPid} = riak_core_vnode_manager:get_vnode_pid(Src, riak_search_vnode),
```
jtuple: This should be `Mod` not `riak_search_vnode`, no? Also, why doesn't this break things -- ie. your provided riak_test passes for me.

jtuple:
As discussed over chat, repair fails when one of the repair sibling vnodes is on the same node as the node under repair. It's gen_server call timeout error:

```
gen_server riak_core_handoff_manager terminated with reason:
{{timeout,
  {gen_server, call,
    [{riak_core_handoff_manager, 'dev1@127.0.0.1'},
     {send_handoff,
      riak_kv_vnode,
      {1438665674247607560106752257205091097473808596992, 0},
      'dev1@127.0.0.1',
      {riak_kv_vnode, repair_filter}}]}},
 {gen_server, call,
   [riak_core_vnode_manager,
    {1438665674247607560106752257205091097473808596992,
     riak_search_vnode, get_vnode}, infinity]}}
```

I'm pretty sure this is a deadlock:
`riak_core_vnode_manager/repair` calls `riak_core_handoff_manager/send_handoff` calls `riak_core_vnode_manager/get_vnode_ip`

EDIT by RZ - This has been addressed, although `handoff_concurrency` of 1 will obviously cause never-ending repair.

Makefile
```diff
@@ -22,7 +24,19 @@ test: all
 docs: deps
 	./rebar skip_deps=true doc
 
+build_plt: compile
```
jtuple, line 27: `riak_core` does not contain an `apps` directory, so `build_plt` and `check_plt` both fail unless I modify the Makefile.

```
dialyzer: No such file, directory or application: "apps/*/ebin"
```

include/riak_core_handoff.hrl
```diff
@@ -13,3 +13,26 @@
         }).
 
 -type ho_stats() :: #ho_stats{}.
+-type ho_type() :: ownership_handoff | hinted_handoff.
```
jtuple: This can also be of type `repair`

include/riak_core_handoff.hrl
```diff
@@ -13,3 +13,26 @@
...
+        { mod_src_tgt           :: mod_src_tgt(),
+          src_node              :: node(),
+          target_node           :: node(),
+          direction             :: inbound | outbound,
+          transport_pid         :: pid(),
+          timestamp             :: tuple(),
+          status                :: any(),
+          stats                 :: dict(),
+          vnode_pid             :: pid() | undefined,
+          type                  :: ownership | hinted_handoff | repair,
```
jtuple: `riak_core_handoff_manager` ends up setting `#handoff_status.type` using a variable that is of type `ho_type()`, so should change this to be `ho_type()`. This would also fix the issue that `ownership_handoff` is what is actually set, not `ownership`.

include/riak_core_handoff.hrl
```diff
@@ -13,3 +13,26 @@
...
+          req_origin            :: node(),
+          filter_mod_fun        :: {module(), atom()}
+        }).
+-type handoff() :: #handoff_status{}.
```
jtuple: Maybe it's a personal preference, but I'm not a huge fan of the type being named `handoff`, yet having all the code in handoff/vnode managers using the `#handoff_status{}` record to get/set elements. A bit confusing until you realise `handoff == #handoff_status`.

src/riak_core_bucket.erl
```diff
@@ -107,6 +110,14 @@ get_bucket(Name, Ring) ->
         {ok, Bucket} -> Bucket
     end.
 
+%% @doc Get bucket properties `Props' for all the buckets in the given
+%%      `Ring'
+-spec get_buckets(riak_core_ring:riak_core_ring()) ->
+                         Props::list().
+get_buckets(Ring) ->
```
jtuple, line 117: Not really relevant to repair, but would it make sense to have a variant of this function that also returned the default bucket props appended to the list as , say, `[{default, Props}]`? Could have used it in the `NValMap` code in repair to avoid needing to query the default `n_val` separately.

src/riak_core_handoff_manager.erl
```diff
@@ -89,6 +80,36 @@ add_outbound(Module,Idx,Node,VnodePid) ->
...
     gen_server:call(?MODULE,{add_inbound,SSLOpts}).
 
+%% @doc Initiate a transfer from `SrcPartition' to `TargetPartition'
+%%      for the given `Module' using the `FilterModFun' filter.
+-spec xfer({index(), node()}, mod_partition(), {module(), atom()}) ->
+                  handoff().
+xfer({SrcPartition, SrcOwner}, {Module, TargetPartition}, FilterModFun) ->
+    %% NOTE: This will not work with old nodes
+    ReqOrigin = node(),
+    {ok, Xfer} = gen_server:call({?MODULE, SrcOwner},
```
jtuple: Cross-node call here that defaults to 5 second timeout. Is 5 second timeout reasonable? When this call fails or times out, is throwing an exception and crashing acceptable? Or better to catch and return some error?

src/riak_core_handoff_manager.erl
```diff
@@ -89,6 +80,36 @@ add_outbound(Module,Idx,Node,VnodePid) ->
...
+                    filter_mod_fun=FilterModFun} = Xfer,
+    xfer({SrcPartition, SrcOwner}, {Module, TargetPartition}, FilterModFun).
+
+-spec xfer_status(handoff()) -> complete | in_progress | not_found.
+xfer_status(HS) ->
+    case HS#handoff_status.status of
+        complete -> complete;
+        _ ->
+            SrcNode = HS#handoff_status.src_node,
+            gen_server:call({?MODULE, SrcNode}, {xfer_status, HS})
```
jtuple: Same as above. Cross-node call here that defaults to 5 second timeout. Is 5 second timeout reasonable? When this call fails or times out, is throwing an exception and crashing acceptable? Or better to catch and return some error?

src/riak_core_handoff_manager.erl
```diff
@@ -97,13 +118,19 @@ status(Filter) ->
...
     gen_server:cast(?MODULE, {status_update, ModIdx, Stats}).
 
 set_concurrency(Limit) ->
     gen_server:call(?MODULE,{set_concurrency,Limit}).
 
+%% @doc Kill the transfer `Xfer' with `Reason'.
+-spec kill_xfer(handoff(), any()) -> ok.
+kill_xfer(Xfer, Reason) ->
+    SrcNode = Xfer#handoff_status.src_node,
+    ok = gen_server:call({?MODULE, SrcNode}, {kill_xfer, Xfer, Reason}).
```
jtuple: Another 5 second default cross-node call. Opinions on this one?

src/riak_core_handoff_manager.erl
```diff
@@ -97,13 +118,19 @@ status(Filter) ->
...
+-spec status_update(mod_src_tgt(), ho_stats()) -> ok.
 status_update(ModIdx, Stats) ->
     gen_server:cast(?MODULE, {status_update, ModIdx, Stats}).
 
 set_concurrency(Limit) ->
     gen_server:call(?MODULE,{set_concurrency,Limit}).
 
+%% @doc Kill the transfer `Xfer' with `Reason'.
+-spec kill_xfer(handoff(), any()) -> ok.
+kill_xfer(Xfer, Reason) ->
```
jtuple: It looks like `kill_xfer` is implicitly assumed to only ever be used on outgoing transfers. What happens if it was used on an inbound transfer? Related fact, inbound transfers don't set `#handoff_status.type`, so when `handle_call(kill_xfer...)` prints the type to the console, it would print undefined.

src/riak_core_handoff_manager.erl
```diff
@@ -172,24 +242,35 @@ handle_cast({add_exclusion, {Mod, Idx}}, State=#state{excl=Excl}) ->
...
-            {noreply, State#state{handoffs=HS2}}
+            HS2 = lists:keyreplace(ModSrcTgt, #handoff_status.mod_src_tgt, HS, HO2),
+            {noreply, State#state{handoffs=HS2}, timeout()}
     end.
 
+handle_info(timeout, State) ->
+    %% do nothing with timeout now but maybe leave it in here?
+    {noreply, State, timeout()};
+
+handle_info({handoff_finished, {_Mod, _Target}, Handoff, _Reason}, State) ->
```
jtuple: I'm pretty sure this message isn't ever sent anymore after some refactoring you've done prior to submitting this pull-request. If this is dead code, let's remove it.

jtuple: Also, `riak_core_vnode_manager:xfer_complete/1` doesn't exist, only `/2`

src/riak_core_handoff_manager.erl
```diff
@@ -248,7 +331,8 @@ build_status(HO) ->
                     stats=Stats,
                     type=Type}=HO,
     {status_v2, [{mod, Mod},
```
jtuple, line 345: Is there anything you know about that uses `status_v2` that would be affected by the renames here? Stands out only because it's a versioned record. On the other hand, I can't seem to find anything in Riak that really does much with this record. So, no idea.

src/riak_core_handoff_manager.erl
```diff
@@ -295,14 +379,29 @@ handoff_concurrency_limit_reached () ->
...
+    send_handoff({Mod, Partition, Partition}, Node, Pid, HS, {none, none}, none).
+
+%% @private
+%%
+%% @doc Start a handoff process for the given `Mod' from
+%%      `Src'/`VNode' to `Target'/`Node' using the given `Filter'
+%%      function which is a predicate applied to the key.  The
+%%      `Origin' is the node this request originated from so a reply
+%%      can't be sent on completion.
+-spec send_handoff({module(), index(), index()}, node(),
```
jtuple, line 404: Seems a little strange to group `{Mod, SrcIdx, TargetIdx}` without also including `TargetNode` in that group -- technically, all 4 ids label a given transfer. Then again, it looks like there's lots of logic that just uses `ModSrcTgt`. I guess that's safe for now,  ownership/repairs are always `primary -> primary` which has the implicit guarantee that `TargetIdx` also uniquely ids `TargetNode` (the primary owner). And secondaries are `secondary -> primary` which has the same benefit. Just something that caught my eye and wanted to note. Maybe we should just document this implicit assumption somewhere.

src/riak_core_handoff_manager.erl
```diff
@@ -364,11 +481,12 @@ receive_handoff (SSLOpts) ->
             {ok, #handoff_status{ transport_pid=Pid,
```
jtuple, line 496: Does it matter that `#handoff_status.type` is never set for inbound transfers and defaults to `undefined`?

src/riak_core_handoff_sender.erl
```diff
@@ -33,18 +31,46 @@
 %% note this is in seconds
 -define(STATUS_INTERVAL, 2).
 
-start_link(TargetNode, Module, Partition, Vnode) ->
+%% Accumulator for the visit item HOF
+-record(ho_acc,
```
jtuple, line 40: I owe you a beer for killing the tuple and moving towards this record. In any case, I've tried to look over this code very closely and ensure tuple to record transition was done properly. It looks good, although it's such a pain that I never know if I'm 100%. Hopefully you, me, and dialyzer are a good enough check here.

src/riak_core_handoff_manager.erl
```diff
@@ -97,13 +118,19 @@ status(Filter) ->
 
 %% @doc Send status updates `Stats' to the handoff manager for a
 %%      particular handoff identified by `ModIdx'.
```
jtuple: Update comment: `ModIdx` to `ModSrcTgt`

src/riak_core_handoff_sender.erl
```diff
@@ -314,13 +359,14 @@ incr_objs(Stats=#ho_stats{objs=Objs}) ->
 %% @doc Check if the interval has elapsed and if so send handoff stats
 %%      for `ModIdx' to the manager and return a new stats record
```
jtuple: Update comment: `ModIdx` to `ModSrcTgt`

src/riak_core_repair.erl
```diff
@@ -0,0 +1,121 @@
...
+%% ===================================================================
+
+%% @doc Generate the hash `Range' for a given `Target' partition and
+%%      `NVal'.
+%%
+%% Note: The type of NVal should be pos_integer() but dialyzer says
+%%       success typing is integer() and I don't have time for games.
+-spec gen_range(partition(), riak_core_ring:riak_core_ring(), integer()) ->
+                       hash_range().
+gen_range(Target, Ring, NVal) ->
```
jtuple, line 73: `Ring` is not a `riak_core_ring()` here, it is a `chash()`. Either change callers in `riak_kv_vnode` and `riak_search_vnode` to pass in an actual ring and extract the chash for the chash funs, or update the spec and rename `Ring` to `CH`, `CHash`, etc.

src/riak_core_repair.erl
```diff
@@ -0,0 +1,121 @@
...
+-spec gen_range_fun(list(), hash_range()) -> function().
+gen_range_fun(RangeMap, Default) ->
+    fun(Bucket) ->
+            case lists:keyfind(Bucket, 1, RangeMap) of
+                false -> Default;
+                {_, Val} -> Val
+            end
+    end.
+
+%% @doc Generate the map from bucket `B' to hash `Range' that a key
```
jtuple, line 111: This function was confusing because the comment and return spec mention buckets, but there are no buckets in the input spec or anywhere within the actual code. Maybe change input spec `list()` to `[{riak_object:bucket(), pos_integer()}]`, and rename `I` to `B` or `Bucket` in the list comprehension? Also, the word bucket is used a lot in this/related modules. Is that the right generic term? For search, these are actually schema definitions, not buckets, right? Does this matter? And is `riak_object:bucket()` still the right type?

src/riak_core_vnode_manager.erl
```diff
@@ -68,6 +88,49 @@ all_vnodes(Mod) ->
 all_vnodes_status() ->
     gen_server:call(?MODULE, all_vnodes_status).
 
+%% @doc Repair the given `ModPartition' pair for `Service' using the
```
jtuple, line 95: I really want to see this `repair` function changed so that it determines the `Owner` using `riak_core_ring:index_owner` and then immediately does a `gen_server:call` to the `Owner` which handles all the repair logic. In other words, move all this logic into the `handle_call` on the `Owner`. Currently, this logic is performed wherever, and then the `repair` call is made to the `Owner` which does the rest. Why not combine the two? It would also guarantee that things are calculated using the owner's ring rather than the ring on the node the user ran repair from.

src/riak_core_vnode_manager.erl
```diff
@@ -68,6 +88,49 @@ all_vnodes(Mod) ->
...
+%%      given `FilterModFun' to filter keys.
+-spec repair(atom(), {module(), partition()}, {module(), atom()}) ->
+                    {ok, Pairs::[{partition(), node()}]} |
+                    {down, Down::[{partition(), node()}]} |
+                    ownership_change_in_progress.
+repair(Service, {_Module, Partition}=ModPartition, FilterModFun) ->
+    %% Fwd the request to the partition owner to guarentee that there
```
jtuple: Typo: `guarentee` -> `guarantee`

src/riak_core_vnode_manager.erl
```diff
@@ -171,6 +235,64 @@ handle_call({Partition, Mod, get_vnode}, _From, State) ->
     {reply, {ok, Pid}, State};
 handle_call(get_tab, _From, State) ->
     {reply, ets:tab2list(State#state.idxtab), State};
+
+handle_call({repair, ModPartition, Pairs, FilterModFun}, _From, State) ->
+    #state{repairs=Repairs} = State,
```
jtuple: Consider using `Repairs = State#state.repairs` to match the syntax in the following handle_calls?

src/riak_core_vnode_manager.erl
```diff
@@ -171,6 +235,64 @@ handle_call({Partition, Mod, get_vnode}, _From, State) ->
...
+    end;
+
+handle_call({repair_status, ModPartition}, _From, State) ->
+    Repairs = State#state.repairs,
+    case get_repair(ModPartition, Repairs) of
+        none -> {reply, not_found, State};
+        #repair{} -> {reply, in_progress, State}
+    end;
+
+handle_call({xfer_complete, Xfer}, _From, State) ->
```
jtuple: There are some implicit assumptions here. It would be helpful to note here that `xfer_complete` will always be called on the origin "node-under-repair" and that the transfer target-partition (which you call `Partition`) is always going to be a local primary of this node (which, again, is the origin).

src/riak_core_vnode_manager.erl
```diff
@@ -171,6 +235,64 @@ handle_call({Partition, Mod, get_vnode}, _From, State) ->
...
+    Repairs = State#state.repairs,
+    {Module, _, Partition} = Xfer#handoff_status.mod_src_tgt,
+    ModPartition = {Module, Partition},
+    case get_repair(ModPartition, Repairs) of
+        none ->
+            lager:error("Received xfer_complete for non-existing repair: ~p",
```
jtuple, line 296: Here we print an error and then happily continue running, but ... `<continued below>`

src/riak_core_vnode_manager.erl
```diff
@@ -171,6 +235,64 @@ handle_call({Partition, Mod, get_vnode}, _From, State) ->
...
+            {reply, ok, State};
+        #repair{minus_one_xfer=MOX, plus_one_xfer=POX}=R ->
+            R2 = if ?XFER_EQ(MOX, Xfer) ->
+                         MOX2 = MOX#handoff_status{status=complete},
+                         R#repair{minus_one_xfer=MOX2};
+                    ?XFER_EQ(POX, Xfer) ->
+                         POX2 = POX#handoff_status{status=complete},
+                         R#repair{plus_one_xfer=POX2};
+                    true ->
+                         throw({xfer_complete_matches_none, R, Xfer})
```
jtuple: `<continued from above>` ...here we throw an exception on error. Why is one error conditional crash worthy and not the other? Or should they be the same? It seems to me that if this `vnode_manager` and the relevant `handoff_manager` ever become out of sync (ie. either one of them crashes and loses state), that either one of these two error conditions is equally likely.

src/riak_core_vnode_manager.erl
```diff
@@ -269,8 +410,7 @@ schedule_management_timer() ->
                                         10000),
     timer:apply_after(ManagementTick, gen_server, cast, [?MODULE, management_tick]).
 
-trigger_ownership_handoff(Mods, Ring, State) ->
-    Transfers = riak_core_ring:pending_changes(Ring),
+trigger_ownership_handoff(Transfers, Mods, State) ->
```
jtuple, line 438: The signature change to `tigger_ownership_handoff` breaks `handle_cast(ring_changed)` code. Please fix the `ring_changed` code.

src/riak_core_vnode_manager.erl
```diff
@@ -534,3 +674,101 @@ maybe_start_vnodes(State=#state{vnode_start_tokens=Tokens,
...
+    case riak_core_handoff_manager:xfer_status(Xfer) of
+        complete -> Xfer;
+        in_progress -> Xfer;
+        not_found -> riak_core_handoff_manager:retry_xfer(Xfer)
+    end.
+
+%% @private
+-spec is_pending_ownership_changes(riak_core_ring:riak_core_ring()) ->
+                                          boolean().
+is_pending_ownership_changes(Ring) ->
```
jtuple: Any objections to renaming this to `are_pending_ownership_changes`? Also, thoughts on moving into `riak_core_ring`?

src/riak_core_vnode_manager.erl
```diff
@@ -534,3 +674,101 @@ maybe_start_vnodes(State=#state{vnode_start_tokens=Tokens,
...
+
+%% @private
+%%
+%% @doc Get the three `{Partition, Owner}' pairs involved in a repair
+%%      operation for the given `Ring' and `Partition'.
+-spec repair_pairs(riak_core_ring:riak_core_ring(), non_neg_integer()) ->
+                          [{Partition::non_neg_integer(), Owner::node()}].
+repair_pairs(Ring, Partition) ->
+    Owner = riak_core_ring:index_owner(Ring, Partition),
+    CH = element(4, Ring),
```
jtuple: How about adding `riak_core_ring:chash(Ring)` that returns the chash and not directly reference the tuple here?

jtuple:
There are numerous changes to `riak_core_handoff_manager.erl` that introduce a `timeout()` parameter to all the `gen_server` callbacks, but then you completely ignore the timeout. I imagine this was used earlier before some refactor. In any case, I would argue for removing all this timeout code.

Also, I don't believe the `gen_server` style timeout is ever what we'd want for the handoff manager, rather than explicit `timer:apply_after` based timers or even a periodic tick. Specifically, `gen_server` timeouts have the nasty property of being reset everytime a message is received -- which makes sense for something like an "idle timer" like vnodes implement. If a user sat around asking for handoff status over and over again, you would never hit your timeout. Which would be problematic if it actually mattered.

**EDIT by RZ** - I removed this timeout.

jtuple:
I've noticed errors in the crash.log related to max concurrent that I didn't noticed before.

```

2012-06-01 14:42:51 =SUPERVISOR REPORT====
     Supervisor: {local,riak_core_handoff_sender_sup}
     Context:    child_terminated
     Reason:     {shutdown,max_concurrency}
     Offender:   [{pid,<0.3394.0>},{name,riak_core_handoff_sender},{mfargs,{riak_core_handoff_sender,start_link,undefined}},{restart_type,temporary},{shutdown,brutal_kill},{child_type,worker}]
```

It's unclear if this is related to a change in this pull-request, or something I missed before. Ie:
https://github.com/basho/riak_core/pull/175/files#L5R210

In any case, I have no problem with console.log and error.log stuff. But, generally am not a fan of having crash.log fill up with stuff that isn't really a "crash". We already have that problem with riak_pipe, for example. Anything we can do here?

**EDIT by RZ** - I'm not wild about this either but the HO mgr relies on the exit reason to determine whether or not to consider the xfer complete.  I guess I'd rather leave this alone for now and fix it when we write a new xfer system.

jtuple:
In `riak_core_handoff_sender.erl`, what would you think of modifying `log_fail` so that you could provide a formatting string and args for the reason? Thus, no longer needing to use `io_lib` to construct strings to pass in.

For example:

``` erlang
log_fail(Module, {Type, Opts}, SrcNode, TargetNode, Reason) when is_list(Reason) ->
    log_fail(Module, {Type, Opts}, SrcNode, TargetNode, {"~s", Reason});
log_fail(Module, {Type, Opts}, SrcNode, TargetNode, {ReasonFmt, ReasonArgs}) ->
    lager:error("~p transfer of ~p from ~p ~p to ~p ~p failed " ++ ReasonFmt,
                [Type, Module, SrcNode, SrcPartition,
                 TargetNode, TargetPartition] ++ ReasonArgs).
```

Which would allow you to replace:

``` diff
-ReasonStr1 = io_lib:format("because ~p died with reason ~p",
-                           [From, E]),
-log_fail(Module, {Type, Opts}, SrcNode, TargetNode, ReasonStr1),
+log_fail(Module, {Type, Opts}, SrcNode, TargetNode,
+         {"because ~p died with reason ~p", [From, E]}),
```

It has the nice benefit of using lager's safer string formatting routines for the entire message.

**EDIT by RZ** - Yea I made all this too complex and awkward.  I replaced `log_fail` with a macro and moved the other log statements back into the start_fold code.  Much cleaner now.

jtuple:
Seems like a list comprehension would make `riak_core_vnode_manager:check_up/2` more clear:

``` erlang
check_up(Pairs, UpNodes) ->
    Down = [Pair || {_Partition, Owner}=Pair <- Pairs,
                    not lists:member(Owner, UpNodes)],
    case Down of
        [] -> true;
        _ -> {false, Down}
    end.
```

**EDIT by RZ** - This is cleaner, good call.

jtuple:
It would be great to have eunit tests in `riak_core_repair.erl` that tests range computation.

Also, the range computation logic needs some documentation. Took me a minute to double check it, and I spend a lot of time in chash land.

Here's my notes when I was trying to understand this code:

```
Based on how preflists work, a vnode X is responisble for index X and (nval-1)
predecessor indices.
So, if we have say n-val of 3, and ring [A,B,C,D,E,F]
D is responsible for indices [B,C,D]
In the chash ring an index X is responsible for the hash range [pred(X), X)
Thus: B -> [A,B)
      C -> [B,C)
      D -> [C,D)
This code computes a unified hash range, so for D it would be [A,D)
To compensate for the open-interval, we go with the range [A,D-1]

Note: predecessors(X) returns a list of predecessors of X in
      ring-counterclockwise order, with X as first element hence why
      we decrement the first element
```

Of course, this begs the question of if using GTE and LT (rather than LTE) would have removed the need
for the decrement by 1.

src/riak_core_handoff_manager.erl
jtuple, line 345: Just noticed the changes in the `riak_kv` pull-request, so I guess this is all good.
https://github.com/basho/riak_kv/pull/325/files#L0R279

jtuple:
I've noticed a lot of overlap between `riak_kv_vnode:repair_filter` and `riak_search_vnode:repair_filter`.
I imagine other `riak_core` applications would have similar looking filters. Likewise, much of your n-val
related calculations would also be useful for the upcoming active anti-entropy work. Would it make sense
to refactor this a bit?

My suggestion would be something like the following.

Add a `riak_core_repair:default_filter/5`:

``` erlang
default_filter(Target, Ring, NValMap, DefaultN, InfoFun) ->
    CH = element(4, Ring), %% preferred: riak_core_ring:chash(Ring)
    RangeMap = riak_core_repair:gen_range_map(Target, CH, NValMap),
    Default = riak_core_repair:gen_range(Target, CH, DefaultN),
    RangeFun = riak_core_repair:gen_range_fun(RangeMap, Default),
    fun(ObjectKey) ->
            {Namespace, Hash} = InfoFun(ObjectKey),
            case RangeFun(Namespace) of
                {nowrap, GTE, LTE} ->
                    Hash >= GTE andalso Hash =< LTE;
                {wrap, GTE, LTE} ->
                    Hash >= GTE orelse Hash =< LTE
            end
    end.
```

Re-write `riak_kv_vnode:repair_filter`:

``` erlang
bucket_nval_map(Ring) ->
    [{riak_core_bucket:name(B), riak_core_bucket:n_val(B)} ||
        B <- riak_core_bucket:get_buckets(Ring)].

object_info({Bucket, _Key}=BKey) ->
    Hash = riak_core_util:chash_key(BKey),
    {Bucket, Hash}.

default_object_nval() ->
    riak_core_bucket:n_val(riak_core_config:default_bucket_props()).

repair_filter(Target) ->
    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
    riak_core_repair:default_filter(Target,
                                    Ring,
                                    bucket_nval_map(Ring),
                                    default_object_nval(),
                                    fun object_info/1).
```

Re-write `riak_search_vnode:repair_filter`:

``` erlang
schema_nval_map() ->
    [{S:name(), S:n_val()} || S <- riak_search_config:get_all_schemas()].

object_info({I, {F, T}}) ->
    Hash = riak_search_ring_utils:calc_partition(I, F, T),
    {I, Hash}.

default_object_nval() ->    
    riak_core_bucket:n_val(riak_core_config:default_bucket_props()).

repair_filter(Target) ->
    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
    riak_core_repair:default_filter(Target,
                                    Ring,
                                    schema_nval_map(),
                                    default_object_nval(),
                                    fun object_info/1).
```

**EDIT by RZ** - I like this suggestion.  I went ahead and made the suggested modifications.

jtuple:
As we discussed over chat, the various `gen_server:calls` in this pull-request have several issues. Specifically, the calls from `riak_core_vnode_manager` to `riak_core_handoff_manager`. Under heavy VM load or heavy handoff traffic (primary, secondary, or repair), the call to the handoff manager can timeout (default 5 second timeout).

When the call fails, it throws an exception, killing the vnode manager. When the vnode manager goes down, vnodes may start crashing; and all not-running secondary vnodes go through the usual start-up/check for hinted data/shutdown cycle.

It's unclear if we should just increase the call timeout and live with the occasion failure. Or, go with an infinity timeout and worry about blocking the vnode manager indefinitely (which is actually worse than crashing). Or, take the time to implement a "more proper" solution.

The "right solution" that comes to mind is to not use calls for repair. The repair logic already tolerates handoffs failing and periodically tries to restart them.

What about combining 'xfer' and 'retry_xfer' into a idempotent cast-based API? You call `xfer(SrcIdxNode, ModTgtIdx, FilterFun)` which sends a cast to the handoff manager. The handoff manager checks to see if an existing `ModSrcTgt` handoff is already on-going; if so, ignore the request; otherwise start it. This would allow the vnode manager to reference handoffs by their signature rather than by an instance of `#handoff_status` record that is better left internal to the handoff manager.

The vnode manager then keeps track of what handoff signatures it wants to see happen (mod, src idx/node, tgt idx) and periodically calls `xfer` until `xfer_complete` eventually gets triggered and the signature removed.

**EXIT by RZ** - Joe, I like the async + idempotent idea a lot.  It makes the implementation cleaner and more resilient.  I have this implemented and working although I need to tighten up the `kill_xfer` stuff.  The tests are passing and the code is cleaner.  Thanks for the great idea.

src/riak_core_handoff_manager.erl
rzezeski: You're right, it's not used anymore.  I'll remove it.

src/riak_core_handoff_manager.erl
rzezeski, line 404: Yea, I don't see a situation in which ModSrcTgt does not uniquely label a transfer.  I guess in future if we do great repartition this would not hold any longer, but I'm coding for the code base as it stands now.

src/riak_core_handoff_manager.erl
rzezeski, line 496: Good catch...I think it's kinda lame that inbound doesn't include some information.  I'll look into making this better.

src/riak_core_handoff_sender.erl
rzezeski, line 40: I was very very careful :)

src/riak_core_handoff_manager.erl
rzezeski: Done

src/riak_core_handoff_sender.erl
rzezeski: Done

src/riak_core_vnode_manager.erl
rzezeski: Done

src/riak_core_handoff_manager.erl
rzezeski: Yea, as we discussed in backchannel this pid is not used for the fold request but as a way to monitor the vnode.  Since repair doesn't use the `handoff_complete` msg this isn't a big deal but if doing repair on KV and vnode crashes then the `HPid` (in sender) will fail and the sender will not notice since it's monitor is on the wrong vnode.  Also, msgs like `handoff_error` will go to the wrong place.

I don't see why we use `HPid`.  I think I used to know the reason but can't remember anymore.  I might see if I can't simplify this.

Makefile
rzezeski, line 27: Done.

rzezeski:
FYI, I'm currently making all my fixes/improvements on https://github.com/basho/riak_core/compare/rz-repair-cmd-3

include/riak_core_handoff.hrl
rzezeski: Actually, it's only used in one place and it cannot be `repair`.  I know...this is confusing.

include/riak_core_handoff.hrl
rzezeski: Yea, I'll make this change so things are less confusing.

include/riak_core_handoff.hrl
jtuple: Technically, the one place that uses `ho_type()` cannot be repair (eg. `send_handoff`). But, there are other places in the code where you use `Type` and `HOType` that don't have specs but are of the spec `ownership_handoff | hinted_handoff | repair`. Example is `#handoff_status.type` noted below, but also the `Type` in `{Type, Opts}` of `riak_core_handoff_sender`.

rzezeski: Yea I ended up changing it to be like you said.

include/riak_core_handoff.hrl
rzezeski: I changed the type name and removed `handoffs()` since it was used in only one place.

src/riak_core_vnode_manager.erl
rzezeski: Done.

src/riak_core_repair.erl
rzezeski, line 73: Done

src/riak_core_handoff_manager.erl
rzezeski: I've taken your advice and make this async + idempotent.

src/riak_core_handoff_manager.erl
rzezeski: With the new async + idempotent xfer this function is no longer used/needed.  Problem solved.

src/riak_core_handoff_manager.erl
rzezeski: I've changed this to be a cast/fire-and-forget.  However, that now has the problem of being lost and not actually being killed.  The other thing I need to do is have the vnode mgr on the target node (which initiated this call) kill any incoming repair xfers.  The problem is this data is not available for incoming xfers because of the way the handoff mgr and handoff receiver are implemented.  However, I can make it work.  In fact it might be easier to just kill the receivers and not even worry about sending this msg, although it will look like an error on the sender side.  Perhaps I could trip a flag in the receiver and send back a special ack msg that tells sender to shutdown.  In any case, this call is going away.

src/riak_core_handoff_manager.erl
rzezeski: Yea, it is assumed to only kill outbound and will only call outbound.  The reason there is no type info is because the HO mgr `add_inbound` is called by the handoff listener.  The relevant info is passed to the HO receiver which doesn't communicate with the HO mgr at all.  I've thought about setting up calls between recv and mgr to share this info but not sure yet.

src/riak_core_handoff_manager.erl
rzezeski, line 345: The reason for this is so that in a mixed cluster the upgraded nodes can differentiate between old/new status and display both.  If you have a cleaner way happy to hear it.

src/riak_core_handoff_manager.erl
rzezeski, line 496: I made my previous comment too early.  I already explained in one of the other comments why this is not set.  I still think it's crap but it will take a little work to have more info on inbound xfers.  This doesn't hurt anything, though.

src/riak_core_vnode_manager.erl
rzezeski, line 438: Good catch.

src/riak_core_vnode_manager.erl
rzezeski, line 95: Great points.  I've made these adjustments in the new branch.

src/riak_core_vnode_manager.erl
rzezeski: Done.

src/riak_core_vnode_manager.erl
rzezeski: Done.

src/riak_core_vnode_manager.erl
rzezeski: You're right, I think both should just log and unexpected case and not crash.

src/riak_core_vnode_manager.erl
rzezeski, line 296: See below.

src/riak_core_bucket.erl
rzezeski, line 117: I agree, that would be nice.  Not sure if I'll get to it given I'm trying to finish today.

rzezeski:
I addressed all but a few comments.

jtuple:
Took me a second to get proper diff'ing to show changes since my last review given the rebase on top of master. For reference, this is the changeset I'm giving a second review over:

https://github.com/basho/riak_core/compare/21a55fa88cde43d89dbbeccefb934e6b4fb2d9e6...rz-repair-cmd-2

slfritchie:
About the comment at https://github.com/basho/riak_core/compare/21a55fa88cde43d89dbbeccefb934e6b4fb2d9e6...rz-repair-cmd-2#L2R518

```
%% @doc Kill and remove _each_ xfer associated with `ModSrcTarget'
%%      with `Reason'.  There might be more than one because repair
%%      can have two simultaneous inbound xfers.
```

Rather than "two", should it be "n_val - 1"?

SLF edit: Hrm, there really are only two repairs, one from the vnode adjacent on the "left" and one from the vnode adjacent on the "right" of the vnode under repair?

rzezeski:
It is always only two, `P-1` and `P+1` where `P` is the partition under repair.  Depending on the `n_val` a wider range of partitions could act as a source of repair but this was the easiest thing to do for now and will work for all `n_val`.  Except 1, of course.

slfritchie:
Well, taking n_val = 5 as an example, then P-2 and P+2 could have data necessary to repair P.

It's also occurred to me that it a repair would be less valuable if there is a pending handoff from SomeOtherVnode->(P+1 or P-1).  For maximum repair, that handoff should finish before doing the repair.  Or when the repair is started, if such a handoff is pending, warn the user?  {shrug}

rzezeski:
1) n_val = 5 - yes but the replicas in `P-2` are in `P-1` and the
replicas in `P+2` are in `P+1`.  In the test I built `N_VAL=5`
passes.

2) Yes, if there is handoff then some of that data (or all of it
depending on the way the fold works) will be missing.  Perhaps
there should be a stronger check when starting repair to only
work when the src partitions are not accepting handoff, of course
that check could pass but only because the fallback vnodes
haven't hit inactivity yet.  Technically, wouldn't we have to
check every potential fallback to make absolutely sure?  Do we
have the same problem with listkeys on a bucket?  I'm not sure I
have a great answer here but I'm also not sure it's a concern for
this cut of repair.

slfritchie:
n_val > 3 ... with the current repair implementation, it's even more important to have all/as-many-as-possible handoffs finished before the repair starts.  (A concrete scenario that I'm thinking of: n_val=5, P-1 and P+1 aren't available at the time some BKey is put (and thus are sitting in non-primary vnodes elsewhere), then P suffers catastrophe and loses its state.  If the non-primary vnodes haven't handed off yet, then a repair using only P-1 and P+1 won't get BKey stored on P.)

About handoff before repair ... such a requirement would be useful in a Perfect World, but this revision is known & accepted as Not Yet Perfect, and I'm fine with that.

rzezeski:
Yes, your scenario is a very real one.  I think the two issues
are a) actually knowing if a primary has a fallback lurking
somewhere (this is even more hairy when a node with a fallback
crashes because it is not guaranteed to start immediately but
only at some time in the future) and b) synchronizing multiple
source partitions (running P-2 and P-1 is and P+1 and P+2 is
gonna be more expensive for potentially no gain in some
circumstances).  In my mind repair is a best effort process that
is more efficient than list keys + read-repair and easier to to
use/explain.  As you said, a NYP version.  Running a repair
multiple times on a partition should be idempotent for KV and
Search but require CPU/IO cycles.  I guess a concern would be
setting expectations for users that a successful repair does
still not always guarantee consistent replicas.  Even outside
repair, that is a hard thing to determine in a non-quiescent
cluster.  My point...try to offer something better than what we
currently have but still NYP.

jtuple:
Testing revealed some issues with repair handoff not being restarted after certain error conditions (handoff receiver dying or receiving vnode dying). After talking to @rzezeski, I slightly re-worked the handoff sender catch condition and pushed those changes out. Manual testing shows all failures being handled properly, with repair being restarted at the next timer tick.

There are still a few issues. If handoff_concurrency is too low, then transfers will fail indefinitely. However, this is all logged, and users should be able to see the error and increase concurrency. This is easier in 1.2 given the new `riak-admin transfer-limit` command.

There is also a mis-match between the sender and receiver object count numbers, likely due to the count on the sender not being filter aware. Example:

```
[info] repair transfer of riak_kv_vnode from 'dev1@127.0.0.1' 22835963083295358096932575511191922182123945984 to 'dev1@127.0.0.1' 0 completed: sent 1865 objects in 0.20 seconds
[info] Handoff receiver for partition 0 exited after processing 1248 objects
```

There is also a slight potentially confusing naming mismatch. Some message use source index in output, others use target index. If I am repairing partition 0 and partition X is a sibling that sends the repair data, then partition X generates log messages on failure that talk about "outbound transfer of partition 0" (target) rather than "partition X" (source). Probably should be "transfer of partition X to partition 0". Example:

```
[info] Starting repair transfer of riak_kv_vnode from 'dev1@127.0.0.1' 1438665674247607560106752257205091097473808596992 to 'dev1@127.0.0.1' 0
[error] An outbound handoff of partition riak_kv_vnode 0 was terminated because the vnode died
```

All the important issues I raised in the last review have been addressed. The code is much more solid now. I triggered hundreds of concurrent repairs and things worked just fine. Manually killed handoff senders, receivers, and vnodes during repairs and things worked well. I tested building mixed-clusters and performing simple upgrades, and things worked as expected. So, basic mixed-version cluster membership + handoff looks good. We will want to make sure this goes the full rounds during integration/release testing, and it wouldn't hurt for me to add repair to the EQC dynamic test framework.

All in all, area for future improvement, but solid and useful enough to make it into this release. Feel free to fix the error messages mentioned above now if you want, but could just as well be done down the road.

+1 merge

PR #176, opened 2012-05-17
Handoffs on a different network

micmac:
This small change is intended to make riak handle the situation when a node's handoff ip is different from the ip associated with its node name. Without this feature, separating handoffs from other cluster communication is complicated and involves advanced routing configuration.

### DIFF
--- a/src/riak_core_handoff_listener.erl
+++ b/src/riak_core_handoff_listener.erl
@@ -29,6 +29,7 @@
          terminate/2, code_change/3]).
 -export([sock_opts/0, new_connection/2]).
 -record(state, {
+          ipaddr :: string(),
           portnum :: integer(),
           ssl_opts :: list()
          }).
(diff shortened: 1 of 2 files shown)


Vagabond:
This isn't safe in the case that the handoff_ip is configured to 0.0.0.0.

micmac:
Yes it has to be fixed for that case (and for the case when it's unset if that's different). Unfortunately I don't speak erlang but need the functionality (and was told by Basho support in the last few weeks that it's in the code) so I patched it then support asked for a pull request. I hope they'll review it and add the necessary 2-3 lines before integrating.

jtuple:
Andrew brought up the only major point I had with the initial version, the 0.0.0.0 case. Since that has been handled, and two Basho engineers (Andrew and I) have now looked over the code, I'm going to go ahead and wrap this up.

I've added a test case to our internal Riak test suite that verifies the correct `handoff_ip` behavior as well as verifies that "0.0.0.0" is handled properly. I've merged this code into an integration branch, and modified it slightly to remove a minor bug and clean-up the exceptional condition in `get_handoff_ip`. Otherwise, all looks good.

Going to merge this into master, and will also look into back-porting it to 1.1.2 in case any existing customers (such as yourself/Prezi), need a hot-patch for the current release.

Thanks for the pull-request.

Issue #177, opened 2012-05-18
Make cluster admin a two-phase (plan/commit) process


jtuple:
The way Riak currently provides cluster administration through `riak-admin` and Riak Control is a fire-and-pray model. For example, you issue `riak-admin join` to join a node to a cluster, and the join is immediately scheduled. It is impossible to atomically add multiple nodes at once, or to perform both joins and leaves at the same time. Instead, Riak immediately calculates the new partition ownership and begins transferring data to the new node. Furthermore, it is impossible to see how the change will affect the cluster until after issuing the join/leave, but by that point there is no way to cancel/stop things if the changes are undesired (eg. join a node and suddenly 128+ partitions become schedule during peak traffic).

Let's move away from this approach, and move towards a two-phase approach to cluster changes. Rather than having join/leaves/etc happen immediately, issuing such commands should instead stage pending changes to the cluster. A user should then be able to issue a command such as `riak-admin plan` to print out the staged changes -- the list of changes, the cluster membership/ring ownership resulting from these changes, the number of transfers that would be required, etc. If the user is satisfied, they could then `riak-admin commit` the plan, and the entire plan would be issued, and transfers scheduled. Otherwise, the user could continue to add/leave additional nodes, or `riak-admin clear` to clear the entire set of staged changes.

jtuple:
Feature completed, reviewed, and merged into master. Will be available in Riak 1.2

PR #181, opened 2012-05-25
Add two-phase/staged approach to cluster administration 

jtuple:
This pull-request adds staged clustering as discussed in issue #177. This new approach to cluster administration operates in two phases. Users first stage intended cluster changes (join/leave/remove), then commit the changes when ready and satisfied with how those changes will impact the cluster.

This pull-request also adds the new `replace` and `force-replace` operations.

The `replace node new-node` operation stages a request for `node` to handoff all of its partitions to `new-node` and then leave the cluster. This is similar to a `leave`, but directs handoff to a specific replacement node.

The `force-replace node new-node` operation stages a request that reassigns all partitions owned by `node` to `new-node` and removes `node` from the cluster. This is similar to a `force-remove`, but reassigns partitions to a specific replacement node.

This feature addition is implemented through a series of refactoring and code additions, and is best reviewed looking at the history of the commits rather than just at the resulting code.

The first commit moves the claimant logic from `riak_core_ring.erl` into a new `riak_core_claimant.erl`. This commit is more or less a straightforward cut/paste and intentionally doesn't compile cleanly.

The second commit than refactors `riak_core_ring.erl` and `riak_core_claimant.erl` so that things compile and work as before the code split. The diff between these two commits should make it clear what was changed. After review, these two commits will be squashed together to ensure there is no broken commit in the history.

Note: only changes to the claimant code itself should be reviewed, the pre-existing claimant code has long since been reviewed. Although, future documentation and code refactoring is planned -- but outside the scope of this current pull-request.

The next four commits add the staged clustering code. The first refactors the claimant code so that it can be run without side-effects. The second adds the staged clustering code itself, the third adds the new `replace` and `force-replace` operations, and the fourth fixes an issue with older non-staged join operation.

There is a related `basho/riak` pull-request that updates `riak-admin` to support the new staged approach.

There is also a related `basho/riak_test` pull-request that implements an appropriate integration test.


Vagabond:
The code in here all looks fine. +1 for this part.

Vagabond:
If I set the transfer limit to 0 across the cluster, I can still join/leave nodes and partitions will be transferred. That seems wrong, no?

Vagabond:
Also, negative numbers are allowed as the value for transfer_limit. Also, passing non-numeric input yields an ugly backtrace to the console.

Vagabond:
I'm +1 on the staged cluster stuff, but the transfer_limit stuff seems flaky. Someone else needs to review it, as I'll be gone.

jtuple:
If you set transfer limit to 0, you shouldn't see transfers occur. Keep in mind that newly started nodes will need to have their transfer limits set to 0 as well. What you may have been observing is ownership behavior in an empty (or near empty) cluster. The transfer limit prevents transfers from occurring, not ownership from changing. An empty partition that is changing owners will change owners regardless of the transfer limit setting -- there is no transfer of data necessary, so ownership will just be reassigned after the appropriate transitions / handshakes.

jtuple:
Added code that should validate provided transfer limit, preventing negative numbers as well non-numeric values.

Vagabond:
Ah yes, I didn't have much, or any, data in the cluster, so that explains the behaviour I saw.

jonmeredith:
transfer limits validation works for me, +1

jonmeredith:
I've taken a run at merging the claimv3/sim code with the staged claim on jdm-merge-staged-clustering, I made the branch off current master and did a git merge-no-ff of the branch on this pull request.

I ended up exposing remove_indices in riak_core_claimant so that sim code could cause that directly rather than doing the refactor to use more of the claimant code.

There were three issues during the merge
1) Conflict between the capabilities stuff and autojoin stuff at https://github.com/basho/riak_core/blob/jdm-merge-staged-clustering/src/riak_core.erl#L146  - I guessed do capability first, then join, but maybe that's wrong.
2) There was a commit 07e68d6d8982e90decc2591602c410ef74236a4d that cleared metadata from the ring on exit and I couldn't see that on the new branch.  Is that a fix from 1.1.x that should have been merged in?
3) not really a merge issue, but I noticed the arguments for https://github.com/basho/riak_core/blob/jdm-merge-staged-clustering/src/riak_core_claimant.erl#L874 were swapped from the next call, it's harmless at the moment, but it could bite somebody who changes it.

The merged branch passes riak_test verify_build_cluster and verify_staged_clustering, but it fails on verify_{claimant,leave,down} tests (as does this gh177-staged-clustering branch for me).

jtuple:
All integration on branch `jdm-merge-staged-clustering` is done. Merging that branch into master and pushing.

PR #182, opened 2012-05-25
Add `app_helper:try_envs`

rzezeski:
This function allows you to try multiple application envs in a
particular order and returns a value that allows the caller to
distinguish where the value came from.  This is useful for the cases
where application envs have changed across releases.


seancribbs:
:+1:

PR #183, opened 2012-05-25
ClaimV3

jonmeredith:
Claim V3 - unlike the v1/v2 algorithms, v3 treats claim as an optimization problem. In it's current form it creates a number of possible claim plans and evaluates them for violations, balance and diversity, choosing the 'best' plan.

Violations are a count of how many partitions owned by the same node are within target-n of one another. Lower is better, 0 is desired if at all possible.

Balance is a measure of the number of partitions owned versus the number of partitions wanted.  Want is supplied to the algorithm by the caller as a list of node/counts.  The score for deviation is the RMS of the difference between what the node wanted and what it has.  Lower is better, 0 if all wants are mets.

Diversity measures how often nodes are close to one another in the preference list.  The more diverse (spread of distances apart), the more evenly the responsibility for a failed node is spread across the cluster.  Diversity is calculated by working out the count of each distance for each node pair
(currently distances are limited up to target N) and computing the RMS on that. Lower diversity score is better, 0 if nodes are perfectly diverse.

For testing and playing around, you may want to look at running riak_core_claim_sim:commission(). which will run comparisons against the v1/v2 claim algorithm for some preset combinations of ring size, number of nodes, and target n.

You could also play with riak_core_claim_sim:print_failure_analysis(Ring, TargetN, NumFailures) which prints reports on the console for how a ring does given a target N and number of nodes to fail.

This code is based on the jdm-claim-sim branch - conflicts with the cluster management changes will need to be resolved between the three branches before we can merge.

### DIFF
--- a/src/riak_core_claim.erl
+++ b/src/riak_core_claim.erl
@@ -63,7 +63,7 @@
          choose_claim_v1/1, choose_claim_v1/2, choose_claim_v1/3,
          choose_claim_v2/1, choose_claim_v2/2, choose_claim_v2/3,
          choose_claim_v3/1, choose_claim_v3/2, choose_claim_v3/3,
-         claim_rebalance_n/2,
+         claim_rebalance_n/2, claim_diversify/3, claim_diagonal/3,
          meets_target_n/2,
          diagonal_stripe/2]).
 
@@ -364,7 +364,7 @@ choose_claim_v3(Ring, _ClaimNode, Params) ->
     TN = proplists:get_value(target_n_val, Params, 4),
     Active = riak_core_ring:active_members(Ring),
     Wants = lists:zip(Active, wants(length(Active), Q)),
-    lager:debug("Claim3 started: Q=~p TN = ~p\n", [Q, TN]),
+    lager:debug("Claim3 started: S=~p Q=~p TN=~p\n", [S, Q, TN]),
     lager:debug("       wants: ~p\n", [Wants]),
     {Partitions, Owners} = lists:unzip(riak_core_ring:all_owners(Ring)),
 
@@ -414,7 +414,7 @@ claim_v3(Wants, Owners, Params) ->
     Q = length(Owners),
     Claiming = [N || {N,W} <- Wants, W > 0],
     Trials = proplists:get_value(trials, Params, 100),
-    case length(Claiming) >= TN of
+    case length(Claiming) > TN of
         true ->
             NIs = build_nis(Wants, Owners),
 
@@ -431,13 +431,15 @@ claim_v3(Wants, Owners, Params) ->
                 0 ->
                     New;
                 _ ->
-                    %% If could not build ring without violations, diversify it
+                    lager:debug("claimv3: Could not make plan without violations, diversifying\n",
+                                []),
+                   %% If could not build ring without violations, diversify it
                     claim_diversify(Wants, Owners, Params)
             end;
         false ->
-            lager:debug("claimv3: Not enough nodes to run (have ~p need ~p), diversifying\n",
+            lager:debug("claimv3: Not enough nodes to run (have ~p need ~p), diagonalized\n",
                         [length(Claiming), TN]),
-            claim_diversify(Wants, Owners, Params)
+            claim_diagonal(Wants, Owners, Params)
     end.
                 
 %% Claim diversify tries to build a perfectly diverse ownership list that meets
@@ -453,6 +455,25 @@ claim_diversify(Wants, Owners, Params) ->
                              riak_core_claim_util:gen_complete_len(Q), Claiming, TN),
     {NewOwners, [diversified]}.
 
+%% Claim nodes in seq a,b,c,a,b,c trying to handle the wraparound
+%% case to meet target N
+claim_diagonal(Wants, Owners, Params) ->
+    TN = proplists:get_value(target_n_val, Params, 4),
+    Claiming = lists:sort([N || {N,W} <- Wants, W > 0]),
+    S = length(Claiming),
+    Q = length(Owners),
+    Reps = Q div S,
+    %% Handle the ring wrapround case.  If possible try to pick nodes
+    %% that are not within the first TN of Claiming, if enough nodes
+    %% are available.
+    Tail = Q - Reps * S,
+    Last = case S >= TN + Tail of
+               true -> % If number wanted can be filled excluding first TN nodes
+                   lists:sublist(lists:nthtail(TN - Tail, Claiming), Tail);
+               _ ->
+                   lists:sublist(Claiming, Tail)
+           end,
+    {lists:flatten([lists:duplicate(Reps, Claiming), Last]), [diagonalized]}.
 
 claim_rebalance_n(Ring0, Node) ->
     Ring = riak_core_ring:upgrade(Ring0),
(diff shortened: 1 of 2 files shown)


src/riak_core_claim.erl
```diff
@@ -327,6 +382,130 @@ meets_target_n([], TargetN, Index, First, Last) ->
                      Last),
     {true, [ Part || {_, _, Part} <- Violations ]}.
 
+
+choose_claim_v3(Ring) ->
+    choose_claim_v3(Ring, node()).
+
+choose_claim_v3(Ring, ClaimNode) ->
+    Params = [{target_n_val, app_helper:get_env(riak_core, target_n_val, 3)}],
```
russelldb: In the riak_core.app file, and in default_choose_params/1 this is target_n_val is 4 (https://github.com/basho/riak_core/blob/a73fcfda4a819c19eac2f93d3d71f0aa8b60158e/src/riak_core_claim.erl#L247) though the docs at the top of the file do say default target_n_val is 3…

src/riak_core_claim.erl
```diff
@@ -182,6 +184,59 @@ wants_claim_v2(Ring, Node) ->
...
+                true ->
+                    application:unset_env(riak_core, force_reclaim),
+                    lager:info("Forced rerun of claim algorithm - "
+                               "unsetting force_reclaim"),
+                    {yes, 1};
+                false ->
+                    %% Otherwise, base wants decision on whether the current 
+                    %% wants versus current ownership if the claim does not
+                    %% manage to claim all requested nodes then the temporary
+                    %% 'claim_v3' metadatawill stop the loop
```
russelldb: 'metadata will'

src/riak_core_claim.erl
```diff
@@ -570,6 +753,306 @@ spaced_by_n(NthA, NthB, TargetN, RingSize) ->
...
+%% overlaoded by (negative)
+wants_owns_diff(Wants, Owns) ->
+    [ case lists:keyfind(N, 1, Owns) of
+          {N, O} ->
+              {N, W - O};
+          false ->
+              {N,W}
+      end || {N, W} <- Wants ].
+    
+%% Given a ring, work out how many partition each wants to be
```
russelldb, line 779: "…how many partitions each _node_ wants…"?

src/riak_core_claim.erl
jonmeredith: We should probably set at 4 until we have a filter on the fallback preflists to ensure node uniqueness as long as possible.

src/riak_core_claim.erl
```diff
@@ -327,6 +382,130 @@ meets_target_n([], TargetN, Index, First, Last) ->
...
+%%
+%% Diversity measures how often nodes are close to one another in the preference
+%% list.  The more diverse (spread of distances apart), the more evenly the
+%% responsibility for a failed node is spread across the cluster.  Diversity is
+%% calculated by working out the count of each distance for each node pair
+%% (currently distances are limited up to target N) and computing the RMS on that.
+%% Lower diversity score is better, 0 if nodes are perfectly diverse.
+%%
+claim_v3(Wants, Owners, Params) ->
+    TN = proplists:get_value(target_n_val, Params, 4),
```
russelldb: This is the 3rd of fourth time this line appears, suggest either/both that the default target_n_val be in a macro (DEF_TARGET_N maybe) as it is fetched so often or a local private fun get_target_n_val(). Kind of thing that is easy to miss in one instance when we change it, and weird things ensue.

In fact, if I was a refactoring tool (hehehe tool) I'd suggest that these 3 lines appear again and again and would be a good candidate for a function

```
TN = proplists:get_value(target_n_val, Params, 4),
Q = length(Owners),
Claiming = [N || {N,W} <- Wants, W > 0],
```

I know this is trivial, just wanted to bring it up.

src/riak_core_claim.erl
```diff
@@ -327,6 +382,130 @@ meets_target_n([], TargetN, Index, First, Last) ->
...
+                0 ->
+                    New;
+                _ ->
+                    lager:debug("claimv3: Could not make plan without violations, diversifying\n",
+                                []),
+                   %% If could not build ring without violations, diversify it
+                    claim_diversify(Wants, Owners, Params)
+            end;
+        false ->
+            lager:debug("claimv3: Not enough nodes to run (have ~p need ~p), diagonalized\n",
```
russelldb, line 474: I found this confusing, with debug enabled, 4 nodes and a target_n_val of 4 this prints

```
[debug] claimv3: Not enough nodes to run (have 4 need 4), diagonalized
```

jonmeredith, line 474: I'll fix the error message.  I was trying to say >4.  'needs more than 5' would probably do it.

src/riak_core_claim.erl
```diff
@@ -327,6 +384,131 @@ meets_target_n([], TargetN, Index, First, Last) ->
...
+%%
+%% Balance is a measure of the number of partitions owned versus the number of partitions
+%% wanted.  Want is supplied to the algorithm by the caller as a list of node/counts.  The
+%% score for deviation is the RMS of the difference between what the node wanted and what it 
+%% has.  Lower is better, 0 if all wants are mets.
+%%
+%% Diversity measures how often nodes are close to one another in the preference
+%% list.  The more diverse (spread of distances apart), the more evenly the
+%% responsibility for a failed node is spread across the cluster.  Diversity is
+%% calculated by working out the count of each distance for each node pair
```
russelldb, line 442: sorts here and before returning from make_plans. Does it matter?

src/riak_core_claim.erl
jonmeredith: It does get repeated a fair bit, but I suppose they didn't seem like a cohesive thing to do in a single function.

```
{TN, Q, Claiming} = do_common_claim_stuff(Params, Owners, Wants)
```

src/riak_core_claim.erl
```diff
@@ -570,6 +756,306 @@ spaced_by_n(NthA, NthB, TargetN, RingSize) ->
...
+
+    Balance = balance(Wants, NIs),
+
+    Violations = length(lists:flatten([Vs || {_, _, _, Vs} <- violations(NIs, Q, TN),
+                                             length(Vs) > 0])),
+
+    {Owners, [{violations, Violations}, {balance, Balance}, {diversity, Diversity}]}.
+
+%% @private
+%% Convert a nodeinfo list with Nodes and lists of owned indices to an
```
russelldb, line 861: Don't like the change from NIs, Q, TN as names briefly here, (changed from call to make_plans in claim_v3 and changed back below in make_plan) just makes reading a little harder

src/riak_core_claim.erl
russelldb: Yeah, probably just my Java brain expecting a CommonClaimParams class or something horrible like that.

src/riak_core_claim.erl
```diff
@@ -570,6 +756,306 @@ spaced_by_n(NthA, NthB, TargetN, RingSize) ->
...
+                    end, {1, {undefined, undefined}}, Plans),
+    FOM.
+
+%% @private
+%% Return true if plan P1 is better than plan P2, assumes the metrics
+%% are ordered [{violations, Violations}, {balance, Balance}, {diversity, Diversity}]}.
+%%
+better_plan(_M1, undefined) ->
+    true; %% M1 is a better plan than no plan
+better_plan(M1, M2) ->
```
russelldb, line 832: When using a 5 node cluster with a ring size of 64, this reports 10 violations. I think there is just one though (that indexes 0, 3 are too close together on some node). Claim still works, but the output in the debug statement in clam_v3 is confusing because of it.

src/riak_core_claim.erl
```diff
@@ -661,5 +1151,163 @@ prop_claim_ensures_unique_nodes() ->
...
+                   check_deltas(Exchanges, NIs, NIs2, Q, TN))
+                   %% conjunction([{len, equals(length(NIs), length(NIs2))},
+                   %%              {delta, check_deltas(Exchanges, NIs, NIs2, Q, TN)}]))
+            end).
+
+tnode(I) ->
+    list_to_atom("n"++integer_to_list(I)).
+    
+%% Check that no node gained more than it wanted to take
+%% Check that none of the node took any partitions that
```
russelldb: That…

russelldb:
A few comments about comments and surprising debug output above. Tested using the provided eqc tests, and a bit of prodding by hand. 

Wondering if we want to add metrics for time to claim, and maybe the claim "metrics" (diversified / diagonal / violations etc)?

All works for me +1 to merge.

src/riak_core_claim.erl
jonmeredith, line 861: Missed it in a refactor I think.

jonmeredith:
rebased and merged

PR #184, opened 2012-05-30
Add riak_core_claim:never_wants_claim/2

jtuple:
Add 2-arity version of never_wants_claim because the claimant logic introduced back in Riak 1.0 uses 2-arity claim functions rather than the older 1-arity functions.

This is necessary for users to disable partition claim on their clusters, which is something we commonly do during support scenarios (add we've manually hot-patched this 2-arity fun into production clusters in recent months to do just that, given that it was missing in our releases).


russelldb:
+1.

jonmeredith:
I think I also added this to the v3 claim branch too.

jonmeredith:
merged version as part of claimv3

PR #188, opened 2012-06-11
Eunit cleanups

lemenkov:
This pull request consists of two small patches. One just removes duplicated eunit inclusion. Another one ensures that all eunit includes are surrounded by -if..-endif block to prevent compiling with implicit compile(all) directive when it's not required etc. Both patches looks to be pretty safe to apply.


rzezeski:
Looks good to me, it compiles and tests pass.

PR #191, opened 2012-06-14
Fix incorrect merge of claim deltas and default

jtuple:
Fix claim_v2 regression uncovered by EQC test and noted by @rzezeski


rzezeski:
Regression fixed, +1 to merge

PR #193, opened 2012-06-19
Fix capability bug and fix staged clustering in mixed-version clusters

jtuple:
This pull-request fixes two bugs.

Capabilities registered without legacy app var compatibility logic were broken, triggering a badmatch. This pull-request fixes this issue.

Joining a pre-staged clustering node to a newer cluster would result in the node joining the cluster in `joining` state but never transitioning to `valid`. This has also been fixed, by defaulting to 'auto-join' behavior unless a new capability `staged_joins` is negotiated.


rzezeski:
+1 to merge

PR #194, opened 2012-06-20
Cache stats in ets, only allow one call to Mod:Stat_fun at a time

russelldb:
Moving to folsom and removing the bottle necky gen_servers for stats means that many concurrent calls to calculate stats are no longer serialized. This PR provides a cache that also ensures only a single call to calculate stats executes at a time. Requests that arrive while stats are being calculated will be notified with the currently executing calls results, rather than calling the calculate code again.

Also, have riak_core_stat spawn calls to folsom.

### DIFF
--- a/src/riak_core_stat.erl
+++ b/src/riak_core_stat.erl
@@ -20,23 +20,62 @@
 
 -module(riak_core_stat).
 
+-behaviour(gen_server).
+
 %% API
--export([get_stats/0, update/1, register_stats/0, produce_stats/0]).
+-export([start_link/0, get_stats/0, update/1,
+         register_stats/0, produce_stats/0]).
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
+-define(SERVER, ?MODULE).
+
 -define(APP, riak_core).
+
+start_link() ->
+    gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
 
 %% @spec get_stats() -> proplist()
 %% @doc Get the current aggregation of stats.
 get_stats() ->
-    riak_core_stat_cache:get_stats(?APP).
+    case riak_core_stat_cache:get_stats(?APP) of
+        {ok, Stats, _TS} ->
+            Stats;
+        Error -> Error
+    end.
 
 update(Arg) ->
-    spawn(fun() ->
-                 update1(Arg) end).
+    gen_server:cast(?SERVER, {update, Arg}).
+
+
+%% gen_server
+
+init([]) ->
+    {ok, ok}.
+
+handle_call(_Req, _From, State) ->
+    {reply, ok, State}.
+
+handle_cast({update, Arg}, State) ->
+    update1(Arg),
+    {noreply, State};
+handle_cast(_Req, State) ->
+    {noreply, State}.
+
+handle_info(_Info, State) ->
+    {noreply, State}.
+
+terminate(_Reason, _State) ->
+    ok.
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
 
 %% @spec update(term()) -> ok
 %% @doc Update the given stat.
--- a/src/riak_core_stat.erl
+++ b/src/riak_core_stat.erl
@@ -41,6 +41,10 @@
 start_link() ->
     gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
 
+register_stats() ->
+    [register_stat({?APP, Name}, Type) || {Name, Type} <- stats()],
+    riak_core_stat_cache:register_app(?APP, {?MODULE, produce_stats, []}).
+
 %% @spec get_stats() -> proplist()
 %% @doc Get the current aggregation of stats.
 get_stats() ->
@@ -52,7 +56,6 @@ get_stats() ->
 
 update(Arg) ->
     gen_server:cast(?SERVER, {update, Arg}).
-
 
 %% gen_server
 
@@ -103,10 +106,6 @@ update1(rebalance_timer_begin) ->
     folsom_metrics:notify_existing_metric({?APP, rebalance_delay}, timer_start, duration);
 update1(rebalance_timer_end) ->
     folsom_metrics:notify_existing_metric({?APP, rebalance_delay}, timer_end, duration).
-
-register_stats() ->
-    [register_stat({?APP, Name}, Type) || {Name, Type} <- stats()],
-    riak_core_stat_cache:register_app(?APP, {?MODULE, produce_stats, []}).
 
 %% private
 stats() ->
--- a/src/riak_core_stat.erl
+++ b/src/riak_core_stat.erl
@@ -21,7 +21,7 @@
 -module(riak_core_stat).
 
 %% API
--export([get_stats/0, update/1, register_stats/0]).
+-export([get_stats/0, update/1, register_stats/0, produce_stats/0]).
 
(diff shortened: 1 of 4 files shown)


src/riak_core_stat.erl
```diff
@@ -32,37 +32,42 @@
 %% @spec get_stats() -> proplist()
 %% @doc Get the current aggregation of stats.
 get_stats() ->
-    produce_stats().
+    riak_core_stat_cache:get_stats(?APP).
+
+update(Arg) ->
+    spawn(fun() ->
+                 update1(Arg) end).
```
beerriot: This sort of un-metered, un-monitored spawning makes me uneasy.  I know one goal of these changes was to remove serialization on a `gen_server`, but it seems like finding the process with the large mailbox will be easier than sorting out thousands of throw away processes, while troubleshooting a node running low on resources. I suppose the key questions are, how often do we expect this function to be called, and how often might it be called in the worst case?

russelldb: I don't like it either. I think a gen_server would be better since there is no longer any blocking gen_server:call to calculate stats, so unlikely to be any large mailbox.

Ideally a process per stat.

This has been tested up to a a few thousand ops a second (each op will call update 2 or 3 times) and seems OK.

To move back to a gen_server is simple, a process per stat for the next release maybe. The amount of load testing I need to do when I change this means a full day to verify that the gen_server approach is better. I went with expediency in this case. Happy to change if there is time.

russelldb: There is a lot of history. Went from gen_sever to process per stat, to no process at all (just call folsom) to all these spawns. With 3 different metrics providers in the back ground. All load tested for 10s of hours. None of it satisfactory.

src/riak_core_stat_cache.erl
```diff
@@ -0,0 +1,137 @@
...
+
+register_app(App, {M, F, A}) ->
+    TTL = app_helper:get_env(riak_core, stat_cache_ttl, ?TTL),
+    register_app(App, {M, F, A}, TTL).
+
+register_app(App, {M, F, A}, TTL) ->
+    gen_server:call(?SERVER, {register, App, {M, F, A}, TTL}).
+
+get_stats(App) ->
+    gen_server:call(?SERVER, {get_stats, App}).
```
russelldb: I think a timeout, infinity, or maybe somewhat shorter, is needed here. I've done some testing on Joyent where it can take 10 seconds or more to calculate the stats.

src/riak_core_stat_cache.erl
```diff
@@ -0,0 +1,137 @@
...
+
+init([]) ->
+    Tab = ets:new(?MODULE, [public, set, named_table]),
```
beerriot: It looks like only this process ever touches this table, so maybe it should be `private` instead of `public`. Or at least `protected` if you want to read it out-of-band for debugging, but still protect against concurrent writes.

src/riak_core_stat_cache.erl
```diff
@@ -0,0 +1,137 @@
...
+
+%% API
+-export([start_link/0, get_stats/1, register_app/2, register_app/3]).
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+-define(SERVER, ?MODULE).
+-define(TTL, 5).
```
beerriot, line 41: a small doc about what units `?TTL` are in would be good

src/riak_core_stat_cache.erl
```diff
@@ -0,0 +1,137 @@
...
+                   folsom_metrics:new_meter({?MODULE, App}),
+                   orddict:store(App, {Mod, Fun, Args, TTL}, Apps0);
+               {true, _} ->
+                   Apps0
+           end,
+    {reply, ok, State0#state{apps=Apps}};
+handle_call({get_stats, App}, From, State0=#state{apps=Apps, active=Active0, tab=Tab}) ->
+    Reply = case registered(App, Apps) of
+                false ->
+                    {reply, {enotregistered, App}, State0};
```
beerriot: Where does `enotregistered` come from? It might be better for code consistency to use `error` instead (maybe `{error, {enotregistered, App}}`), unless that atom is part of another interface this cache is mimicking.

src/riak_core_stat_cache.erl
```diff
@@ -0,0 +1,137 @@
...
+                {true, {M, F, A, TTL}} ->
+                    case cache_get(App, Tab, TTL) of
+                        No when No == miss; No == stale ->
+                            Active = maybe_get_stats(App, From, Active0, {M, F, A}),
+                            {noreply, State0#state{active=Active}};
+                        Hit ->
+                            {reply, Hit, State0}
```
beerriot: I wonder if exposing the staleness of the Hit would be of any benefit here.  That is, would debugging be helped by returning `{ok, Hit, Timestamp}` instead of just `Hit`? Or would the timestamp be to likely to mislead on a heavily-loaded system?

src/riak_core_stat_cache.erl
```diff
@@ -0,0 +1,137 @@
...
+                    end
+            end,
+    Reply;
+handle_call(_Request, _From, State) ->
+    Reply = ok,
+    {reply, Reply, State}.
+
+handle_cast({stats, App, Stats}, State0=#state{tab=Tab, active=Active}) ->
+    ets:insert(Tab, {App, folsom_utils:now_epoch(), Stats}),
```
beerriot: This is calculating the timestamp after it has been delivered to the cache. If the cache has been backlogged by lots of requests, this means the staleness calculation will be later than if the timestamp had been calculated as soon as folsom returned in `do_get_stats/2`. Maybe that's okay, but it seemed worth mentioning.

beerriot:
I have not performance-tested this PR and the kv/search/repl PRs that go along with it, but functional testing for all (doing gets/puts, searches, mdc, then grabbing stats) works as expected (seeing two requests less than 5sec apart return the same response, but requests farther apart return different values).

The code makes sense to me, despite my comments for improvement.

If it has, indeed, already passed sufficient performance tests, I see no reason to hold up its merge. (soft +1)

russelldb:
Thanks for the review and comments. I'm going to act on them all and re-run perf tests and post numbers here. Hopefully you'll be able to re-review at that time.

src/riak_core_stat_cache.erl
russelldb: Sort of an in joke. Not the place for it. Have changed as you recommend.

russelldb:
https://www.dropbox.com/sh/2tk139929rymla6/aG3MT7xiJe/mc contains all the basho_bench output from a series of tests of 1.1.2 and various 1.2 stats code. 

https://www.dropbox.com/sh/2tk139929rymla6/g2Rx9ao6o7/mc/1.2_stats_cached_read_write_9concurrent_gen_server_bc
and
https://www.dropbox.com/sh/2tk139929rymla6/uRmxkpow5o/mc/1.2_stats_cached_read_write_9concurrent_gen_server_lev

Show the perf of the latest commits (using a gen_server rather than all those spawns) The `get_stats.png` shows the throughput of 9 concurrent workers calling the REST `stats` endpoint. `Summary.ng` shows the throughput of riak. The config files show the basho_bench config.

I think the comparison between 1.1.2 and 1.2 are compelling,

Ive updated all repos to use a gen_server rather than the spawns.

russelldb:
Pushed commits that answer your comments. (Will squash down for merge.)

beerriot:
"All 92 tests passes." Hooray!

russelldb:
Taking that as a +1 and merging (when you plus one the others)

beerriot:
Yes, indeed. +1 merge. Sorry for the confusion.

russelldb:
Squashed and merged. Many thanks for the proper review!

Issue #190, opened 2012-06-13
Guard against reading corrupt/empty ring files


ghost:
Right now, we currently attempt to read the latest ring file, regardless if it is corrupt, empty, et cetera. This has been seen several times, and typically we delete the most recent ring file, and use the previous one. 

Riak itself could attempt to implement this functionality. 

See also: https://issues.basho.com/show_bug.cgi?id=1078

dizzyd:
There are a few questions that arise here:
1. Do we use the last-known-good ring file if the latest is corrupt? This might lead to strange issues if the LKG was very old or in a previous configuration.
2. Do we delete corrupt ring files? We can't parse them, so they're basically useless.

I do agree we can do better. As a interim fix, we could update riaknostic to check for this problem so that it's at least easier to diagnose.

Techmind:
I think 1).

Actually this can be repeatable problem if disk became full unexpectedly: 

1)you can`t write data, but fs seems to have allocated some free space in advance so it can create inode and edit directory to have new file.
2) you move/delete some files
3) restart riak, riak crashes because of corrupted ring file
4) you delete ring files
5) compaction or something else occurs - disk is full again =)
6) empty ring files are there again, riak crashes, you delete ring files, etc.
7) repeat X times =)

very entertaining)

jonmeredith:
The pull request above should help the out of disk space or crash during writing case.  It won't fix filesystem corruption, but then you're probably dealing with a whole world of hurt.

slfritchie:
Resolved via https://github.com/basho/riak_core/pull/195

PR #195, opened 2012-06-22
Change write_ringfile to create a temporary ring file, check and rename.

jonmeredith:
Mitigation for #190 - check the written ringfile before making it usable.

### DIFF
--- a/src/riak_core_ring_manager.erl
+++ b/src/riak_core_ring_manager.erl
@@ -115,17 +115,30 @@ force_update() ->
     ok.
 
 do_write_ringfile(Ring) ->
-    {{Year, Month, Day},{Hour, Minute, Second}} = calendar:universal_time(),
-    TS = io_lib:format(".~B~2.10.0B~2.10.0B~2.10.0B~2.10.0B~2.10.0B",
-                       [Year, Month, Day, Hour, Minute, Second]),
     case app_helper:get_env(riak_core, ring_state_dir) of
         "<nostore>" -> nop;
         Dir ->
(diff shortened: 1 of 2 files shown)


slfritchie:
Jon, how paranoid do you want to be?  IIRC, the most paranoid wisdom is:
1. write .tmp file data
2. fsync(2) the .tmp file
3. rename(2) the .tmp file
4. fsync(2) the parent directory  (not 100% certain on if timing of opening the parent dir's fd is important)

If we don't care 100% about the renaming operation happening, then omitting step 4 would be ok.

jonmeredith:
I can't see a way to persuade file:write_file to take an O_SYNC but that's probably a linuxism anyway.  I was mostly trying to guard against out of space issues, but it's worth doing right.

I'll switch to an open/write/fsync/close.

Any suggestions on how to do step 4? file:sync/datasync take an IoDevice and

```
Eshell V5.9.1  (abort with ^G)
1> file:open(".",[read]).
{error,eisdir}

```

slfritchie:
Given that ring updates aren't really coordinated, and a race with a system crash would still leave the old ring file in place, I think number 4 is optional, especially since the Erlang file driver is picky about dealing with regular files {sigh}.

slfritchie:
@jonmeredith I've added commit cc6edac.  My intent is to use that same func (cut-and-paste, alas) for Bitcask's lock file creation, since we need the same thing there.  Whaddya think?

jonmeredith:
Tests good for me  +1 merge.

PR #201, opened 2012-06-27
Remove redundant stat mode slide and spiraltime

russelldb:


seancribbs:
+1, tests pass now. Merging

Issue #197, opened 2012-06-23
Ability to delete custom bucket properties


ghost:
Right now, while we have the ability in `riak_core_bucket` to set bucket properties:

``` erlang
%% @spec set_bucket(riak_object:bucket(), BucketProps::riak_core_bucketprops()) -> ok
%% @doc Set the given BucketProps in Bucket.
set_bucket(Name, BucketProps0) ->
    case validate_props(BucketProps0, riak_core:bucket_validators(), []) of
        {ok, BucketProps} ->
            F = fun(Ring, _Args) ->
                        OldBucket = get_bucket(Name),
                        NewBucket = merge_props(BucketProps, OldBucket),
                        {new_ring, riak_core_ring:update_meta({bucket,Name},
                                                              NewBucket,
                                                              Ring)}
                end,
            {ok, _NewRing} = riak_core_ring_manager:ring_trans(F, undefined),
            ok;
        {error, Details} ->
            lager:error("Bucket validation failed ~p~n", [Details]),
            {error, Details}
    end.
```

We do not have the ability to delete custom bucket properties. Since these are stored in the ring, this adds up when creating hundreds/thousands of buckets with custom properties. A use case that has been seen is where certain buckets require the search post-commit hook, and others do not.

Right now, even if you delete all the keys and attempt to set the bucket properties back to default, it still will be kept in the ring.

It should be feasible to delete the dict key for a certain bucket's properties, and have it revert back to default properties.

jerith:
The slowdown in setting bucket properties when there is already a lot of metadata is quite extreme. See https://gist.github.com/2978270 for a script that demonstrates this.

slfritchie:
All constituent tickets reviewed, +1 to merge, thanks!
- riak_core #197
- riak_core #202
- riak_core #248
- riak_test #81
- riak_kv   #422

PR #202, opened 2012-06-28
Add functions for clearing bucket properties.

jerith:
This is an implementation of the `riak_core` functionality required for #197. Additional work is required in `riak_kv` to expose this to the outside world.


src/riak_core_bucket.erl
```diff
@@ -84,6 +85,17 @@ merge_props(Overriding, Other) ->
     lists:ukeymerge(1, lists:ukeysort(1, Overriding),
                     lists:ukeysort(1, Other)).
 
+
+%% @spec clear_bucket(riak_object:bucket()) -> ok
```
rzezeski, line 89: riak_object is not visible in riak_core.  I think `binary()` is good enough if there isn't already a bucket type in core.  Also, do you mind using module attributed specs?  We are trying to move away from edoc specs.

jerith:
This branch really needs to be updated to match the state of the repo. I'll probably have time to do that next week when I'm finished with all my traveling.

src/riak_core_bucket.erl
jerith, line 89: When I merge in all the 1.2 work, I'll update my changes to match whatever the surrounding code uses. Would that be sufficient?

rzezeski:
Sounds good, safe travels.

jerith:
No merge conflicts and everything still works fine. Same for basho/riak_kv#357, except the tests fail there because they're being run against the wrong `riak_core` branch.

@rzezeski: Everything else in that file still uses edoc specs and I ran into various problems switching to compiler specs. I don't know enough about specs to fix them properly, so I'd rather leave it for someone more experienced.

src/riak_core_ring.erl
```diff
@@ -500,6 +501,24 @@ update_meta(Key, Val, State) ->
             State
     end.
 
+% @doc Clear a key in the cluster metadata dict
```
rzezeski, line 504: Why not just do it all in one case to cut done on number of lines:

``` erlang
case dict:find(...) of
    {ok, _} ->
        VClock = ...,
        State?CH....
    error ->
        State
end
```

src/riak_core_bucket.erl
rzezeski, line 89: I still see `riak_object` in the spec.  It may be used in other places in core but that is a mistake.  Please change this to `binary()`.

src/riak_core_ring.erl
rzezeski, line 504: Actually, this isn't going to work because it doesn't play nice with metadata reconciliation.  The short story is that you can't just remove the meta entry.  Depending on the state of cluster ring reconciliation might make it reappear.  I think you can achieve your goal with `riak_core_ring:update_meta/3` and using an empty list as value but I would have to run through the code to verify.

jerith, line 504: Cool, thanks for the feedback.

rzezeski:
@jerith I'm terribly sorry I let this linger for so long.  I have
verified today that just removing the meta would be a bad idea.  The
[merge_meta](https://github.com/basho/riak_core/blob/1.2.1/src/riak_core_ring.erl#L942) function is a [dict:merge](http://www.erlang.org/doc/man/dict.html#merge-3) which has the
following contract:

> All the Key - Value pairs from both dictionaries are included in the
> new dictionary.

Thus, without a tombstone value, the old properties would become
resurrected on the next ring reconcile.  You can verify this yourself
by adding the following lines to the [metadata_inequality_test](https://github.com/basho/riak_core/blob/1.2.1/src/riak_core_ring.erl#L1330).

``` erlang
    Ring3 = clear_meta(key,Ring2),
    ?assertEqual(undefined, get_meta(key, Ring3)),
    ?assertEqual(undefined, get_meta(key, ?CHSTATE{meta=merge_meta(Ring2?CHSTATE.meta, Ring3?CHSTATE.meta)})).
```

A solution that would work is to write a special _tombstone_ value
instead, e.g. `'$tombstone$'`.  Then change functions like
[get_bucket](https://github.com/basho/riak_core/blob/1.2.1/src/riak_core_bucket.erl#L105) to be tombstone aware and return default bucket props
in that case.

Does this make sense?  Would you like to try updating your branch to
use the method I suggest?  If not I could take a stab at it myself.
Let me know.

jerith:
No worries, I've been rather busy myself. We've made so changes to our test setup that mitigate the problem this causes, so the situation isn't as bad as it was, but I'd still like to get this resolved.

As far as I can tell, using tombstone values would look like deletion from the outside (which is fine), but wouldn't solve the problem of unbounded metadata growth which leads to dramatic performance hits when operating on bucket properties. We'd have to also add a mechanism for removing tombstoned properties from the metadata after reconciliation has finished. Is this something that's feasible to implement?

rzezeski:
I believe a similar mechanism to how nodes are removed could be used but I have a feeling it would be a lot of work.  I need to consult with @jtuple to be sure.  Currently the ring is a dump for many different types of cluster-wide data.  We want to add a system that allows to share facts like bucket properties without using the ring.  But I'm not sure when work on that might start.

Adding tombstone support would help alleviate growing metadata.  It's not perfect, but is an improvement.  An atom is much less overhead than a dict.  It could act as a decent stop-gap until we implement something better.  If you'd like to modify this branch to implement tombstones I will do my best to push it through.  If not, I might just start my own branch.  What do you think?

jrwest:
@jerith @rzezeski I'm planning to open up a new PR that has an implementation using the tombstone method. going to close this one out. Ill attach my PR to the original issue: https://github.com/basho/riak_core/issues/197

PR #203, opened 2012-06-28
Fix handoff_ip bug in mixed version clusters

jtuple:
This pull-request fixes the handoff ip bug identified by @rzezeski when investigating failures in the rolling upgrade basho_expect test. The change replaces the use of a `gen_server:call` with an `rpc:call` to prevent older nodes from crashes because they do not expect the `handoff_ip` message.

This pull-request also addresses another issue with cluster transitions stalling due to either missing ring change events, or due to a race between cluster capabilities negotiation and claimant transitions. This bug was hit when working on an automated test to trigger the original handoff ip bug that was being fixed.

I've also added a new `verify_basic_upgrade` to `riak_test` that verifies correct behavior after these changes (failing without). We will need to re-build packages and verify against the basho_expect rolling upgrade test to see if there are any other issues.

### DIFF
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -160,6 +160,7 @@ claimant() ->
 %%%===================================================================
 
 init([]) ->
+    schedule_tick(),
     {ok, #state{changes=[], seed=erlang:now()}}.
 
 handle_call(clear, _From, State) ->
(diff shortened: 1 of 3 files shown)


Vagabond:
+1 with 2 caveats: the use of the timer module and the fact that you had to force a ring update might be more appropriate as a warning.

Vagabond:
Also, I forgot to actually run this code, or the riak_test, it was purely a visual review.

jtuple:
Force pushed new code based on your feedback. Going to merge.

PR #204, opened 2012-06-29
Unpin the protobuffs dependency.

seancribbs:
When using the tag, this prevents riak_pb from using the correct dependency version, breaking lots of functionality. riak_core should be using the same version as riak_pb.


jaredmorrow:
+1

PR #205, opened 2012-07-02
Use a sliding one minute window not a EWMA meter

russelldb:
Depends on https://github.com/boundary/folsom/pull/27…

Use a one minute count and total count, not a weighted moving average.


rzezeski:
I believe `rebar.config` needs to be updated to pull the Basho fork of folsom now.  No?

russelldb:
Yes :$

russelldb:
Well, I was hoping Joe W. would merge upstream…

rzezeski:
I don't think we have the convenience of waiting for that to happen.

russelldb:
updated rebar.config

rzezeski:
+1 to merge

PR #206, opened 2012-07-05
Fix bug that caused ring to be forcibly updated continuously

jtuple:
Change riak_core_claimant to only force a ring update of a stalled ring if the ring is "ready" and has therefore already converged across the cluster. Without this change, ring convergence always appears to be stalled if a node is offline and therefore the force update happens over and over.

### DIFF
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -470,8 +470,10 @@ tick(State) ->
     State.
 
 maybe_force_ring_update() ->
-    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
-    case riak_core_ring:claimant(Ring) == node() of
+    {ok, Ring} = riak_core_ring_manager:get_raw_ring(),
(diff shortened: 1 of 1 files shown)


jtuple:
If you want to test this out, here's the general approach.

To observe the broken behavior:
1. Take master and build a 3-node dev release.
2. Join `dev2` to `dev1`: `dev/dev2/bin/riak-admin join -f dev1@127.0.0.1`
3. Stop `dev2`: `dev/dev2/bin/riak stop`
4. Join `dev3` to `dev1`: `dev/dev3/bin/riak-admin join -f dev1@127.0.0.1`
5. Since `dev2` is down, the joining of `dev3` to the cluster stalls. This is expected, and `riak-admin ring_status` will warn you of this behavior. What is not desired, is for the message "Forcing update of stalled ring" to be printed every 10 seconds or so on the console/console.log of `dev1`. But, this is what you'll observe.

Try the above again with this branch, and you won't see the message printed over and over.

chardan:
+1

PR #209, opened 2012-07-17
Don't crash the cache when a stat mod crashes

russelldb:
Since calls to calculate stats are spawn_linked to ensure
that the calculating processes exits if the cache exits, the cache
will crash if a spawned process exits. This is bad. This change
traps_exits on spawned processes and enforms awaiting callers of
the error, without crashing the cache or effecting other registered
mods.

PR #210, opened 2012-07-17
Don't crash the cache when a stat mod crashes

russelldb:
See #209 for lengthy discussion

---

Since calls to calculate stats are spawn_linked to ensure
that the calculating processes exits if the cache exits, the cache
will crash if a spawned process exits. This is bad. This change
traps_exits on spawned processes and enforms awaiting callers of
the error, without crashing the cache or effecting other registered
mods.


russelldb:
Merging since #209 is +1'd

PR #211, opened 2012-07-19
Register stat mods with the riak_core app

russelldb:
When the stat cache crashes, we must re-register stat mods with
the cache so that it works when re-started.


jonmeredith:
A couple of accidental files included, but once removed, +1.

```
src/#riak_core_stat_cache.erl#
src/.#riak_core_stat_cache.erl
```

russelldb:
Thanks Jon.

PR #212, opened 2012-07-20
Restart vnode worker pool in case of crash.

kellymclaughlin:
If a vnode's worker pool crashes it is not restarted and this can lead to the vnode crashing if it tries to use the pool pid since it is now just the atom undefined. Change the vnode to restart the worker pool in case of a crash.

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -516,6 +516,9 @@ handle_info({'EXIT', Pid, Reason},
             continue(State#state{pool_pid=undefined});
         _ ->
             lager:error("~p ~p worker pool crashed ~p\n", [Index, Mod, Reason]),
+            lager:debug("starting worker pool ~p with size"
+                        "of ~p for vnode ~p.",
+                        [WorkerModule, PoolSize, Index]),
             {pool, WorkerModule, PoolSize, WorkerArgs}=PoolConfig,
             {ok, NewPoolPid} =
                 riak_core_vnode_worker_pool:start_link(WorkerModule,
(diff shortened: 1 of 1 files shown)


src/riak_core_vnode.erl
```diff
@@ -501,15 +505,26 @@ handle_sync_event(core_status, _From, StateName, State=#state{index=Index,
...
-            ok;
+            continue(State#state{pool_pid=undefined});
         _ ->
-            lager:error("~p ~p worker pool crashed ~p\n", [Index, Mod, Reason])
-    end,
-    continue(State#state{pool_pid=undefined});
+            lager:error("~p ~p worker pool crashed ~p\n", [Index, Mod, Reason]),
+            {pool, WorkerModule, PoolSize, WorkerArgs}=PoolConfig,
+            {ok, NewPoolPid} =
+                riak_core_vnode_worker_pool:start_link(WorkerModule,
```
Vagabond, line 534: A log message about restarting the pool might be nice, even if it was at the debug level, like when a pool is originally started.

Vagabond:
Confirmed working, +1 to merge.

reiddraper:
Merge it now that we're post 1.2?

kellymclaughlin:
This test is a little noisy due to the process death. I could try and fool around with the logging nonsense to shut it up if it's deemed egregious. I also considered breaking out the setup into an eunit suite, but since there is only one test case right now it seemed pointless.

reiddraper:
nice test, +1 to merge

PR #214, opened 2012-07-22
Make the ring manager responsible for loading the ring

jtuple:
Change riak_core_ring_manager and riak_core_app so that the ring manager
is responsible for loading the ring file from the disk rather than starting
with an initially empty ring and then relying upon the riak_core app to
later load the ring. This avoids a race condition with the ring manager
writing the empty ring to the disk before the riak_core app loads the
prior ring.

Note: Riak previously relied upon starting with a fresh ring in order to
ensure secondary vnodes were started in case any had fallback data that
needed to be handed off. The act of starting secondaries has long since
been moved to the riak_core_vnode_manager that periodically starts up
secondary vnodes over time, therefore there is no longer any need to
start with a fresh ring. This commit will therefore always load a saved
ring when the ring_manager starts, rather than starting with a fresh ring.


slfritchie:
Re-review: +1 to merge, thanks!

PR #216, opened 2012-07-24
Fix capability system race condition

jtuple:
Prevent the node from crashing when the capability system attempts to
negotiate capabilities before the node has registered any capabilities.


Vagabond:
Seems reasonable. +1 on visual review.

PR #217, opened 2012-07-25
restructure supervision tree so that folsom is an included app

russelldb:
All stat mods depend on folsom, yet they are not linked to it.
This change brings folsom under supervision of a core stat sup,
which also supervises the riak stat subsystem. Now when folsom exits
everyone gets to restart clean and recover.

```
riak_core_sup
   |
   riak_core_stat_sup (rest_for_one)
    \
       - folsom_sup
       - riak_core_stats_sup (one_for_one)
        \
          - riak_*_stat
          - riak_stat_cache
```

riak_core_stats_sup will start and supervise gen_server stat mods at
registration time, and will re-start them should the sup crash.

### DIFF
--- a/ebin/riak_core.app
+++ b/ebin/riak_core.app
@@ -55,6 +55,8 @@
              riak_core_ring_util,
              riak_core_stat,
              riak_core_stat_cache,
+             riak_core_stat_sup,
+             riak_core_stats_sup,
              riak_core_status,
              riak_core_sup,
              riak_core_sysmon_handler,
@@ -76,6 +78,7 @@
              vclock
             ]},
   {registered, []},
+  {included_applications, [folsom]},
   {applications, [
                   kernel,
                   stdlib,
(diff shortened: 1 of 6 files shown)


argv0:
+1 to merge.  Tested various crash scenarios, does what it says on the tin.

yrashk:
I am wondering if this was a good decision. Systems that use riak_core would also like to use folsom (this is our case, too) — but we can't nicely initialize our metrics for these reasons:
- folsom_sup doesn't read out folsom application env to initialize metrics declared there
- with riak_core being an included application (at least in our case) we can't initialize our metrics _before_ we start our supervisor because in this case riak_core startup will fail — because we would have to start folsom to initialize our metrics, and riak_core_stat_sup will obviously choke on that
- also, we can't initialize our metrics _after_ we start our own supervision tree because some processes in that supervision tree need the metrics to be initialized right away, at their startup time

I only have two workarounds, and neither of them is pretty:
- Start a temporary (or a permanent) worker right after riak_core to let it declare the metrics. It can also find folsom_sup and link to it, I guess
- Initialize required metrics in their respective workers — which makes metrics initialization really scattered around and in certain case, leads to runtime initialization duplication overhead.

Any thoughts?

PR #218, opened 2012-07-25
Fix spurious "Forcing update of stalled ring"

jtuple:
Changed the force update logic in riak_core_claimant to not perform
a forced update if we have pending staged joins and no auto-joining
nodes. Forcing a ring update because of staged joins will not actually
change the ring, because staged joins will not transition until
committed. This was a false positive detection of a stalled ring.


jonmeredith:
+1 don't see the message any more and the join runs nicely.

PR #220, opened 2012-08-16
update rebar to 2.0.0

bookshelfdave:


seancribbs:
A few tests fail for me but are unrelated to the rebar change. :+1:

PR #223, opened 2012-09-08
Change ticks from timer to more efficient erlang:send_after

jtuple:
Title says it all: replace certain uses of `timer` with `erlang:send_after`. Thanks go to @Vagabond for mentioning that `erlang:send_after` uses a timer-wheel built into erts and is more efficient than the user-space `timer` module.


Vagabond:
+1 to merge. Verified that calls to erlang:send_after appear regularly when using dbg tracing.

I really think we should use self() rather than ?MODULE for the destination though, going through the process dictionary has 0 benefit here, as far as I can tell.

PR #224, opened 2012-09-08
erlang:now() -> os:timestamp() in all the places it is safe

Vagabond:
There are a few places I didn't touch as it was unclear if the values
needed to be monotonic or not. Specifically core_claimant, core_gossip,
core_ring and core_ring_manager.


jonmeredith:
+1 merge

PR #225, opened 2012-09-10
remove -author attributes from source

argv0:


beerriot:
+1 when do the rest of the repos get this treatment?

PR #226, opened 2012-09-10
cluster_info removal

evanmcc:
Cluster info has been folded into riaknostic, so rip it out here.


evanmcc:
doesn't look like this is going to happen, closing.,

PR #229, opened 2012-09-18
Remove extra erl_opts

DeadZen:
Remove duplicate erl_opts, keep debug_info

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,8 +1,7 @@
 {erl_first_files, ["src/gen_nb_server.erl", "src/riak_core_gen_server.erl"]}.
 {cover_enabled, true}.
-{erl_opts, [{parse_transform, lager_transform}]}.
+{erl_opts, [{parse_transform, lager_transform}, debug_info]}.
 {edoc_opts, [{preprocess, true}]}.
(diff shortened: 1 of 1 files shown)


seancribbs:
Should probably add fail_on_warning in there to replace warnings_as_errors.

DeadZen:
I thought that, but decided since riak_core is meant to be included in other distributed applications; it might be dev overkill.

seancribbs:
@DeadZen Oh, I just meant that the option name has changed, was deprecated in R14. Unless I don't remember correctly?

DeadZen:
Gotcha, either way, I think it is best to leave it out of this rebar.config, for a few reasons:
It may prematurely impede development with riak_core and it can always be added to the embedding application.
Additionally the extra erl_opts, prevented it along with debug_info from being applied previously anyway.
I believe removing the additional erl_opts and keeping just the debug_info, is the path of least astonishment.

rzezeski:
@seancribbs It's still `warning_as_errors`.  I'm going to accept the patch as-is.  It can always be added against later if need be.

@DeadZen  Looks good to me.  I'm merging.

PR #230, opened 2012-09-21
Remove publish_capabilities race

reiddraper:
See commit message for more detail.

To exhibit the bug, I ran a fresh single-node cluster with the following diff:

``` patch

diff --git a/src/riak_core_capability.erl b/src/riak_core_capability.erl
index a3a4fe6..e6e371c 100644
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -378,6 +378,7 @@ publish_supported(State) ->
                 end
         end,
     spawn(fun() ->
+                  timer:sleep(crypto:rand_uniform(1, 1000)),
                   riak_core_ring_manager:ring_trans(F, ok)
           end),
     ok.
```

After running this a few times, I observed missing capabilities in the ring, as seen by
`riak_core_ring_manager:get_my_ring()`.

All credit goes to @jonmeredith for being suspicious of this call in a spawn.

### DIFF
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -377,9 +377,7 @@ publish_supported(State) ->
                         ignore
                 end
         end,
-    spawn(fun() ->
-                  riak_core_ring_manager:ring_trans(F, ok)
-          end),
+    riak_core_ring_manager:ring_trans(F, ok),
     ok.
 
 %% Add a node's capabilities to the provided ring


jtuple:
So, this looks reasonable and makes sense as to how it fixes the race. I do wonder if this was put in originally to avoid some rare deadlock between the ring manager and the capability manager (ie. they both call and wait on each other), but I can't seem to reproduce anything bad like that with manual testing.

I would love to have checked this against the `verify_capabilities` and `rolling_capabilities` tests in `riak_test`, but those tests currently fail on master. Have some ideas on how to fix those, but don't want to hold up getting this merged in and another 1.2.1 RC build made for testing. We can just focus on fixing capabilities test afterwards and make passing capabilities test one of the release criteria.

+1 merge

PR #231, opened 2012-09-26
Improve exception handling in riak_core_vnode_proxy

kellymclaughlin:
The `call` functions in `riak_core_vnode_proxy` only expected a result of `{ok, Res}`. This change calls a new function, `call_reply`, to better handle the case when an exception is thrown and wrap the reason in an error tuple.


beerriot:
+1 code looks good to me, and tests pass locally

PR #232, opened 2012-10-01
Address high memory use by riak_core_sysmon_handler

kellymclaughlin:
Change `riak_core_sysmon_handler` to use hibernation to free up resources when it is idle since it does not do a good job of freeing these resources on its own. Also force garbage collection on the `riak_core_sysmon_handler` process if it receives a `large_heap` message about itself. This is to avoid a feedback loop that can lead to severe memory usage by the handler, node slowness, and even cause the node to crash after consuming all available memory.

Ryan mentioned [this](http://www.lshift.net/blog/2010/02/28/memory-matters-even-in-erlang) blog post by LShift about some troubles rabbitmq had with process hibernation. I do not feel that it should be a concern here because there are no resources used by this module that are referenced between hibernation periods (_i.e._ It just processes transient events), but it would be good for someone else to consider this as well.

### DIFF
--- a/src/riak_core_sysmon_handler.erl
+++ b/src/riak_core_sysmon_handler.erl
@@ -87,7 +87,6 @@ handle_event({monitor, Pid, Type, _Info},
     %% Reset the inactivity timeout
     NewTimerRef = reset_timer(TimerRef),
     maybe_collect_garbage(Type),
-    erlang:garbage_collect(),
     {ok, State#state{timer_ref=NewTimerRef}};
 handle_event({monitor, Pid, Type, Info}, State=#state{timer_ref=TimerRef}) ->
     %% Reset the inactivity timeout
--- a/src/riak_core_sysmon_handler.erl
+++ b/src/riak_core_sysmon_handler.erl
@@ -31,10 +31,12 @@
          get_pretty_proc_info/1, get_pretty_proc_info/2]).
 
 %% gen_event callbacks
--export([init/1, handle_event/2, handle_call/2, 
+-export([init/1, handle_event/2, handle_call/2,
          handle_info/2, terminate/2, code_change/3]).
 
--record(state, {}).
+-record(state, {timer_ref :: reference()}).
+
+-define(INACTIVITY_TIMEOUT, 5000).
 
 %%%===================================================================
 %%% gen_event callbacks
@@ -65,7 +67,7 @@ add_handler() ->
 %% @end
 %%--------------------------------------------------------------------
 init([]) ->
-    {ok, #state{}}.
+    {ok, #state{}, hibernate}.
 
 %%--------------------------------------------------------------------
 %% @private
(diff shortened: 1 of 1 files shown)


src/riak_core_sysmon_handler.erl
```diff
@@ -80,14 +82,17 @@ init([]) ->
 %%                          remove_handler
 %% @end
 %%--------------------------------------------------------------------
-handle_event({monitor, Pid, Type, Info}, State) ->
+handle_event({monitor, Pid, Type, Info}, State=#state{timer_ref=TimerRef}) ->
+    %% Reset the inactivity timeout
+    NewTimerRef = reset_timer(TimerRef),
+    maybe_collect_garbage(Pid, Type),
```
Vagabond: What about suppressing the sysmon message entirely if it is for the sysmon process?

src/riak_core_sysmon_handler.erl
```diff
@@ -122,6 +127,10 @@ handle_call(_Call, State) ->
 handle_info(die_for_testing_purposes_only, _State) ->
     %% exit({told_to_die, lists:duplicate(500000, $x)});
     exit({told_to_die, lists:duplicate(50, $x)});
+handle_info(inactivity_timeout, _State) ->
```
Vagabond: Underscore variables are sort of misleading here.

src/riak_core_sysmon_handler.erl
```diff
@@ -192,3 +201,21 @@ get_pretty_proc_info(Pid, Acf) ->
...
             RNL ++ [ICT, {Acf, CF}, {message_queue_len, MQL}]
     end.
+
+
+%% @doc If the message type is due to a large heap warning
+%% and the source is ourself, go ahead and collect garbage
+%% to avoid the death spiral.
+-spec maybe_collect_garbage(pid(), atom()) -> ok.
+maybe_collect_garbage(Pid, large_heap) when Pid =:= self() ->
+    erlang:garbage_collect(Pid),
```
Vagabond: garbage_collect/0 should be fine here.

src/riak_core_sysmon_handler.erl
Vagabond: You could even have a new clause of handle_event like this:

``` erlang
handle_event({monitor, Pid, large_heap, _Info}, State)  when Pid == self() ->
    erlang:garbage_collect(),
    {ok, State}
```

Vagabond:
Do you have a good way to provoke the sysmon death spiral in question so I can compare the behavior before/after this change?

src/riak_core_sysmon_handler.erl
kellymclaughlin: I do like the idea of suppressing messages about the sysmon itself. I wrote it that way at first, but then went back and changed it, but if you agree, I'll reinstate that behavior.

kellymclaughlin:
The only way I've been able to really observe this is on the BOS cluster with a 42 node cluster and many millions of objects in a bucket. But I also haven't really done much time trying to reproduce it locally or on a vm setup so it may be possible.

src/riak_core_sysmon_handler.erl
Vagabond: I think it is prudent, we really want to avoid death spiraling, and printing messages about yourself only seems like it'd add to the problem.

Vagabond:
Why did you choose a 5 second timeout? Other than that everything looks good. I'd like to see the impact in BOS before +1ing it, but I won't insist on it.

kellymclaughlin:
5 seconds was arbitrary. Seemed like enough of time window that if nothing happened then it may be ok to hibernate, but not too long.

Vagabond:
+1 after observing improvements in BOS.

jaredmorrow:
Merging after +1 from @kellymclaughlin and @Vagabond in chat

PR #233, opened 2012-10-17
Add Health callback option to node watcher

lordnull:
Add a second way to monitor a service, allowing for a callback to be used to determine health, and automatically set a service up or down.


lordnull:
Meant to make the target 1.2, not master.

PR #242, opened 2012-10-17
Update RINGTOP calculation

D6:
RINGTOP is calcuated incorrectly as 

``` erlang
trunc(math:pow(2,160)-1)
```

because substracting one from such a large float before the trunc call has no
effect. Subtracting one after the trunc call results in the correct
value.

``` erlang
trunc(math:pow(2,160))-1
```

### DIFF
--- a/src/chash.erl
+++ b/src/chash.erl
@@ -52,7 +52,7 @@
 
 -export_type([chash/0]).
     
--define(RINGTOP, trunc(math:pow(2,160))-1).  % SHA-1 space
+-define(RINGTOP, trunc(math:pow(2,160))).  % SHA-1 space
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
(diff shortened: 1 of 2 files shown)


D6:
The Travis build failure was caused by the unit test for the chash size. Should I update the tests also? 

Apologies in advance if that is the dumbest question you've seen in a while. I'm new to open source collaboration

slfritchie:
Time for some belated repo curating ... @jtuple @jonmeredith Thoughts?

jrwest:
Last I discussed this offline a while back we decided that since its old working code, although slightly ambiguous, it can be left alone. Newer code like chashbin has [done its own calculation](https://github.com/basho/riak_core/blob/develop/src/chashbin.erl#L227) -- which is necessary, anyways, since the `-define` isn't available outside the module. If we are going to change it I'd vote for the `bsl` approach.

slfritchie:
Hi, based on Jordan's comment, I'm going to close this ticket.  Many thanks for the pull request, and many apologies for the huge delay before finally making a decision!

Issue #243, opened 2012-10-18
Notify user of stale rings


evanmcc:
When a node starts, if there is a stale or mismatching ring file on the machine, the node dies with an uninformative error.  This can happen when attempting to change the nodenames of a cluster, or when an install has brought up an unconfigured node, as in: https://github.com/basho/node_package/issues/26

Example error:

19:54:20.781 [info] Application lager started on node 'riak-new1@192.168.0.10'^M
19:54:20.864 [info] Upgrading legacy ring^M
19:54:21.015 [error] gen_server riak_core_capability terminated with reason: no function clause matching orddict:fetch('riak-new1@XXX.XXX.XXX.XXX', [{'riak1@YYY.YYY.YYY.YYY',[{{riak_core,staged_joins},[true,false]},{{riak_core,vnode_routing},[proxy,...]},...]},...]) line 72^M
/usr/lib64/riak/lib/os_mon-2.2.9/priv/bin/memsup: Erlang has closed.
Erlang has closed

It'd be nice if we caught this and exited with a sensible suggestion.  E.G:

"Stale ring file found, you may wish to remove it or move it aside."

slfritchie:
Time for some belated repo curating ... +1, yay.

jrwest:
we should totally still fix this. seems like a good easy fix for a `2.0.1`. marking as such. may slip.

bearcage:
:+1: for fixing this one, continues to crop up periodically in support issues as well.

Looks like there's an attempted fix in place at https://github.com/basho/riak_core/blob/1.4.9/src/riak_core_capability.erl#L413-L415 that doesn't seem to be matching the call for some reason.

EDIT: relevant support ticket: #[8264](https://basho.zendesk.com/agent/#/tickets/8264)

PR #244, opened 2012-10-18
Clarify error message for stale ring failures.

evanmcc:
Add a meaningful warning to the failure when the ring file
doesn't have a entry for the current node name.

Fixes issue #243

### DIFF
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -412,7 +412,11 @@ negotiate_capabilities(Node, Override, State=#state{registered=Registered,
 renegotiate_capabilities(State=#state{supported=[]}) ->
     State;
 renegotiate_capabilities(State) ->
-    Caps = orddict:fetch(node(), State#state.supported),
+    Caps = 
+        case orddict:find(node(), State#state.supported) of 
+            {ok, Val} -> Val;
+            error -> error("Node name mismatch, move or remove stale ring file")                       
+        end,
     Overrides = get_overrides(Caps),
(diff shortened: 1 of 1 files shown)


src/riak_core_capability.erl
```diff
@@ -412,7 +412,11 @@ negotiate_capabilities(Node, Override, State=#state{registered=Registered,
 renegotiate_capabilities(State=#state{supported=[]}) ->
     State;
 renegotiate_capabilities(State) ->
-    Caps = orddict:fetch(node(), State#state.supported),
+    Caps = 
+        case orddict:find(node(), State#state.supported) of 
+            {ok, Val} -> Val;
+            error -> error("Node name mismatch, move or remove stale ring file")
```
jrwest, line 418: is it safe to just move the stale ring file aside? In the case where I forgot to rename a node and a cluster has been resized wont moving the file cause the node to startup possibly with the old ring size meaning it will have problems when it receives a gossiped ring from another member?

evanmcc, line 418: We often direct people to use older ring files when theirs are corrupted.
If this is unsafe under a resize we may have some more thinking to do.  In
any case, this is usually something that people see when they're just
starting up, rather than on mature clusters.

jrwest, line 418: ah yes, older ring files, that addresses my concern.

slfritchie:
@evanmcc Based on the comments that I see here and on the referenced tickets, there's enough +1 to +1.  +1 from me also.

PR #245, opened 2012-10-26
gen_nb_server: Support IPv6 addresses for handoff_ip.

### LINKED ISSUE #167: Riak can't listen on ipv6 addresses
I'm trying to deploy a riak cluster on machines with no IPv4 addresses.

I seem to be able to configure the http setting to listen on {"::", 8098}, but handoff_ip doesn't seem to be able to be given an IPv6 address.

When I configure:

```
{handoff_port, 8099 },
{handoff_ip, "::" },
```

riak won't start and I get the following error in `riak console`:

```
Attempting to restart script through sudo -u riak
Exec: /usr/lib64/riak/erts-5.8.5/bin/erlexec -boot /usr/lib64/riak/releases/1.1.1/riak             -embedded -config /etc/riak/app.config             -pa /usr/lib64/riak/basho-patches             -args_file /etc/riak/vm.args -- console
Root: /usr/lib64/riak
Erlang R14B04 (erts-5.8.5) [source] [64-bit] [smp:2:2] [rq:2] [async-threads:64] [hipe] [kernel-poll:true]

** /usr/lib64/riak/lib/observer-0.9.10/ebin/etop_txt.beam hides /usr/lib64/riak/lib/basho-patches/etop_txt.beam
** Found 1 name clashes in code paths
04:50:49.545 [info] Application lager started on node 'riak@test1'
04:50:49.584 [error] Supervisor riak_core_handoff_listener_sup had child riak_core_handoff_listener started with riak_core_handoff_listener:start_link() at undefined exit with reason bad argument in call to erlang:list_to_integer("::") in gen_nb_server:'-convert/1-lc$^0/1-0-'/1 in context start_error
/usr/lib64/riak/lib/os_mon-2.2.7/priv/bin/memsup: Erlang has closed.
                                                                     Erlang has closed
                                                                                      {"Kernel pid terminated",application_controller,"{application_start_failure,riak_core,{shutdown,{riak_core_app,start,[normal,[]]}}}"}

Crash dump was written to: /var/log/riak/erl_crash.dump
Kernel pid terminated (application_controller) ({application_start_failure,riak_core,{shutdown,{riak_core_app,start,[normal,[]]}}})
```

I assume this is convert() in the bottom of riak_core/src/gen_nb_server.erl, where it tries to naively split on a period ('.').  I believe if this were to simply be changed to split on either a period or a colon (for IPv6 addresses) it should work...

tomlanyon:
Update gen_nb_server to use inet_parse:address() to support both ipv4 + ipv6 addresses for Riak's handoff_ip.  Fixes #167.

### DIFF
--- a/src/gen_nb_server.erl
+++ b/src/gen_nb_server.erl
@@ -145,21 +145,17 @@ code_change(_OldVsn, State, _Extra) ->
 %% IpAddr = string()
 %% Port = integer()
 %% Result = {ok, port()} | {error, any()}
-listen_on(CallbackModule, IpAddr, Port) ->
-  SockOpts = [{ip, convert(IpAddr)}|CallbackModule:sock_opts()],
-  case gen_tcp:listen(Port, SockOpts) of
-    {ok, LSock} ->
-      {ok, _Ref} = prim_inet:async_accept(LSock, -1),
-      {ok, LSock};
+listen_on(CallbackModule, IpAddrStr, Port) ->
+  case inet_parse:address(IpAddrStr) of
+    {ok, IpAddr} ->
+      SockOpts = [{ip, IpAddr}|CallbackModule:sock_opts()],
+      case gen_tcp:listen(Port, SockOpts) of
+        {ok, LSock} ->
+          {ok, _Ref} = prim_inet:async_accept(LSock, -1),
+          {ok, LSock};
+        Err ->
+          Err
+      end;
     Err ->
       Err
   end.
- 
-%% @hidden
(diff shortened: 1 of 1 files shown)


evanmcc:
Hi Tom, 

This looks good, but could you add support for the address-as-tuple cases as well?

Also the whitespace looks weird.  Our code standard is more or less what emacs produces in erlang mode with:

(setq-default tab-width 4)
(setq-default indent-tabs-mode nil)

seancribbs:
Closing in favor of #249.

PR #246, opened 2012-10-27
Vnode shutdown message severity should be info

slfritchie:
Currently, when a vnode is shutdown, it's possible for handoffs in progress to be aborted and generate a warning severity message like this:

```
2012-10-26 20:22:09.489 UTC [error] <0.7853.42>@riak_core_handoff_sender:start_fold:215 ownership_handoff transfer of riak_kv_vnode from 'riak@10.10.0.51' 158424493890361546797469742608893960138484875264 to 'riak@10.10.0.53' 158424493890361546797469742608893960138484875264 failed because of error:{badmatch,{error,vnode_shutdown}} [{riak_core_handoff_sender,start_fold,5,[{file,"src/riak_core_handoff_sender.erl"},{line,161}]}]
```

This PR changes the error severity to an info message and also omits the stack trace, see below for an example.

There are two extra commits here, one to add a bit of bogus demo code to simulate the vnode shutdown without resorting to trying to provoke a race (in case any reviewer were curious).  The other commit removes that code. The tell-tale sign that the artificial trigger is hit and the new info severity response looks like this:

```
2012-10-26 19:08:47.897 [warning] <0.26559.0> Ha, caught a FOO {riak_core_fold_req_v1,#Fun<riak_core_handoff_sender.1..........
2012-10-26 19:08:47.897 [info] <0.27857.0>@riak_core_handoff_sender:start_fold:161 ownership_handoff transfer of riak_kv_vnode from 'dev3@127.0.0.1' 296867520082839655260123481645494988367611297792 to dev4@127.0.0.1' 296867520082839655260123481645494988367611297792 failed because the local vnode was shutdown
```

### DIFF
--- a/src/riak_core_handoff_sender.erl
+++ b/src/riak_core_handoff_sender.erl
@@ -31,6 +31,10 @@
 %% note this is in seconds
 -define(STATUS_INTERVAL, 2).
 
+-define(log_info(Str, Args),
+        lager:info("~p transfer of ~p from ~p ~p to ~p ~p failed " ++ Str,
+                   [Type, Module, SrcNode, SrcPartition, TargetNode,
+                    TargetPartition] ++ Args)).
 -define(log_fail(Str, Args),
         lager:error("~p transfer of ~p from ~p ~p to ~p ~p failed " ++ Str,
                     [Type, Module, SrcNode, SrcPartition, TargetNode,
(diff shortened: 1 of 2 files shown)


kellymclaughlin:
Looks good. +1 to merge.

PR #247, opened 2012-10-31
Change all gen_server:call() calls to use infinity timeout

slfritchie:
If the handoff manager gets busy enough to not reply in the
5 second default timeout, then the clients that time out and
crash can cause a supervisor to fail due to too many failures,
which in turn can trigger other supervisor failures that will
cause the entire app to shut down.  E.g.,

```
2012-10-31 01:04:43.370 UTC [error] <0.214.0> Supervisor riak_core_handoff_listener_sup had child riak_core_handoff_listener started with riak_core_handoff_listener:start_link() at {restarting,<0.16005.43>} exit with reason reached_max_restart_intensity in context shutdown
```

### DIFF
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -93,8 +93,6 @@ xfer({SrcPartition, SrcOwner}, {Module, TargetPartition}, FilterModFun) ->
 %% @doc Associate `Data' with the inbound handoff `Recv'.
 -spec set_recv_data(pid(), proplists:proplist()) -> ok.
 set_recv_data(Recv, Data) ->
-    %% Let this timeout and crash receiver if handoff mgr is
-    %% unresponsive.
     gen_server:call(?MODULE, {set_recv_data, Recv, Data}, infinity).
 
 status() ->
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -74,10 +74,10 @@ init([]) ->
     {ok, #state{excl=ordsets:new(), handoffs=[]}}.
 
 add_outbound(Module,Idx,Node,VnodePid) ->
-    gen_server:call(?MODULE,{add_outbound,Module,Idx,Node,VnodePid}).
+    gen_server:call(?MODULE,{add_outbound,Module,Idx,Node,VnodePid},infinity).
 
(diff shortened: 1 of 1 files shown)


src/riak_core_handoff_manager.erl
```diff
@@ -95,13 +95,13 @@ xfer({SrcPartition, SrcOwner}, {Module, TargetPartition}, FilterModFun) ->
 set_recv_data(Recv, Data) ->
     %% Let this timeout and crash receiver if handoff mgr is
     %% unresponsive.
-    gen_server:call(?MODULE, {set_recv_data, Recv, Data}).
+    gen_server:call(?MODULE, {set_recv_data, Recv, Data}, infinity).
```
seancribbs, line 96: Given the comment and that this seems to update the status of an existing transfer, why is this not a cast? On the other hand, why was it important to crash the receiver in the original version? /cc @rzezeski

seancribbs, line 96: According to our mumble discussion, removing the comment above the `gen_server:call` should be sufficient.

seancribbs:
:+1:

src/riak_core_handoff_manager.erl
rzezeski, line 96: @seancribbs @slfritchie This comment is referring to the fact that in order [kill_repairs](https://github.com/basho/riak_core/blob/master/src/riak_core_vnode_manager.erl#L830) to work correctly the _module_ and _partition_ data **MUST** be set before the receiver may continue.  When a repair is invoked 2 handoffs must execute.  One from each adjacent partition to the partition under repair.  Those two handoffs will result in two incoming handoff receivers.  The problem is that if you look at [receive_handoff](https://github.com/basho/riak_core/blob/master/src/riak_core_handoff_manager.erl#L487) you'll notice it doesn't add this data, and it can't because the information is not provided at that time.  Thus the data has to be set from the [initialization phase](https://github.com/basho/riak_core/blob/master/src/riak_core_handoff_receiver.erl#L97) of the handoff receiver.

The whole time I was writing this explanation I expected to end by saying that this needs to stay as a call.  But thinking about it more I'm not sure a cast would hurt since both the messages go to the same server and both are local calls.  So I feel like I've just muddied the waters now.

To be safe, I say leave it a call and remove the comment.

PR #248, opened 2012-11-01
Logical Deletion of Ring Meta & Bucket Properties

jrwest:
Ability to reset bucket properties to default and remove other ring metadata as described in https://github.com/basho/riak_core/issues/197 and originally implemented in https://github.com/basho/riak_core/pull/202. The difference between this implementation and the original is deletion is logical (a tombstone value is written) instead of physical. Tombstones are never reaped currently but it is possible to do so in the future. 

Corresponding `riak_test` for this functionality can be found here: https://github.com/basho/riak_test/pull/81

### DIFF
--- a/src/riak_core_bucket.erl
+++ b/src/riak_core_bucket.erl
@@ -29,6 +29,7 @@
          set_bucket/2,
          get_bucket/1,
          get_bucket/2,
+         reset_bucket/1,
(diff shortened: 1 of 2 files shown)


jerith:
Awesome, thanks!

smn:
\o/

PR #250, opened 2012-11-05
Make vnode check for existing handoff before starting another

jtuple:
Change riak_core_vnode to keep track of the handoff sender pid and not initiate a new handoff if an existing handoff is already running. In practice, this is not necessary as the handoff manager ensures that only one handoff is running per source/target pair. This change is an optimization to avoid additional work that is potentially expensive on some backends.

**Note:** This is tangentially related to basho/riak_kv#423 as one of the blocking scenarios was "double handoff". The second handoff attempt would call `bitcask:is_empty` and therefore trigger the reported blocking behavior. That issue is resolved by the Bitcask changes associated with the issue, but changing handoff behavior to not unnecessarily attempt to handoff is a reasonable optimization to do regardless.

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -94,6 +94,7 @@ behaviour_info(_Other) ->
           modstate :: term(),
           forward :: node(),
           handoff_node=none :: none | node(),
+          handoff_pid :: pid(),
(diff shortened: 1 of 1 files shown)


reiddraper:
+1

PR #252, opened 2012-11-06
Add a per stat calculation process for fine grained control

russelldb:
Per stat process for fine grained concurrency control of stat
calculation and caching.

A simple 'query' module for selecting stats by path from folsom.


src/riak_core_stat_q.erl
```diff
@@ -0,0 +1,108 @@
...
+%% are named with a tuple key. The key (like `{riak_kv, node, gets}` or
+%% `{riak_kv, vnode, puts, time}`) can
+%% be seen as an hierarchical path. With `riak_kv` at the root and
+%% the other elements as branches / leaves.
+%% This module allows us to get only the stats at and below a particular key
+%% `Path` is a list of atoms or the empty list.
+%% an example path might be `[riak_kv]` which will return every
+%% stat that has `riak_kv` in the first element of its key tuple.
+%% You may use the atom `'_'` at any point
+%% in `Path` as a wild card.
```
coderoshi: You use `{riak_kv, node, gets}` as an example, however, only lists work, not tuples... is the doc or the code out of sync?

coderoshi:
Looks good, works great, just a question about using `get_stats`. Assuming it's just a doc glitch, +1

src/riak_core_stat_q.erl
russelldb: Neither, but haven't been very clear (clearly.) The keys in folsom (the actual stat names) are tuples. But you must pass a list to this function.

src/riak_core_stat_q.erl
```diff
@@ -0,0 +1,108 @@
...
+calc_stat({Name, {metric, _Tags, histogram, _HistLen}}) ->
+    folsom_metrics:get_histogram_statistics(Name);
+calc_stat({Name, {metric, _Tags, _Type, _HistLen}}) ->
+    folsom_metrics:get_metric_value(Name).
+
+%% some crazy people put funs in folsom gauges
+%% so that they can have a consistent interface
+%% to access stats from disperate sources
+calc_guage(Val) when is_function(Val) ->
+    Val();
```
beerriot: It might be better to match a `{Module, Function}` tuple, and call `Module:Function()`, because of hot code loading, explained more in [my comment on riak_api's PR18](https://github.com/basho/riak_api/pull/18/files#r2115818)

src/riak_core_stat_calc_proc.erl
```diff
@@ -0,0 +1,143 @@
...
+check_freshness(TStamp, TTL) ->
+    case (TStamp + TTL) > folsom_utils:now_epoch() of
+        true ->
+             hit;
+        false ->
+            stale
+    end.
+
+maybe_get_stat(Stat, From, undefined, Awaiting) ->
+    %% if a process for  getting stat value is not underway start one
```
beerriot, line 132: This looks like it _always_ starts a new process for getting the stat value. Am I missing another check somewhere? Oh … or that it throws a `function_clause` error when the second waiter arrives, because `#state.active` is no longer `undefined`?

Confirmed. The patch is simple, though (just a second `maybe_get_stat/4` clause matching `Pid` to the third argument). I'll add it on this branch momentarily.

Fixed by lines 134 and 135 from commit d369f25. Demonstration, because I thought using redbug was fun: https://gist.github.com/46c4671b95028f9c583c

src/riak_core_stat_q.erl
```diff
@@ -0,0 +1,108 @@
...
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% @doc riak_core_stat_q is an interface to query folsom stats
+%%      To use, call `get_stats/1` with a query `Path`.
+%%      A `Path` is a list of atoms. The module creates a set
+%%      of `ets:select/1` guards, one for each element in `Path`
+%%      For each stat that has a key that matches `Path` we calculate the
+%%      current value and return it. This module mkaes use of
```
beerriot: typo: "mkaes"

src/riak_core_stat_calc_proc.erl
russelldb, line 132: Huh, yes. Erm, I have no idea why the code in your patch wasn't part of the code already…thank you.

russelldb:
added commit that addresses review comments and also allows binaries in stat key names (see https://github.com/basho/riak_kv/pull/414 for discussion)

beerriot:
+1 looks good!

slfritchie:
A friendly note to all participants ... today is code freeze day.

russelldb:
ooops. smerged already.

PR #254, opened 2012-11-24
Export path and stat_name types from riak_core_stat_q

beerriot:
These types are useful in other modules providing stats (a riak_pipe PR will make use of them shortly).


seancribbs:
:+1:

PR #255, opened 2012-11-29
upgrade legacy ring only if needed

vinoski:
Check for a legacy ring before upgrading it; for non-legacy rings, this
eliminates the confusing unconditional "Upgrading legacy ring" message
users mentioned on the riak-core mailing list.


beerriot:
:+1: thanks!

PR #258, opened 2012-12-17
health check changes

jrwest:
- change the checking processes to use `gen_server:cast` instead of `exit` for message passing between it and the `riak_core_node_watcher` process for valid return values (true/false). Invalid return values are still handled via exit.
- provide functions (`resume_health_checks/suspend_health_checks`) on `riak_core_node_watcher` to enable/disable all health checks. This will also toggle the new
  `healths_enabled` flag in the node_watcher's state which is used to prevent re-starting checks when a node goes down and comes back up

### DIFF
--- a/src/riak_core_node_watcher.erl
+++ b/src/riak_core_node_watcher.erl
@@ -203,22 +203,28 @@ handle_call({service_up, Id, Pid, MFA, Options}, From, State) ->
 
     State2 = remove_health_check(Id, State1),
 
-    %% install the health check
-    CheckInterval = proplists:get_value(check_interval, Options,
+    case app_helper:get_env(riak_core, enable_health_checks, true) of
+        true ->
+            %% install the health check
+            CheckInterval = proplists:get_value(check_interval, Options,
                                         ?DEFAULT_HEALTH_CHECK_INTERVAL),
-    IntervalTref = case CheckInterval of
-        infinity -> undefined;
-        N -> erlang:send_after(N, self(), {check_health, Id})
+            IntervalTref = case CheckInterval of
+                               infinity -> undefined;
+                               N -> erlang:send_after(N, self(), {check_health, Id})
+                           end,
+            CheckRec = #health_check{
+              callback = MFA,
+              check_interval = CheckInterval,
+              service_pid = Pid,
+              max_health_failures = proplists:get_value(max_health_failures, Options, 1),
+              max_callback_failures = proplists:get_value(max_callback_failures, Options, 3),
+              interval_tref = IntervalTref
+             },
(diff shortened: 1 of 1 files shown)


jtuple:
Haven't done a full review yet, but one thing I immediate notice is that we're missing the ability to disable health checks via app.config. Having the resume/suspend functions is great, but we need to provide a way for users to setup a cluster of nodes such that the new health check code is disabled upon start-up and remains disabled across node restarts.

jrwest:
ah good point. I was focusing on the "something w/ the health check is breaking everything, shut it off quickly" scenario. Will add something to app.config.

jrwest:
pushed change addressing @jtuple's comment re: disable via app.config

src/riak_core_node_watcher.erl
```diff
@@ -120,6 +123,12 @@ check_health(Service) ->
     ?MODULE ! {check_health, Service},
     ok.
 
+suspend_health_checks() ->
+    gen_server:call(?MODULE, suspend_healths).
+
+resume_health_checks() ->
+    gen_server:call(?MODULE, resume_healths).
+
```
jtuple: To be safe, let's make both gen_server calls have infinity timeout. If node watcher goes crazy for some reason, we don't want to be stuck failing on default 5s timeout when trying to disable it.

jtuple:
Tested health check returning true and false; throwing an errors; as well as performing an exit(normal). All works as expected. Although, a check that exit(normal) is basically ignored. No big deal, don't write checks that exit(normal).

Tested suspending/resuming on a node that has health checks enabled. Works as expected. Correct, but easily overlooked behavior to note, is that if a service becomes disabled due to being unhealthy, and then health checks are suspended, the service isn't coming back up even when healthy. Again, this is obvious and correct behavior. But, just wanted to note in case someone stumbles across the pull-request down the road when trying to understand this scenario.

Lastly, disabling health check in the app.config works. One thing to note is that this is a semi-hard disable. Changing the app var live doesn't suddenly enable health checks. Likewise, calling `resume_health_checks` doesn't enable things either. Either change app.config and restart, or change app var and then restart `riak_core_node_watcher` (eg. `exit(whereis(riak_core_node_watcher), kill)`) to re-enable health checks.

Code looks good.

We'll want to do a pull-request against Riak to add `{enable_health_checks, true}` to the app.config w/ a comment. Also, we'll want to write a solid `riak_test` that covers both the underlying health check code and the `riak_kv` health check logic. Including tests that verify behavior of suspend, resume, disabling from app.config, etc.

In any case, that isn't part of this specific pull-request. +1 merge.

PR #259, opened 2012-12-17
Adjust riak_core_sup child order for cleaner shutdown

jtuple:
The child definition for riak_core_handoff_manager is incorrect, starting before several other components that it depends upon. These dependency violations aren't hit in startup, but do occasionally trigger errors at shutdown. When Riak fails to shutdown cleanly, BEAM will delay shutdown by an additional 10 minutes or so. This pull-request moves the riak_core_handoff_manager definition lower in the child definition list to fix this issue.


joedevivo:
I tested this with riak_test on the cents 6 builder, where I can consistently reproduce the shutdown delay with the verify_capabilities test. While this change didn't stop the long shutdown issues entirely. It did stop some error messages and process crashes I was seeing in the logs during the long shutdown.

+1 to merge

rzezeski:
This looks good to me, and the tests pass on this branch (they fail when merged into master, but master itself has a regression somewhere involving core_vnode_eqc).  Given that Joe has verified that it helped shutdown I also say +1 to merge.

PR #260, opened 2012-12-18
Fix vnode eqc test regression

rzezeski:
The commit d769cfb2645050aa3abd053e260840d1a2893ae6 broke the test
because now instead of always going through the mgr it will
shortcut and hit the ets tab directly.  This means there can be a
slice of time where the supervisor reports a vnode pid but the ets tab
hasn't indexed it yet.  Calling `get_vnode_pid` after starting the
vnode ensures that all previous msgs in vnode mgr mailbox have been
handled and the ets tab is up-to-date for all subsequent calls.

### DIFF
--- a/src/bloom.erl
+++ b/src/bloom.erl
@@ -198,6 +198,9 @@ simple_shuffle(L) ->
     {_, L3} = lists:unzip(lists:keysort(1, L2)),
     L3.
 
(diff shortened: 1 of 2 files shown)


joedevivo:
after increasing some delays here, it works for me. +1

PR #261, opened 2012-12-19
Fix bug in riak_core_format:human_time + add test

jtuple:
Fix bug with `riak_core_format:human_time` as well as add test to verify correct results. For example, the expected output of the test values is:

```
5.0 us
5.0 ms
5.0 s
5.0 min
5.0 hr
5.0 d
```

which this branch will give you. Whereas master will print:

```
5.0 us
5.0 ms
5.0 s
300.0 s
18.0 min
7.2 hr
```

### DIFF
--- a/src/riak_core_format.erl
+++ b/src/riak_core_format.erl
@@ -25,6 +25,10 @@
          human_size_fmt/2,
          human_time_fmt/2]).
 
+-ifdef(TEST).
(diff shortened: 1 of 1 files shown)


jrwest:
+1

rzezeski:
+1 from me as well

PR #262, opened 2012-12-20
Add ability to selectively disable incoming/outgoing handoff

jtuple:
This pull-request provides the ability to selectively disable incoming or outgoing handoff by setting an app var. This is designed mostly for support/recovery scenarios in which getting data onto or off-of a particular node is more important than the normal balanced in/out behavior of Riak.

Ideally, we would change `handoff_concurrency` to support an inbound and outbound concurrency rate, but this current approach is straightforward, simple, and has been used in certain customer clusters as a hotpatch in support situations. Might as well get it in to Riak proper for Riak 1.3, and re-evaluate alternate approaches in the future.


coderoshi:
+1, despite being a secret, undocumented configuration ;)

PR #264, opened 2013-01-02
Make vnode terminate backend for any exit reason

engelsanchez:
This should fix issue basho/riak_test#137, where partition repair was
killing vnodes with reason kill_for_test and experiencing sporading
bitcask data corruption.

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -579,21 +579,24 @@ terminate(Reason, _StateName, #state{mod=Mod, modstate=ModState,
     %% Shutdown if the pool is still alive and a normal `Reason' is
     %% given - there could be a race on delivery of the unregistered
     %% event and successfully shutting down the pool.
-    case is_pid(Pool) andalso is_process_alive(Pool) andalso ?normal_reason(Reason) of
-        true ->
-            riak_core_vnode_worker_pool:shutdown_pool(Pool, 60000);
-        _ ->
-            ok
-    end,
-    case ModState of
-        %% Handoff completed, Mod:delete has been called, now terminate.
-        {deleted, ModState1} ->
-            Mod:terminate(Reason, ModState1);
-        _ ->
-            Mod:terminate(Reason, ModState)
-    end;
-terminate(_, _, _) ->
-    ok.
+    try
+        case is_pid(Pool) andalso is_process_alive(Pool) andalso ?normal_reason(Reason) of
+            true ->
+                riak_core_vnode_worker_pool:shutdown_pool(Pool, 60000);
+            _ ->
+                ok
+        end
+    catch C:T ->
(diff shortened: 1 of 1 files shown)


src/riak_core_vnode.erl
```diff
@@ -575,11 +575,11 @@ handle_info(Info, StateName, State=#state{mod=Mod,modstate=ModState}) ->
     end.
 
 terminate(Reason, _StateName, #state{mod=Mod, modstate=ModState,
-        pool_pid=Pool}) when ?normal_reason(Reason) ->
```
rzezeski, line 578: Along with this change you can remove the 2nd clause for terminate since this one will always match.

src/riak_core_vnode.erl
```diff
@@ -575,11 +575,11 @@ handle_info(Info, StateName, State=#state{mod=Mod,modstate=ModState}) ->
...
+        pool_pid=Pool}) ->
     %% Shutdown if the pool is still alive and a normal `Reason' is
     %% given - there could be a race on delivery of the unregistered
     %% event and successfully shutting down the pool.
-    case is_pid(Pool) andalso is_process_alive(Pool) of
+    case is_pid(Pool) andalso is_process_alive(Pool) andalso ?normal_reason(Reason) of
```
rzezeski: If shutdown_pool errors then Mod:terminate will not be called.  Think we should use a try/catch/after, like so:

``` erlang
try
  shutdown_pool
catch _:_ ->
  log about failing to shutdown pool
after
  Mod:terminate
end
```

engelsanchez:
@rzezeski this tries to incorporate the changes we talked about. I ran it through the partition repair test with some hacked in dummy exceptions

engelsanchez:
@rzezeski Dat Travis build failed for no good reason. The one in the branch passed, some magical merge into master failed on the first line of the build script (cd ~/builds). So if you think the change looks good, we can probably merge.

src/riak_core_vnode.erl
```diff
@@ -575,25 +575,28 @@ handle_info(Info, StateName, State=#state{mod=Mod,modstate=ModState}) ->
...
-    ok.
+    try
+        case is_pid(Pool) andalso is_process_alive(Pool) andalso ?normal_reason(Reason) of
+            true ->
+                riak_core_vnode_worker_pool:shutdown_pool(Pool, 60000);
+            _ ->
+                ok
+        end
+    catch C:T ->
+        lager:error("Error while shutting down vnode worker pool ~p:~p", [C, T])
```
rzezeski: You also want to log the stack trace here.  That way, if it does fail, we know where it failed.  The error reason alone is often not enough to determine that.  You can get trace via `erlang:get_stacktrace()`.

engelsanchez:
@rzezeski there, added nice stack trace

rzezeski:
+1 to merge

PR #265, opened 2013-01-03
Fix bug in riak_core_util:rpc_every_member_ann

jtuple:
The utility fun `rpc_every_member_ann`, uses by several status/console commands, was written assuming `rpc:multicall` returned results in the same order as the provided node list. This is an invalid assumption. This commit fixes `rpc_every_member_ann` as well as adds several additional utility functions to `riak_core_util`: `pmap`, `multi_rpc`, `multi_rpc_ann`, and `multicall_ann`.

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -40,6 +40,13 @@
          orddict_delta/2,
          rpc_every_member/4,
          rpc_every_member_ann/4,
+         pmap/2,
+         multi_rpc/4,
+         multi_rpc/5,
+         multi_rpc_ann/4,
+         multi_rpc_ann/5,
+         multicall_ann/4,
+         multicall_ann/5,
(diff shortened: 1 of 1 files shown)


src/riak_core_util.erl
```diff
@@ -236,6 +243,22 @@ ensure_started(App) ->
 	    ok
     end.
 
+%% @doc Invoke function `F' over each element of list `L' in parallel,
+%%      returning the results in the same order as the input list.
+-spec pmap(function(), [node()]) -> [any()].
+pmap(F, L) ->
```
joedevivo, line 249: I'm not in love with the exact same code from riak_test being duplicated here, I can't think of anything better to do in a timely fashion. Just let the record show that we should do something about it eventually

joedevivo:
+1

Issue #227, opened 2012-09-11
Intermittent crash on startup


jonmeredith:
Intermittent crash on startup - 6 node cluster, 32 vnodes.

```
 {"Kernel pid terminated",application_controller,"{application_start_failure,riak_core,{bad_return,{{riak_core_app,start,[normal,[]]},{'EXIT',{{function_clause,[{orddict,fetch,['riak@mgmt.r1s15',[{'riak@mgmt.r1s16',[{{riak_core,staged_joins},[true,false]},{{riak_core,vnode_routing},[proxy,legacy]}]}]],[{file,[111,114,100,100,105,99,116,46,101,114,108]},{line,72}]},{riak_core_capability,renegotiate_capabilities,1,[{file,[115,114,99,47,114,105,97,107,95,99,111,114,101,95,99,97,112,97,98,105,108,105,116,121,46,101,114,108]},{line,417}]},{riak_core_capability,handle_call,3,[{file,[115,114,99,47,114,105,97,107,95,99,111,114,101,95,99,97,112,97,98,105,108,105,116,121,46,101,114,108]},{line,216}]},{gen_server,handle_msg,5,[{file,[103,101,110,95,115,101,114,118,101,114,46,101,114,108]},{line,588}]},{proc_lib,init_p_do_apply,3,[{file,[112,114,111,99,95,108,105,98,46,101,114,108]},{line,227}]}]},{gen_server,call,[riak_core_capability,{register,{riak_core,vnode_routing},{capability,[proxy,legacy],legacy,{riak_core,legacy_vnode_routing,[{true,legacy},{false,proxy}]}}},infinity]}}}}}}"}
```

evanmcc:
Is this not the mismatched nodename issue?

slfritchie:
@evanmcc Time for some belated repo curating ... "mismatched" meaning changing vm.args while leaving behind an old ring with a different node name?  Or something else?

Looking at a recent'ish version of riak_core_capability:renegotiate_capabilities() ... if the orddict:fetch is a orddict:find() instead, and if find doesn't find it, then ... do what the first func clause does, i.e., nothing?

jrwest:
this is a known nodename mismatch issue and duplicate of both https://github.com/basho/riak_core/pull/266 and https://github.com/basho/riak_core/pull/244. closing.

PR #266, opened 2013-01-06
Use riak_core_capabilities:get_supported/2 to avoid crash

janpieper:
I'm quite new to erlang and riak_core but i think i just found a bug which can cause a crash. Two weeks ago i already created an issue for Sean Cribbs' riak_id (see seancribbs/riak_id#3 for further details) because i thought the bug is located within riak_id. Since i got no response for about two weeks i tried to debug the error and found `orddict:fetch(node(), State#state.supported)` which will raise an error if `node()` cannot be found in `State#state.supported`. It seems like `get_supported/2` is already handling that case so i changed `orddict:fetch(...)` to `get_supported(...)`.

### DIFF
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -412,7 +412,7 @@ negotiate_capabilities(Node, Override, State=#state{registered=Registered,
 renegotiate_capabilities(State=#state{supported=[]}) ->
     State;
 renegotiate_capabilities(State) ->
-    Caps = orddict:fetch(node(), State#state.supported),
+    Caps = get_supported(node(), State),
     Overrides = get_overrides(Caps),
     State2 = negotiate_capabilities(node(), Overrides, State),
     process_capability_changes(State#state.negotiated,


slfritchie:
Time for some belated repo curating ... @jrwest @jtuple @jonmeredith I think this addresses #227.  If not, it'd be great to have another eye to bless this small PR.

jrwest:
I'm concerned about silently handling this case. Granted to users the crash is completely uninformative but if the node name changes, renegotiating capabilities may now succeed but the node name has not changed in the ring, etc. Perhaps we should catch the case earlier, instead, so the assumption `renegotiate_capabilities` makes in its use of `orddict:fetch` is valid?

EDIT: looks like https://github.com/basho/riak_core/issues/243 is what i'm talking about

slfritchie:
> EDIT: looks like #243 is what i'm talking about

Hrmm, so if #243 is merged (i.e., soon 'cause it's now been positively reviewed), then ... what, sorry?

jrwest:
@slfritchie you lost me. #243 is an issue not a PR? Was trying to say, if we handle the case that @evanmcc calls a stale ring file (where we would hit the code path this PR is meant to address) then if the node was exited this PR isn't needed.

jrwest:
oh wait, i see #244 attached to #243 now. I'm +1 on that approach as long as the node crashes but that PR replaces this one doesn't it?

slfritchie:
Hi, Jan.  Many thanks for the bugfix.  Evan's later fix in https://github.com/basho/riak_core/pull/244 takes care of the same problem while also addressing another problem.  Manymany apologies for the delay!  Closing.

PR #268, opened 2013-01-23
Fix riak_core_wm_urlmap

seancribbs:
When the webmachine dispatch list was moved to ETS, this resource was not changed to reflect that, and so it still hit the application env for the dispatch list, resulting in a badmatch error because the key was not set.

This addresses the Ruby client failures on GiddyUp, e.g. http://giddyup.basho.com/#/projects/riak/scorecards/9/9-285-client_ruby_verify-centos-6-64-memory/3036


engelsanchez:
OK, finally got the right setup and confirmed that the ruby client test fails without this, passes with it. :+1: :dancer:

PR #269, opened 2013-01-24
VClock QuickCheck (incomplete)

seancribbs:
This is a QuickCheck property for vclock. The property was created so that the `dominates/2` could be added, with confidence that other parts of the API are solid as well. `dominates/2` essentially implements the `>` relation (strict-order) where `descends/2` implements the `>=` relation (partial-order). This is/will be used in various parts of `riak_kv` where vclocks are compared.

The property also includes an incomplete test for pruning. The model does not yet prune properly, but we did discover that without all of the property bucket properties set, pruning does not occur at all. `vclock` assumes that they are all set and blindly compares integers to the atom `undefined` in places. This is either a bug or a design decision (in which case we should probably have predicates that warn/error when the bucket properties aren't set).  Regardless, the pruning portion of the statem property needs to be finished or discarded.

(h/t @joedevivo and @Quviq for their help on this)


slfritchie:
Time for some belated repo curating ... I've rebased this branch onto 'develop', but the test finds a counterexample quite quickly.  Going back to your original branch, it finds the same (?) thing quite quickly.  The example here fails after merge(), it isn't clear to my far-too-hasty reading if this is a result of the pruning bug that you mention.

```
Shrinking..(2 times)
[{set,{var,1},{call,vclock_qc,fresh,[]}},
 {set,{var,3},{call,vclock_qc,increment,[a,{var,1}]}},
 {set,{var,4},{call,vclock_qc,merge,[[{var,3}]]}}]

vclock_qc:fresh() ->
  {[{a, 0, undefined}, {b, 0, undefined}, {c, 0, undefined},
    {d, 0, undefined}, {e, 0, undefined}],
   []}
vclock_qc:increment(a,
            {[{a, 0, undefined}, {b, 0, undefined}, {c, 0, undefined},
              {d, 0, undefined}, {e, 0, undefined}],
             []}) ->
  {[{a, 1, 1}, {b, 0, undefined}, {c, 0, undefined},
    {d, 0, undefined}, {e, 0, undefined}],
   [{a, {1, 1}}]}
vclock_qc:merge([{[{a, 1, 1}, {b, 0, undefined},
           {c, 0, undefined}, {d, 0, undefined}, {e, 0, undefined}],
          [{a, {1, 1}}]}]) ->
  {[{a, 1, undefined}, {b, 0, undefined}, {c, 0, undefined},
    {d, 0, undefined}, {e, 0, undefined}],
   [{a, {1, 1}}]}

Reason: {postcondition, false}
```

seancribbs:
@slfritchie That looks like a different bug. The pruning problem was in setting the various pruning parameters: big/small/young/old_vclock. If they are not all set properly, pruning is pretty unpredictable, especially if some of them are unset.

seancribbs:
At @russelldb's request I've removed the pruning modeling from the test and rebased. Reopening against develop.

PR #271, opened 2013-02-03
Arg dialyzer fixes

argv0:
got rid of (i think) all the riak_core_ring warnings


jrwest:
nice! seems to get rid of all those "non-local return" warnings. The only warning I still see is: 

```
riak_core_ring.erl:1252: The pattern 'true' can never match the type 'false'
```

which I was only able to get rid of by exporting `equal_cstate` (because the `true` branch is only called by code defined in an `ifdef(TEST)` block).

PR #273, opened 2013-02-18
Notify caller of sync/spawn command errors

DeadZen:
sync_command times out when a vnode command errors or exits.. 
sync_spawn_command will wait for infinity under the same condition.

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -207,14 +207,20 @@ vnode_command(Sender, Request, State=#state{index=Index,
                                             forward=Forward,
                                             pool_pid=Pool}) ->
     %% Check if we should forward
-    case Forward of
+    Action = case Forward of
         undefined ->
-            Action = Mod:handle_command(Request, Sender, ModState);
+            case catch Mod:handle_command(Request, Sender, ModState) of
+                {'EXIT', ExitReason} ->
+                    reply(Sender, {vnode_error, ExitReason}),
+                    lager:error("~p command failed ~p", [Mod, ExitReason]),
+                    {stop, ExitReason, ModState};
+                Else -> Else
+            end;
         NextOwner ->
             lager:debug("Forwarding ~p -> ~p: ~p~n", [node(), NextOwner, Index]),
             riak_core_vnode_master:command({Index, NextOwner}, Request, Sender,
                                            riak_core_vnode_master:reg_name(Mod)),
-            Action = continue
+            continue
     end,
     case Action of
         continue ->
--- a/src/riak_core_vnode_master.erl
+++ b/src/riak_core_vnode_master.erl
@@ -117,16 +117,23 @@ sync_command({Index,Node}, Msg, VMaster, Timeout) ->
     %% Issue the call to the master, it will update the Sender with
     %% the From for handle_call so that the {reply} return gets
     %% sent here.
-    gen_server:call({VMaster, Node},
-                    make_request(Msg, {server, undefined, undefined}, Index), Timeout).
+    Request = make_request(Msg, {server, undefined, undefined}, Index),
+    case gen_server:call({VMaster, Node}, Request, Timeout) of
+        {vnode_error, {Error, _Args}} -> error(Error);
+        {vnode_error, Error} -> error(Error);
+        Else -> Else
+    end.
 
 %% Send a synchronous spawned command to an individual Index/Node combination.
 %% Will not return until the vnode has returned, but the vnode_master will
 %% continue to handle requests.
 sync_spawn_command({Index,Node}, Msg, VMaster) ->
-    gen_server:call({VMaster, Node},
-                    {spawn, make_request(Msg, {server, undefined, undefined}, Index)},
-                    infinity).
+    Request = make_request(Msg, {server, undefined, undefined}, Index),
+    case gen_server:call({VMaster, Node}, {spawn, Request}, infinity) of
+        {vnode_error, {Error, _Args}} -> error(Error);
+        {vnode_error, Error} -> error(Error);
+        Else -> Else
+    end.
 
 
 %% Make a request record - exported for use by legacy modules
--- a/test/mock_vnode.erl
+++ b/test/mock_vnode.erl
@@ -34,7 +34,9 @@
          asyncnoreply/2,
          asyncreply/2,
          asynccrash/2,
-         crash/1,
+         crash/1, 
+         spawn_error/2,
+         sync_error/2,
          get_crash_reason/1,
          stop/1]).
 -export([init/1,
@@ -94,6 +96,12 @@ asynccrash(Preflist, AsyncDonePid) ->
 crash(Preflist) ->
     riak_core_vnode_master:sync_command(Preflist, crash, ?MASTER).
 
+spawn_error(Preflist, Cmd) ->
+    riak_core_vnode_master:sync_spawn_command(Preflist, {sync_error, Cmd}, ?MASTER).
+
+sync_error(Preflist, Cmd) ->
+    riak_core_vnode_master:sync_command(Preflist, {sync_error, Cmd}, ?MASTER).
+
 get_crash_reason(Preflist) ->
     riak_core_vnode_master:sync_command(Preflist, get_crash_reason, ?MASTER).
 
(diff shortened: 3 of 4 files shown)


slfritchie:
I have no major problem with this PR.  There's a larger question of whether or not we want the vnode to crash when `Mod:handle_command()` and friends misbehave, but this PR leaves the current vnode-will-crash-when-Mod-misbehaves behavior effectively unchanged.  @jtuple Any comments when wearing your core/kv shepherd hat?

jonmeredith:
I would be concerned that this upsets consumers of vnodes. For example the get/put FSM in Riak would crash on any of these responses rather than wait for good nodes or timeout.

The larger scale 'core 2.0' type improvement would be to do better at defining the messaging for interacting with vnodes.  It seems like you could break most requests against a vnode to an incoming stream of messages that relate to a single logical request (with some kind of backpressure/flow control) creating an optional response of zero or messages back (with backpressure/flow control), including standard ways to say "the vnode or node is overloaded, I'm shedding work" and "the vnode crashed, here's why".  Supporting rolling upgrades and mixed clusters is such a pain.

I know it isn't perfect, but until something like that exists, your other option would be to implement your own handling mechanism in the vnode callback code

```
handle_command(Req, From, State) ->
  try
    really_handle_command(Req, From, State)
  catch
    _:Err ->
         {reply, {vnode_err, Err}, State}
  end
```

DeadZen:
I agree that those type of improvements would be of considerable benefit, however, this patch only triggers on a sync/sync_spawn command, which I don't believe the current get/put FSM behavior in Riak uses.
Furthermore it would only trigger on error, which would crash the respective vnode anyway, this simply enables the vnode consumer to be notified of that event instead of inexplicably converting any errors into timeouts.
Finally, yes placing try/catch's in every handle_command that could be called in this manner is a known and far from ideal approach which imho would only cover up an inherent problem at the wrong level of abstraction.

jonmeredith:
Agreed, try/catch is an ugly bandaid. core_vnode is a much nicer place to fix.

I was rushing a bit this morning when I first scanned it - as it's for the sync/sync_spawn versions it looks much nicer.  

Before we could merge, somebody would need to investigate interactions with AAE (riak_kv_index_hashtree), riak_search and the 2I format upgrade - but tbh I suspect they may not handle a crash very gracefully anyway.

DeadZen:
ping... any additional tasks for me to do here?

slfritchie:
Pedram, we're working on meeting a code freeze deadline coming up soon, sorry about the delay.

Would you mind changing the `'$vnode_error'` atom to simply `vnode_error`?  My jetlagged memory says that Ericsson has a convention that atoms that begin with `$` should be used by OTP libraries only.

Jon, I'll take a peek today.

DeadZen:
Ok cool, thank you very much, hopefully this makes it in before that.. 
Also good to know about that convention.. I've made those changes as requested...

slfritchie:
Alright, sorry, I'm being dense here.  The two timing conditions that you mention in the problem description ... why do they happen?  In both cases, the code today is using gen_server:call(), which creates a monitor to the destination process ... and that monitor should inform the client about any problem with the remote server vnode crashing.

DeadZen:
You're anything but dense Scott, this is just a well hidden bug.. and agreed, it should but, when the vnode dies from an error before it can reply to a synchronous call, the gen_server:call waits around until a timeout condition occurs... 
This might be an issue from the message having to propagate out from the vnode proxy system. 
You can verify the timeouts occur simply by adding my tests and not my fixes to riak_core_vnode\*

slfritchie:
Oops, sorry about the silence.

So there is another proxy-like process in the chain, which (of course) is impractical for the client side of the gen_server:call() function to know about.  I'm wondering about the wisdom of stopping after the first error.

It's quite possible that there are many other commands waiting in the vnode's mailbox.  If we stop & terminate, those will get thrown away, and you'll have the same problem: no reply will be sent, so you must wait for the client-side timeout.

An alternative is to keep the vnode running.  We caught the error, the vnode process is still alive.  The problem is that not all backends are 100% functionally pure.  If they all were pure, then we'd have no problem: we can return using the `ModState` that we started with.  But if the backend is not purely functional, e.g. `riak_kv_bitcask_backend` or `riak_kv_eleveldb_backend`, then we don't know if the backend's internal state is still consistent.  That damn thing may be assuming that any bug will kill the vnode and restart it in order to return to a sane state.

(Actually, I don't believe any of Riak's backends are 100% pure, except perhaps for the yessir backend).

An OTP-style way of defending against this would be to use a supervisor-like scheme: if there are fewer than `N` errors within `T` seconds, then you continue.  When `N` failures happen within `T` seconds, then shut down.  But I would still have to ponder if that behavior works well in all situations for Riak and non-Riak applications.

@yrashk @jonmeredith @jtuple Other opinions?

DeadZen:
No problem, thank you for returning to this issue. 

It's funny you mention keeping the vnode alive as my first draft to the mailing list did just that.
I think there is evidence to go in either direction; should the vnode continue to serve other requests?
Or should there be less tolerance for sick vnodes? I think the latter approach is actually the more OTP way.
From a current operational stand point, the vnode has already crashed, this is a do not resuscitate approach.  
This would only happen if there was an actual error or exit condition in the vnode code handling the request. 

Following gen_server semantics here..there is no concern for the contents of a processes mailbox when it errors, it is terminated with extreme prejudice. Since riak_core has more concerns wrt handoffs and maintaining a valid and current member ring status, maybe a combination approach of using a supervisor vnode death strategy with hooks to the c/c++ backends for supervisor messages and (presuming of course, this wasn't an executive decision for riak_core coordinators) an ets table of active requests to ets:give_away which could then notify all request  waiters/peers that it has taken over for it's counterpart .. @vinoski ?

At least it does seem that this is a known condition, which has never really been fully addressed..
http://lists.basho.com/pipermail/riak-users_lists.basho.com/2011-April/003794.html

And it does seem especially in the case of a NIF being involved that when completely decoupled from the cleanup of an exit condition with the vnode that there can be unintended side effects. 
http://riak-users.197444.n3.nabble.com/Node-shutdown-error-tp4025227p4025229.html

While we're contemplating, how should our approach differ for things like coverage requests that may error? 
Maybe we should reconsider if a vnode proxy is absolutely necessary? What was the special or fundamental reasoning behind this? Could everything be moved to the proxy, or could we somehow try to simplify the function of a vnode? 

All questions for the distributed computing think tank ;)

DeadZen:
ping

lenards:
I'm curious about the status of this too.

slfritchie:
Time for some belated repo curating ... @jonmeredith @vinoski Opinions?  I'm tempted to take the partial solution: perfect, enemy, good, yaddayadda....

Vagabond:
I just ran into this, we really really should address this for 2.0!

slfritchie:
Update: This was mentioned in last Friday's Basho engineering status meeting.  We agree that this is a significant problem and needs a fix sooner rather than later.

DeadZen:
Well... shall we merge? =D

slfritchie:
@jonmeredith @engelsanchez @jtuple @Vagabond The caremad locomotive is near full pressure and will be departing the station in the next few days.  Its destination is: https://github.com/basho/riak_core/compare/slf-integrate-deadzen-273 ... Your comments have the power to stop the train.  ^_^

jonmeredith:
As long as the caremad considers the way that every vnode request that we
make across kv, search1, search2 and pipe, it's a good thing to do, but may
be too late to successfully review it before hitting freeze.

I really wish we'd defined two interfaces for vnode commands, a
request/response and a streaming one - both with some kind of backpressure,
but perhaps we can get there from here.

Jon

On Thu, Dec 12, 2013 at 1:29 AM, Scott Lystig Fritchie <
notifications@github.com> wrote:

> @jonmeredith https://github.com/jonmeredith @engelsanchezhttps://github.com/engelsanchez
> @jtuple https://github.com/jtuple @Vagabondhttps://github.com/VagabondThe caremad locomotive is near full pressure and will be departing the
> station in the next few days. Its destination is:
> https://github.com/basho/riak_core/compare/slf-integrate-deadzen-273 ...
> Your comments have the power to stop the train. ^_^
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/basho/riak_core/pull/273#issuecomment-30396597
> .

## 

Jon Meredith
VP, Engineering
Basho Technologies, Inc.
jmeredith@basho.com

DeadZen:
Works with riak_kv and riak_pipe here, and has for nearly a year.
I don't personally use riak_search. I can't imagine it would do
anything bad for riak_search, at worst just report more accurate
errors; which aren't common in official basho releases, but it's
quite common in development with riak_core.

The issue might still exist in the case of vnode_coverage calls. I
haven't targeted that only vnode_commands, so affected areas
should be fairly insignificant since even then it's only for the sync
or sync spawn cases, which I'm not aware of any basho libs using.

It is my sincere goal to improve riak_core in efficiency, simplicity
and readability. I feel this is a small, and possibly temporary step.
I know this is the most straight forward fix for a pretty thorny problem,
however, I always welcome more feedback basho has on these issues.

It is likely that this patch should not be necessary, but I think to make it
unnecessary would require a much more involved change to the codebase.
Would it not? @jonmeredith @engelsanchez @jtuple @Vagabond @vinoski

I leave it in the capable hands of the caremad train conductor. @slfritchie

Thanks,
Pedram

slfritchie:
I'll merge on Tuesday JST if there's objection.

slfritchie:
Merged in commit 632af2b3ce and then commit acd2a390c.  Many, many apologies again for the delay!  Closing.

PR #274, opened 2013-02-21
Allow parallel vnode initialization

engelsanchez:
Solution to basho/platform_tasks#18

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -41,6 +41,7 @@
          rpc_every_member/4,
          rpc_every_member_ann/4,
          pmap/2,
+         pmap/3,
          multi_rpc/4,
          multi_rpc/5,
          multi_rpc_ann/4,
@@ -51,6 +52,7 @@
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
+-export([counter_loop/1,incr_counter/1,decr_counter/1]).
 -endif.
 
 %% R14 Compatibility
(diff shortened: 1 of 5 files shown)


src/riak_core_util.erl
```diff
@@ -454,5 +530,72 @@ build_tree_test() ->
...
+        {counter_value, N} -> N
+    after
+        3000 ->
+            ?assert(false)
+    end.
+
+decr_counter(CounterPid) ->
+    CounterPid ! down.
+
+bounded_pmap_test_() ->
```
jonmeredith, line 558: Should #ifdef(TEST) around this

engelsanchez, line 558: It was added to an existing ifdef TEST region, so it doesn't show up in the diff, but it is.

jonmeredith, line 558: I think I'll make some phone calls or call a meeting or something.

engelsanchez, line 558: Don't give into the dark side Jon!!!!!! :)
http://i.livescience.com/images/i/000/020/685/iFF/luke-skywalker-111005.jpg?1317834193

evanmcc:
do we have any benchmarks or assessments of performance impact?

engelsanchez:
Only on a developer laptop with an SSD, 16 partitions and 12GB of bitcask data. The riak_kv service took 4min+ to initialize without this, but only 1min + a few seconds with this change. All coming from being able to better saturate the I/O with concurrent work.  This will be tested on real clusters ASAP. If you want to lend a hand, let me know. I know you have something set up and still warm that perhaps we could use?  With the latest commit, now concurrency can be controlled with the previously existing vnode_rolling_start parameter. The default should be good (16), and you would get the old, sequential behavior if set to 1 or less.  Tweaking on rotational drives should be interesting as I would expect it to turn around and get worse as you tune that setting up. I'm dying to see the real effect.

evanmcc:
code looks good

testing with 16 vnodes containing 320GB of data on a rotational disk raid system indicated a 10X speedup.

:+1:

PR #276, opened 2013-02-25
Add encoding capability for handoff.

chardan:
Motivation:
Zlib forces serialization on an internal Erlang lock when opening new port drivers, which we are concerned may adversely affect scalability on larger systems. This feature provides a new capability for specifying the encoding method used for handoff, and offers a new option to use raw encoding so that users may choose a tuning most appropriate for their situation.

Note: Since capabilities are not re-negotiated, in some corner cases of rolling upgrade, it's possible for handoff to crash if a capability changes mid-up/downgrade; in this case, handoff will still finish, as it is designed to crash and will restart. Future work is expected to eliminate this case by encoding capabilities for the entire handoff rather than per-session.

See also:
https://github.com/basho/riak_kv/pull/495
https://github.com/basho/riak_test/pull/223

### DIFF
--- a/src/riak_core_app.erl
+++ b/src/riak_core_app.erl
@@ -78,7 +78,7 @@ start(_StartType, _StartArgs) ->
             riak_core:register(riak_core, [{stat_mod, riak_core_stat}]),
             ok = riak_core_ring_events:add_guarded_handler(riak_core_ring_handler, []),
 
-            %% Register capabilities
+            %% Register capabilities (note: first setting in capability list is default ):
             riak_core_capability:register({riak_core, vnode_routing},
                                           [proxy, legacy],
                                           legacy,
@@ -89,7 +89,7 @@ start(_StartType, _StartArgs) ->
                                           [true, false],
                                           false),
             riak_core_capability:register({riak_core, handoff_data_encoding},
-                                          [encode_zlib, encode_raw],
+                                         [encode_raw, encode_zlib],
                                           encode_raw),
 
             {ok, Pid};
--- a/src/riak_core_app.erl
+++ b/src/riak_core_app.erl
@@ -88,6 +88,9 @@ start(_StartType, _StartArgs) ->
             riak_core_capability:register({riak_core, staged_joins},
                                           [true, false],
                                           false),
+            riak_core_capability:register({riak_core, handoff_data_encoding},
+                                          [encode_zlib, encode_raw],
+                                          encode_raw),
 
             {ok, Pid};
         {error, Reason} ->
--- a/src/riak_core_handoff_sender.erl
+++ b/src/riak_core_handoff_sender.erl
@@ -272,7 +272,7 @@ visit_item(K, V, Acc) ->
 
     case Filter(K) of
         true ->
-            { BinObj, _EncodingMethod } = Module:encode_handoff_item(K, V),
+            BinObj = Module:encode_handoff_item(K, V),
 
             M = <<?PT_MSG_OBJ:8,BinObj/binary>>,
             NumBytes = byte_size(M),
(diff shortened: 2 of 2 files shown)


src/riak_core_handoff_sender.erl
```diff
@@ -272,7 +272,8 @@ visit_item(K, V, Acc) ->
 
     case Filter(K) of
         true ->
-            BinObj = Module:encode_handoff_item(K, V),
+            { BinObj, _EncodingMethod } = Module:encode_handoff_item(K, V),
```
jonmeredith: You could make this backward compatible with a 

```
{BinObj, _EncodingMethod} = case Module:encode_handoff_item(K, V) of
                                              BinObj where is_binary(BinObj) ->
                                                 {BinObj, encode_zlib}; %% or whatever your default should be
                                              EncodedItem ->
                                                 EncodedItem
                                            end,
```

But perhaps this is part of a multi-step PR that I don't have full context on.

jonmeredith: Don't forget riak_core is used in projects outside Riak.

chardan: An excellent reminder! In this case, the encoding type tuple is an artifact of an earlier iteration of the code and now no longer needed at all. Joe and I thought it would be good information to keep around, but you're right-- it does break that API. Rather than requiring code at the call site, maybe another helper function that gets the encoding type from the current capability set (or default) and lets the user's code decide rather than just returning it is the way to go now that the information's not currently used. 
(Note: This is indeed a change in both kv and core; also affects riak_test.).

src/riak_core_app.erl
```diff
@@ -78,7 +78,7 @@ start(_StartType, _StartArgs) ->
             riak_core:register(riak_core, [{stat_mod, riak_core_stat}]),
             ok = riak_core_ring_events:add_guarded_handler(riak_core_ring_handler, []),
 
-            %% Register capabilities
+            %% Register capabilities (note: first setting in capability list is default ):
```
engelsanchez, line 81: I believe the default is given in the third parameter to register, not from the first list item. This comment might be misleading.

src/riak_core_handoff_sender.erl
```diff
@@ -273,6 +273,7 @@ visit_item(K, V, Acc) ->
     case Filter(K) of
         true ->
             BinObj = Module:encode_handoff_item(K, V),
+
```
engelsanchez, line 276: Unfortunate whitespace only change. Tidy up those commits before pushing! :)

src/riak_core_app.erl
```diff
@@ -88,6 +88,9 @@ start(_StartType, _StartArgs) ->
             riak_core_capability:register({riak_core, staged_joins},
                                           [true, false],
                                           false),
+            riak_core_capability:register({riak_core, handoff_data_encoding},
```
jrwest, line 91: any particular reason we are registering this capability in `riak_core` vs. `riak_kv`. Isn't the zlib encoding specific to `riak_kv_vnode`?

engelsanchez, line 91: Da man is right: encoding/decoding is not being handled or queried by core, this should move.

src/riak_core_app.erl
jtuple, line 81: Default is third parameter. But, order of list determines preference. First item is the "most preferred" capability, and therefore the one chosen if all nodes speak it.

src/riak_core_app.erl
chardan, line 91: I'll move this. Thank you both!

chardan:
Moved capability to riak_kv.

PR #277, opened 2013-02-28
Move riak-admin reference to riak_kv

engelsanchez:
riak-admin is a Riak specific script, so refactoring to avoid mention in
riak_core.

riak_kv takes over this message in basho/riak_kv#497 and basho/riak#284 points the script to the kv console


jrwest:
+1

engelsanchez:
Dropping for now as there are many other places doing the same (mentioning riak-admin) in core. We need a better solution.

PR #281, opened 2013-03-08
Porting parallel vnode init fix to 1.3 + revert switch

engelsanchez:


jtuple:
Overall, things work as intended and there doesn't appear to be any major issues. Although, have a few minor comments.

Also, given that this work has already merged into `master` and this a backport, it's unclear what changes we want to do here versus on `master`. Could ship with this PR, change `master` later, or change things both here and on `master`. Not sure what works best.

---

The main issue is the [change](https://github.com/basho/riak_core/pull/281/files#L1L126) of `Mod:start_vnode` moving from being passed an index to a list of indices. This assumes that such a change is safe for all `riak_core` apps. Currently, this isn't safe for `riak_search`, but you already have a simple PR to fix that. Of course, we have no idea what people are doing with `riak_core` in the wild. One option is to instead keep the semantics of `start_vnode` the same and add a new function such as `start_vnodes` that accepts a list. And then to fallback to `start_vnode` if the module in question doesn't implement `start_vnodes`. Of course, this requires all `riak_core` apps to be modified with this new `start_vnodes` function in order to use parallel vnode initialization. On the other hand, it does make it so this is no longer an API breaking change. The main idea would be to move to something like:

``` erlang
case lists:member({start_vnodes, 1}, Mod:module_info(exports)) of
    true ->
        Mod:start_vnodes(Startable);
    false ->
        [Mod:start_vnode(I) || I <- Startable]
end.
```

---

My remaining comments are best left as line notes. Going to go add those now.

src/riak_core_util.erl
```diff
@@ -259,6 +261,80 @@ pmap(F, L) ->
...
+                 lists:foldl(fun pmap_worker/2,
+                             #pmap_acc{mapper=Mapper,
+                                       fn=Fun,
+                                       max_concurrent=MaxP},
+                             List),
+    % Collect pending work
+    Collect =
+        fun(Pid, Acc) ->
+                receive
+                    {pmap_result, Pid, R} ->
```
jtuple: I'm not a fan of using selective receive on `Pid` here. That can lead to `O(N^2)` mailbox scans, as well as keeping items in the mailbox longer then necessary. To illustrate, let's say we have pids A-D, but due to worker timing or whatever we receive messages in the order D-A. The first iteration of `Collect` is going to do a `receive {pmap_result, <A>, R}`, skip over messages D-B, find A, return. Then for B it'll start over, re-scan until it finds B at the end, return. Rinse/repeat for the rest. Likewise, if A took longer than the other workers, we'd stick around waiting for A while keeping D-B messages sitting in the mailbox the entire time.

For vnode init, this may not be a major issue. But, given that `pmap/3` is designed to be a general purpose utility function, we might as well consider what happens if someone decides to use 1000 workers or something. Really don't want to scan 1 million messages in the worst case.

Any reason you can't just do a receive on `{pmap_result, _, R}`, thus handling each response in the order they are sent? It looks like you end up sorting the results on line 297 so you're not particularly sensitive to how you end up constructing `All`.

On a different note, I really wish we could use optimized receives in both this `pmap/3` as well as the older `pmap/2`. But, the strict requirements of how you must structure your code to get the Erlang compiler to use the optimized receive won't work for the way either of these pmap implementations are coded. This is unfortunate, as each `receive` for the `pmap_result` is going to scan over all non-pmap messages in the mailbox. This is kinda annoying for the `riak_core_vnode_manager` which receives 100s to 10000s of messages regularly. One of the top serialization points in all of Riak.

One option would be to do the pmap itself in a separate process. Ie. calling pmap spawns a process that then spawns the workers, waits for results, then sends the combined results back to the original process. This would decouple the calling processes mailbox from the pmap process. Anyway, food for thought.

src/riak_core_util.erl
```diff
@@ -259,6 +261,80 @@ pmap(F, L) ->
...
+                        [R|Acc]
+                end
+        end,
+    All = lists:foldl(Collect, Done, Pending),
+    % Restore input order
+    Sorted = lists:sort(fun({I1, _}, {I2, _}) -> I1 < I2 end, All),
```
jtuple: I would prefer using `lists:keysort` here for same effect.

src/riak_core_util.erl
```diff
@@ -259,6 +261,80 @@ pmap(F, L) ->
...
+    lists:map(fun({_, R}) -> R end, Sorted).
```
jtuple: Prefer using a list comprehension over `lists:map` here. LCs are more concise, [faster](https://gist.github.com/jtuple/af55f3489cc1e685eff7), and keep getting faster every few Erlang releases due to smarter compiler optimizations.

src/riak_core_vnode.erl
```diff
@@ -130,6 +133,39 @@ send_command_after(Time, Request) ->
 
 init([Mod, Index, InitialInactivityTimeout, Forward]) ->
     process_flag(trap_exit, true),
+    State = #state{index=Index, mod=Mod, forward=Forward,
+                   inactivity_timeout=InitialInactivityTimeout},
+    case app_helper:get_env(riak_core, revert_parallel_vnode_init, false) of
```
jtuple: Any reason to go with a negative option ala `revert_parallel_vnode_init` versus a more standard option ala `parallel_vnode_init` that you just set to false to disable/"revert"? Second case seems more consistent with existing naming conventions in Riak.

src/riak_core_vnode_manager.erl
```diff
@@ -510,17 +512,42 @@ delmon(MonRef, _State=#state{idxtab=T}) ->
...
+                 lager:debug("Will start VNode for partition ~p", [Idx]),
+                 {ok, Pid} =
+                     riak_core_vnode_sup:start_vnode(Mod, Idx, ForwardTo),
+                 lager:debug("Started VNode, waiting for initialization to complete ~p, ~p ", [Pid, Idx]),
+                 ok = riak_core_vnode:wait_for_init(Pid),
+                 lager:debug("VNode initialization ready ~p, ~p", [Pid, Idx]),
+                 {Idx, Pid}
+        end,
+    MaxStart = app_helper:get_env(riak_core, vnode_rolling_start,
+                                  ?DEFAULT_VNODE_ROLLING_START),
```
jtuple, line 540: I would suggest having a separate option for defining the maximum number of vnodes started in parallel rather than re-using the existing `vnode_rolling_start` option. Esp. since the `vnode_rolling_start` option is still used for the existing rolling start logic which has nothing to do with starting primary vnodes. I could conceivably see folks wanting different values for these two different concepts.

To recap how Riak operates, when a node first comes online it must start up all primaries vnodes before marking a service as up. This code path is handled in `riak_core_ring_handler` and is the thing that calls `start_vnode` and therefore this newer parallel init code. After starting all primaries, the service is marked up and Riak starts operating as normal. However, Riak also starts all secondary vnodes at least once per node restart, to trigger any potential hinted handoff. This is handled by the rolling vnode start logic in `riak_core_vnode_manager` (see: `maybe_start_vnodes`) that starts up to `vnode_rolling_start` vnodes every vnode manager tick (with a frequency determined by `vnode_management_timer` config). Likewise, this code path starts vnodes sequentially, and doesn't use the new parallel vnode logic at all (it would be great to change rolling start logic to use parallel init as well, but I'd consider that out-of-scope for this PR).

src/riak_core_vnode_manager.erl
```diff
@@ -510,17 +512,42 @@ delmon(MonRef, _State=#state{idxtab=T}) ->
...
+    Pairs = Started ++ riak_core_util:pmap(StartFun, NotStarted, MaxStart),
+    % Return Pids in same order as input
+    [begin
+         {_, Pid} = lists:keyfind(Idx, 1, Pairs),
```
jtuple: I highly suggest not using `lists:keyfind` here, as such + the outer LC leads to O(N^2) behavior. At least 50% of the work I've done on the "large ring" branch has been replacing code like this because we assumed in the past that the input list would always be small and turned out to be wrong "in the large".

Easy fix is to convert `Pairs` into a `dict` before the LC, eg:

``` erlang
PairsDict = dict:from_list(Pairs),
[begin
     {_, Pid} = dict:fetch(Idx, 1, PairsDict),
     ...
 end || Idx <- IdxList].
```

jtuple:
I'm a bit curious about the `started(timeout, ...)` and `started(wait_for_init, ....)` cases in `riak_core_vnode`. It seems to me that we'll always transition into `started(timeout, ...)` immediately from `init` before ever receiving any `wait_for_init` messages, and the call to `riak_core_vnode:wait_for_init` in the vnode manager ends up just hitting the default case in the `active` state [here](https://github.com/basho/riak_core/blob/1e543e49fe909925f6292c73b4a6c016e0223843/src/riak_core_vnode.erl#L367).

Am I mistaken? Or is the entire `wait_for_init` code path dead?

src/riak_core_util.erl
engelsanchez: Collect only handles the last batch of work (a number of messages never bigger than max_concurrent). Within the main fold, messages are received in the order they arrived (see line 328). I can fix this collection of the last batch for 1.4, but I could see that as something easy to mess up right before shipping this.

I'm aware of the perils of selective receives in this general purpose utility, but thought that any application sensitive to that should run pmap in a separate process that will only collect pmap messages anyway. That way it can handle its mailbox in whatever way is more efficient (say a gen_fsm would spawn a function that ran pmap and sent an event back to it, or a gen_server would send itself back a call, etc).

In the case of initialization, we have already been blocking here for hours at a time without processing messages, so I don't think that this particular case is any worse. It should be a lot better now just due to the fact that it will likely block for a 10th of the time it did before.

src/riak_core_util.erl
engelsanchez: No problem, will change. I didn't realize that existed.

src/riak_core_util.erl
engelsanchez: Will change. It must have been a random synapse firing that made me write it like this.

src/riak_core_vnode.erl
engelsanchez: Maybe I'll just have parallel_vnode_init be a number config, use that instead of rolling for the concurrency and here check if 1 or less. Having one less config would be great.

src/riak_core_vnode_manager.erl
engelsanchez: Will do. I don't want this to become problematic with large number of vnodes.

engelsanchez:
@jtuple about the started state: my reading of the OTP code was that timeout is simply the timeout of the receive of the main loop, which is executed after acking the process intialization. This is the race condition I was worried about: somebody initializating the vnode asynchronously and sending it a message immediately (or another process sending a request right there). This could potentially arrive before the receive code in gen_fsm is touched, and so that message would be processed and passed to the initial state (see loop function in gen_fsm.erl). I think this would only happen once in a blue moon, but it could unless I missed something.

engelsanchez:
@jtuple Just pushed changes to address your comments (with some stuff left for later). Please let me know if you agree or have objections and will address immediately

engelsanchez:
@jtuple BTW, don't forget this now has a related PR in riak_kv for this to work basho/riak_kv#511

engelsanchez:
@jtuple I went ahead and changed pmap to avoid using lists and lists:member (using sets now). Eunit test passes. Hopefully this addresses some of your concerns.  Anyway, this list of pending work was bounded by max_concurrent anyway, but I guess why use lists:member when you don't have to...

jtuple:
Like how you refactored to reuse `parallel_collect_one`, and the `<= 1` option to `vnode_parallel_start` to disable the changes is a nice approach. Everything else looks great as well.

Re-tested things manually. Also tested with and without the accompanying `riak_kv` PR and verified that `start_vnode` / `start_vnodes` was called correctly depending on the code. 

+1

PR #282, opened 2013-03-12
Extract out and export pending claim function.

cmeiklejohn:
Necessary for basho/riak_control to be able to compute future claim.

@jtuple


evanmcc:
:+1:

PR #284, opened 2013-03-13
initial add of the Riak Core Connection Manager

bookshelfdave:
This is Chris Tilt's work, I'm just moving files between repos.
see repl PR 212.

cc @buddhisthead @lordnull 

**NOTE** Ranch has been added as a dependency for the connection manager tests to run.

### DIFF
--- a/src/riak_core_service_mgr.erl
+++ b/src/riak_core_service_mgr.erl
@@ -382,7 +382,7 @@ choose_version({ClientProto,ClientVersions}=_CProtocol, HostProtocols) ->
                     {error,protocol_version_not_supported,Rest};
                 [{Major,CN,HN}] ->
                     {ok, {ClientProto,Major,CN,HN},Rest};
-                [{Major,CN,HN}, _] ->
+                [{Major,CN,HN} | _] ->
                     {ok, {ClientProto,Major,CN,HN},Rest}
             end
     end.
--- a/src/riak_core_service_mgr.erl
+++ b/src/riak_core_service_mgr.erl
@@ -1,0 +1,453 @@
+%% -------------------------------------------------------------------
+%%
+%% Riak Subprotocol Server Dispatcher
+%%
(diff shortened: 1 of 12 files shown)


buddhisthead:
Dave, after more consideration, I believe that we should remove the functions around cluster_finder from the connection manager. It is currently used only by the test modules and I think it is non-used API. We should remove it now, before making it open source, because the connection manager should be devoid of cluster-related connections.

This functionality has been replaced by remote targets, registration of them, etc. The cluster_finder is no longer used. Please cut it all out of the connection_mgr and from the associated test.

src/riak_core_service_mgr.erl
```diff
@@ -0,0 +1,453 @@
...
+            case lists:reverse(lists:keysort(1,CommonVers)) of
+                [] ->
+                    %% oops! No common major versions for Proto.
+                    ?TRACE(?debugFmt("Failed to find a common major version for protocol: ~p",
+                                     [ClientProto])),
+                    lager:error("Failed to find a common major version for protocol: ~p", [ClientProto]),
+                    {error,protocol_version_not_supported,Rest};
+                [{Major,CN,HN}] ->
+                    {ok, {ClientProto,Major,CN,HN},Rest};
+                [{Major,CN,HN}, _] ->
```
lordnull: This line should likly be:

```
[{Major,CN,HN} | _] ->
```

As is only 2 version permutations are allowed, anything else will get a case_clause error.

I found this while building a riak_test for cascading repl, which supports 2.0, 1.1, and 1.0 versions.

buddhisthead: Ah! Good catch!
+1 to that change

bookshelfdave:
I added Ranch for the connection manager tests, however it looks like the dep doesn't work with R14B04 | R14B03
https://travis-ci.org/basho/riak_core/builds/5595971

buddhisthead:
Tests Pass.
+1

PR #286, opened 2013-03-15
Rather than calculate stats on demand when stale, backfill the cache

russelldb:
Always server the stats that are in the cache, no matter how old they are.
Add a timestamp to the stats so consumers know how stale they are.
Fill the cache continuously in the background.

See https://github.com/basho/riak_kv/issues/505 and https://github.com/basho/riak_kv/issues/506 for graphs, bench configs etc.

### DIFF
--- a/src/riak_core_stat_cache.erl
+++ b/src/riak_core_stat_cache.erl
@@ -18,7 +18,7 @@
 %%
 %% -------------------------------------------------------------------
 
-%% @ doc fetches stats for registered modules and stores them
+%% @doc fetches stats for registered modules and stores them
 %% in an ets backed cache.
 %% Only ever allows one process at a time to calculate stats.
 %% Will always serve the stats that are in the cache.
@@ -37,6 +37,8 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
+-type registered_app() :: {MFA::mfa(), RerfreshRateMillis::non_neg_integer()}.
+
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 -endif.
@@ -46,6 +48,7 @@
 -define(REFRESH_RATE, 1).
 -define(REFRSH_MILLIS(N), timer:seconds(N)).
 -define(ENOTREG(App), {error, {not_registered, App}}).
+-define(DEFAULT_REG(Mod, RefreshRateMillis), {{Mod, produce_stats, []}, RefreshRateMillis}).
 
 -record(state, {tab, active=orddict:new(), apps=orddict:new()}).
 
@@ -61,7 +64,7 @@ register_app(App, {M, F, A}) ->
     register_app(App, {M, F, A}, RefreshRate).
 
 register_app(App, {M, F, A}, RefreshRateSecs) ->
-    gen_server:call(?SERVER, {register, App, {M, F, A}, ?REFRSH_MILLIS(RefreshRateSecs)}).
+    gen_server:call(?SERVER, {register, App, {{M, F, A}, ?REFRSH_MILLIS(RefreshRateSecs)}}).
 
 get_stats(App) ->
     gen_server:call(?SERVER, {get_stats, App}, infinity).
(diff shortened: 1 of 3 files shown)


src/riak_core_stat_cache.erl
```diff
@@ -18,14 +18,20 @@
 %%
 %% -------------------------------------------------------------------
 
+%% @ doc fetches stats for registered modules and stores them
```
rzezeski: I think this needs to be `@doc` ;)

src/riak_core_stat_cache.erl
```diff
@@ -87,29 +101,42 @@ handle_call({get_stats, App}, From, State0=#state{apps=Apps, active=Active0, tab
...
-                            {reply, {ok, Stats, TS}, State0}
+                            FreshnessStat = make_freshness_stat(App, TS),
+                            {reply, {ok, [FreshnessStat | Stats], TS}, State0}
                     end
             end,
     Reply;
+handle_call({clear, App}, _From, State=#state{apps=Apps, tab=Tab}) ->
+    Reply = case registered(App, Apps) of
+                false ->
+                    {reply, ?ENOTREG(App), State};
```
rzezeski: I could go either way on this but perhaps just return `ok` in this case too?  The benefit I see is less chance of a `badmatch` error.  If an app doesn't exist then it's already cleared, right?

src/riak_core_stat_cache.erl
```diff
@@ -136,10 +167,19 @@ code_change(_OldVsn, State, _Extra) ->
...
+    {make_freshness_stat_name(App), TS}.
+
+make_freshness_stat_name(App) ->
+    list_to_atom(atom_to_list(App) ++ "_stat_ts").
+
+register_mod(App, Mod, Fun, Args, RefreshRateMillis, Apps) ->
     folsom_metrics:new_histogram({?MODULE, Mod}),
     folsom_metrics:new_meter({?MODULE, App}),
-    orddict:store(App, {Mod, Fun, Args, TTL}, Apps).
+    orddict:store(App, {Mod, Fun, Args, RefreshRateMillis}, Apps).
```
rzezeski: I doubt it matters at all in terms of performance but it's interesting that there are 3 different data structures representing the Mod/Fun/Arg/RefreshRate data.
1. Here you store it as `{M, F, A, RR}`
2. Each is passed to `register_mod` as its own argument
3. The register `handle_call` has `{M, F, A}` as and `RR`

Maybe just pass around as `{{M, F, A}, RR}` everywhere to be consistent?  I dunno, I'm just nitpicking and at end of day shouldn't really matter.

src/riak_core_stat_cache.erl
```diff
@@ -87,29 +101,42 @@ handle_call({get_stats, App}, From, State0=#state{apps=Apps, active=Active0, tab
     Reply = case registered(App, Apps) of
                 false ->
                     {reply, ?ENOTREG(App), State0};
-                {true, {M, F, A, TTL}} ->
-                    case cache_get(App, Tab, TTL) of
-                        No when No == miss; No == stale ->
+                {true, {M, F, A, _RefreshRateMillis}} ->
+                    case cache_get(App, Tab) of
+                        No when No == miss ->
```
rzezeski: Why not just `miss ->`?

russelldb: Sorry, used to be `miss` or `stale`, quick fixed it. Will change.

src/riak_core_stat_cache.erl
russelldb: Yeah, I only added it for the test really and cutandpasted it. `ok` will do fine.

src/riak_core_stat_cache.erl
```diff
@@ -136,10 +167,19 @@ code_change(_OldVsn, State, _Extra) ->
...
 %% internal
-register_mod(App, Mod, Fun, Args, TTL, Apps) ->
+schedule_get_stats(After, App, MFA) ->
+    erlang:send_after(After, ?SERVER, {get_stats, {App, MFA}}).
+
+make_freshness_stat(App, TS) ->
+    {make_freshness_stat_name(App), TS}.
+
+make_freshness_stat_name(App) ->
+    list_to_atom(atom_to_list(App) ++ "_stat_ts").
```
rzezeski, line 181: Is there any reason we should worry about making stat names atoms?  Is there any way to users to cause stat names to be created?

russelldb, line 181: You have to register a module with riak_core:register to get to be a stat name. Or call riak_core_stat_cache:register. But the `App` must already be an atom. So we're taking an existing atom and adding "_ts" to it.

src/riak_core_stat_cache.erl
russelldb: Ok,  I agree. It is messy and confusing. However, for the 1.3.1 release I don't want to change the `register_app` public interface to the `{{M, F, A}, RR}`, as that will mean changing all those mods that call that function, but from that boundary inwards I will change it to be consistent. Is that OK?

russelldb:
Pooshed changes that address comments. Please to re-review.

src/riak_core_stat_cache.erl
```diff
@@ -67,18 +80,22 @@ stop() ->
...
     process_flag(trap_exit, true),
     Tab = ets:new(?MODULE, [protected, set, named_table]),
-    TTL = app_helper:get_env(riak_core, stat_cache_ttl, ?TTL),
+    RefreshRateSecs = app_helper:get_env(riak_core, stat_cache_ttl, ?REFRESH_RATE),
+    RefreshRateMillis = ?REFRSH_MILLIS(RefreshRateSecs),
     %% re-register mods, if this is a restart after a crash
-    RegisteredMods = lists:foldl(fun({App, Mod}, Registerd) ->
-                                         register_mod(App, Mod, produce_stats, [], TTL, Registerd) end,
+    RegisteredMods = lists:foldl(fun({App, Mod}, Registered) ->
+                                         register_mod(App, ?DEFAULT_REG(Mod, RefreshRateMillis), Registered),
```
jrwest: think this needs to be:

```
Registered1 = register_mod(...),
schedule_get_stats(...)
Registered1
```

or something to that effect

jrwest: actually, seems there is a similar pattern below when handling `{register, App, Registraton}`. Any reason to not just roll `schedule_get_stats` into `register_mod`?

russelldb: Shit, yes it should, good catch.

I'll see about rolling schedule into `register_mod`

jrwest:
code changes look good. tests pass (including `verify_riak_stats`). +1.

russelldb:
Squashed and merged.

PR #287, opened 2013-03-19
Make stats more robust in the face of failure

russelldb:
In some cases underlying ets tables for stats go away. When this
happens the effected stats break and stay broken.

When a stat is broken the stat calculation throws an error. For
the sake of robustness this commit wraps stat calculation in a try
catch, and returns the atom `unavailable` if a stat cannot be
calculated. Broken stats are expected to be detected and repaired
when they are updated.


russelldb:
Pooshed.

engelsanchez:
:+1: :dancer:

PR #288, opened 2013-03-22
Failure to calculate a stats value should be temporary so warn only

russelldb:


seancribbs:
:+1:

engelsanchez:
Gave it a spin, works as advertised, trivial code :+1: :dancer:

PR #290, opened 2013-04-01
Add support for tracking progress of individual handoffs

jrwest:
Vnodes can optionally provide the size of their data (in bytes or # of objects/keys) when returning from `is_empty/1` (the original API is also preserved). The size returned should be considered an upper-bound on the amount of bytes/objects that may be sent during handoff. For vnodes whose size may change during handoff (e.g. a non read-isolated iterator in riak_kv like ets) a function may be provided instead. The function is called to calculate the size of the vnode each time transfer stats are requested from the handoff manager. The provided size as well as the existing handoff stats are used to calculate completion percentage. 

This PR also moves the `transfers/1` console function from `riak_kv_console` to `riak_core_console`. It is not specific to kv and other applications should be able to take advantage of it. ~~In addition, `transfers/1` now prints completion percentage w/ progress bar and can optionally take a number of seconds, N, and a maximum, M. The output
is printed every N seconds a total of M times. If M is not given or is not a number the output will be printed every N seconds until CTRL+C is pressed or the nodetool timeout is reached. If N is not a number it is ignored.~~

EDIT: based on review, the  support for calling `transfers/1` passing in seconds/max calls was removed since this can be done in the shell.

### DIFF
--- a/include/riak_core_handoff.hrl
+++ b/include/riak_core_handoff.hrl
@@ -34,6 +34,7 @@
           vnode_mon             :: reference(),
           type                  :: ho_type(),
           req_origin            :: node(),
-          filter_mod_fun        :: {module(), atom()}
+          filter_mod_fun        :: {module(), atom()},
+          size                  :: {non_neg_integer(), bytes | objects}
         }).
 -type handoff_status() :: #handoff_status{}.
--- a/src/riak_core_console.erl
+++ b/src/riak_core_console.erl
@@ -23,7 +23,7 @@
          stage_leave/1, stage_remove/1, stage_replace/1,
          stage_force_replace/1, print_staged/1, commit_staged/1,
          clear_staged/1, transfer_limit/1, pending_claim_percentage/2,
-         pending_nodes_and_claim_percentages/1]).
+         pending_nodes_and_claim_percentages/1, transfers/1]).
 
 %% @doc Return list of nodes, current and future claim.
 pending_nodes_and_claim_percentages(Ring) ->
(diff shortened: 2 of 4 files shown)


engelsanchez:
About the output from the console command: It would be nice to have a visual separator between the different calls. It's hard to track progress visually when doing riak-admin <n> and scrolling back and forth.

engelsanchez:
Also, I've been finding that riak-admin transfers times out frequently. I leave it running and then it had stopped a while back. Would it be possible to leave it repeatedly retrying when a timeout occurs until Ctrl-C is hit?

jrwest:
re: visual seperator. What did you have in mind? could you toss up some quick example output of what it might look like (or are you just talking about printing something like `---------`)? 

re: timeout, are you referring to the node-tool timeout or the legacy gossip timeout? The former I'm not sure how much control we have from inside the console commands but i think its a nice safety feature in case you leave `riak-admin transfers` running on a machine (its cheap but its not free work). The latter we should probably handle more gracefully although in reality this should go away once legacy gossip is removed. Do you have some example output from the latter if thats the case? EDIT: i did not see the latter case in my testing. also, what handoff concurrency were you running at?

engelsanchez:
Yes, a simple line of characters in between runs would be great.  The timeouts are nodetool RPC timeouts, I presume:

RPC to 'dev2@127.0.0.1' failed: timeout

This makes it more convenient for users to run the tool as as bash loop instead of using the interval and max parameters.  like so:

while sleep 3; do riak-admin transfers; done

which would simply print the timeout but keep retrying. I think a simply try/catch in the repeat function would be ok. If you are repeating the command over and over, a temporary failure probably shouldn't kill it.  Say, someone uses this to pipe to a log to do some fancier things with it.

Another thought is, maybe we should just skip those parameters altogether and just let people run this in shell loops if they want it to run every X seconds and that? That way you can completely avoid going down this little rabbit hole of reimplementing simple shell commands in our console code.

jrwest:
Getting back to this finally, sorry for the delay. So I agree that its probably simpler to remove the code I added around rechecking transfers and just offload that to bash. fwiw, the timeout I _think_ comes from nodetool so its not something you can try/catch in that function call but I could be mistaken. Anyways, that won't really be a problem if we just remove the looping.

Ill get this up to date w/ master and that code removed soon

engelsanchez:
I've tested the latest simplified version and eyeballed the code which I had reviewed already. It all looks good and works as advertised
:+1: :dancer: :boat:

PR #291, opened 2013-04-02
SSL support

Vagabond:

PR #292, opened 2013-04-03
use custom Ranch build to support R14B03|4

bookshelfdave:
custom Ranch tag excludes -callbacks so builds can occur in R14B03 + R14B04.


Vagabond:
The builds fail because this fix isn't on master yet:

https://github.com/basho/riak_core/pull/291/files#L4R54

I say merge it.

PR #297, opened 2013-04-08
don't use hardcoded app names in SSL utils

bookshelfdave:
cc @Vagabond

### DIFF
--- a/src/riak_core_ssl_util.erl
+++ b/src/riak_core_ssl_util.erl
@@ -41,14 +41,14 @@ maybe_use_ssl(App) ->
         {cacerts, load_certs(app_helper:get_env(App, cacertdir, undefined))},
         {depth, app_helper:get_env(App, ssl_depth, 1)},
         {verify_fun, {fun verify_ssl/3,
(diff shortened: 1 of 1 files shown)


src/riak_core_ssl_util.erl
```diff
@@ -41,14 +41,14 @@ maybe_use_ssl(App) ->
         {cacerts, load_certs(app_helper:get_env(App, cacertdir, undefined))},
         {depth, app_helper:get_env(App, ssl_depth, 1)},
         {verify_fun, {fun verify_ssl/3,
-                get_my_common_name(app_helper:get_env(App, certfile,
-                        undefined))}},
+                {get_my_common_name(app_helper:get_env(App, certfile,
```
Vagabond: The tuple field ordering disagrees with

``` erlang
verify_ssl(Cert, valid_peer, {App, MyCommonName})
```

below.

bookshelfdave: That's what I get for coding while mumbling.

Vagabond:
+1, merge it!

Issue #298, opened 2013-04-11
Race in vnode worker pool


beerriot:
When a riak_core_vnode_worker finishes work, it sends checkin messages to both poolboy and riak_core_vnode_worker_pool. The latter maintains a queue of work to be handled when there's room in the pool. As soon as RCVWP gets the checkin message, it asks poolboy if there is a worker available (expecting that the worker just checked in will now be available).

The problem is that poolboy may receive RCVWP's message before receiving the worker's checkin message. If this happens, it will tell RCVWP that the pool is full. RCVWP then sticks in the 'queueing' state until it receives another checkin message from a worker. Since another checkin may never arrive, the pool may become frozen.

The test defined by worker_pool_pulse.erl on the [bwf-pool-race branch of riak_core](https://github.com/basho/riak_core/tree/bwf-pool-race) demonstrates this race. Under PULSE execution, the test will fail with deadlock.

In order to run the test, you will need the pulse_otp beams from https://github.com/Quviq/pulse_otp on your path. You will also need to compile poolboy with PULSE annotations - the [bwf-pool-race branch of basho/poolboy](https://github.com/basho/poolboy/tree/bwf-pool-race) provides this.

Since I think PULSE's graphical output is quite cool, I'll include it here:

![PULSE-prop-small-pool](https://f.cloud.github.com/assets/190975/370108/a85815d0-a2e7-11e2-8b48-a9afabe698b5.jpg)

Key:
- "root" (red): test process
- "spawn_opt" (blue): riak_core_vnode_worker_pool
- "spawn_opt1" (cyan): poolboy fsm
- "spawn_opt2" (green): poolboy_sup
- "spawn_opt3" (magenta): poolboy_worker/riak_core_vnode_worker

The problem is illustrated by the last four steps of "spawn_opt1", the poolboy process (in cyan), and the last six steps of "spawn_opt", the riak_core_vnode_worker_pool_process (in blue). The `$gen_all_state_event` that is sent to spawn_opt from the last step in spawn_opt3 is one of the two checkin messages, and the `$gen_sync_event` sent to spawn_opt1 is the other. The `$gen_event` sent from spawn_opt to spawn_opt1 is the checkout message, and since spawn_opt1 receives that message before the checkin is delivered, you can see it reply `{_,full}`. After that, everything is stuck waiting with no way forward.

Now fixing begins…

beerriot:
[...]

PR #300, opened 2013-04-15
Fix worker pool races

beerriot:
Races in the vnode worker pool, which could cause the pool to become frozen, were described in #298. This PR resolves those races.
## Resolutions

The fix for the checkin race is for the worker to only checkin to riak_core_vnode_worker_pool, and allow riak_core_vnode_worker_pool to decide when to checkin to poolboy. If there is outstanding work in the queue, the worker will just be reused immediately instead of being checked in.

The fix for the DOWN race is for the worker to send a 'worker_started' message to riak_core_vnode_worker_pool, when the worker starts. While the worker is sending this message, poolboy is blocking, so riak_core_vnode_worker_pool can send poolboy a checkout message immediately, and still be guaranteed that it will be received after poolboy has put the worker in its pool.
## Testing

The new test `worker_pool_pulse` requires modules to be compiled with the `PULSE` macro defined, so it must be run on its own. A new make target, `pulse`, has been added to do just this. Before running the `pulse` target, ensure three things:
1. You have [Quviq's pulse_otp](https://github.com/Quviq/pulse_otp) on your path.
2. You have a copy of poolboy with PULSE annotations (currently in [the bwf-pool-race branch of basho/poolboy](https://github.com/basho/poolboy/tree/bwf-pool-race)
3. You have not yet built beams for `riak_core` or `poolboy` (so, either right after `./rebar get-deps` or `make clean`)

Running the tests using just the first commit in this PR should fail with EQC/PULSE complaining of `deadlock`. If it fails for another reason, you may have missed one of the three requirements above (or I only achieved WOMM status). Using both commits, the tests should pass.

Only `worker_pool_pulse` will run when using the `pulse` make target. After running this test, and before running other tests, you should clean and recompile, or the PULSE annotations may cause other tests to fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 APPS = kernel stdlib sasl erts ssl tools os_mon runtime_tools crypto inets \
 	public_key mnesia syntax_tools compiler
 COMBO_PLT = $(HOME)/.riak_core_combo_dialyzer_plt
+PULSE_TESTS = worker_pool_pulse
 
 .PHONY: deps test
 
@@ -20,6 +21,12 @@ distclean: clean
 
 test: all
 	./rebar skip_deps=true eunit
+
+# You should 'clean' before your first run of this target
+# so that deps get built with PULSE where needed.
+pulse:
+	./rebar compile -D PULSE
+	./rebar eunit -D PULSE skip_deps=true suite=$(PULSE_TESTS)
 
 docs: deps
 	./rebar skip_deps=true doc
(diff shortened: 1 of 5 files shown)


engelsanchez:
I have repeatedly run the test with and without the fix with the expected result.  I've read the code and changes up and down and I can not see any potential problems with this. Everything looks clean. It's definitely better to centralize control of the checkins in the worker pool code.  As far as I'm concerned :+1: :dancer: :boat: . Thanks @beerriot!

PR #301, opened 2013-04-15
Ring Resizing

jrwest:
Ring Resizing enables `riak_core` applications to change the number of partitions (running vnodes) during normal operations, under load. Previously, a cluster was limited to always having `ring_creation_size` partitions. In order to change the number of partitions, a seperate cluster would need to be stood up along side the original and the data would be migrated between the two by external means.

The intended purpose of this feature is to support users who create a cluster with either too few or two many partitions and have a need to change this without all the hassle mentioned above. It _is not intended_ as a scaling feature for clusters to add or remove concurrent processing ability. Riak already has support for that via adding/removing nodes. However, since the number of partitions ultimately limits the number of nodes in the cluster,  ring resizing can be used to increase capacity in that regard. In short, the feature is intended for infrequent use in specific scenarios.

This feature is a work-in-progress that will undergo several more iterations. Some details on that work as well as issues with the current implementation are discussed below.

**For a more detailed description of the implemention see [here](https://github.com/basho/riak_core/blob/jrw-dynamic-ring/docs/ring-resizing.md).**
## Goals

Besides the obvious goal that the cluster should safely transition between the old and new rings -- ensuring requests for unmodified keys return the same data and new writes are reflected in bolth rings. The implemention was designed with the following goals in mind:
- During the operation requests for existing keys should not return `not_found`
- The minimum number of transfers to ensure safety should be made
- ~~The implementation should be ignorant of the N values used to store keys in Riak~~ The implementation should be as ignorant as possible of the N values used to store keys in riak
- The operation should be cancelable and support forceful replacement of nodes in the case that SHTF
## The Rest of the Story

There are some caveats and future work that are worth mentioning here.
- **Rehashing**: When the ring expands some existing partitions are not moved to new owners. In this case none of their data is removed. In the future, a "rehashing" operation will be submitted during the cleanup cycle that removes keys no longer owned by these existing partitions. 
- **Repl**: replication does not support clusters of different ring sizes. The only available approach that currently works is to disable realtime until bolth cluster have been resized and then restablish the repl link, at which point a full-sync will probably be necessary.
- **Perf. Improvements**: ~~There is a known optimization for how to determine if keys should be transferred that is not included here because it still has some kinks to work out. Performance improvements are also expected thanks to the many improvements coming to handoff/gossip in general.~~ Different optimization now included see [1] for microbenchmark. 
- **Memory & Storage Considerations**: Like the old method of resizing (using two clusters), to safely transition to the new ring, in the worst possible case, the entire data set is copied (an ugly consequence of not cleaning up data until the new ring is installed). Clusters should have the available disk space (& memory if using bitcask) to account for this.

[1] https://gist.github.com/jrwest/4940f009b2adb928ef59

### DIFF
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -122,7 +122,13 @@ replace(Node, NewNode) ->
 force_replace(Node, NewNode) ->
     stage(Node, {force_replace, NewNode}).
 
-%% @doc Stage a request to resize the ring %% TODO: moar doc
+%% @doc Stage a request to resize the ring. If committed, all nodes
+%%      will participate in resizing operation. Unlike other operations,
+%%      the new ring is not installed until all transfers have completed.
+%%      During that time requests continue to be routed to the old ring.
+%%      After completion, the new ring is installed and data is safely
+%%      removed from partitons no longer owner by a node or present
+%%      in the ring.
 -spec resize_ring(integer()) -> ok | {error, atom()}.
 resize_ring(NewRingSize) ->
     %% use the node making the request. it will be ignored
@@ -438,12 +444,13 @@ valid_force_replace_request(Node, NewNode, Changes, Ring) ->
 %% @private
 %% restrictions preventing resize along with other operations are temporary
 valid_resize_request(NewRingSize, [], Ring) ->
-    case riak_core_ring:num_partitions(Ring) =/= NewRingSize of
-        true -> true;
-        false -> {error, same_size}
-    end;
-valid_resize_request(_, _Changes, _) ->
-    {error, non_empty_changes}.
+    Capable = riak_core_capability:get({riak_core, resizable_ring}, false),
+    IsResizing = riak_core_ring:num_partitions(Ring) =/= NewRingSize,
+    case {Capable, IsResizing} of
+        {true, true} -> true;
+        {false, _} -> {error, not_capable};
+        {_, false} -> {error, same_size}
+    end.
 
 %% @private
 %% @doc Filter out any staged changes that are no longer valid. Changes
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -436,12 +436,11 @@ valid_force_replace_request(Node, NewNode, Changes, Ring) ->
     end.
 
 %% @private
-%% restrictions preventing resize to smaller ring and along
-%% with other operations are temporary
+%% restrictions preventing resize along with other operations are temporary
 valid_resize_request(NewRingSize, [], Ring) ->
-    case riak_core_ring:num_partitions(Ring) < NewRingSize of
+    case riak_core_ring:num_partitions(Ring) =/= NewRingSize of
         true -> true;
-        false -> {error, smaller_or_equal_ring}
+        false -> {error, same_size}
     end;
 valid_resize_request(_, _Changes, _) ->
     {error, non_empty_changes}.
@@ -565,8 +564,7 @@ maybe_compute_resize(Orig, MbResized) ->
     OrigSize = riak_core_ring:num_partitions(Orig),
     NewSize = riak_core_ring:num_partitions(MbResized),
 
-    %% restriction on expanding ring only is temprorary
-    case OrigSize < NewSize of
+    case OrigSize =/= NewSize of
         false -> MbResized;
         true -> compute_resize(Orig, MbResized)
     end.
@@ -577,12 +575,24 @@ compute_resize(Orig, Resized) ->
     %% original ring will be reset when changes are scheduled
     CState0 = riak_core_ring:future_ring(Resized),
 
+    Type = case riak_core_ring:num_partitions(Orig) < riak_core_ring:num_partitions(Resized) of
+        true -> larger;
+        false -> smaller
+    end,
+
     %% Each index in the original ring must perform several transfers
     %% to properly resize the ring. The first transfer for each index
     %% is scheduled here. Subsequent transfers are scheduled by vnode
     CState1 = lists:foldl(fun({Idx, _} = IdxOwner, CStateAcc) ->
-                                  NextOwner = riak_core_ring:index_owner(CStateAcc, Idx),
-                                  schedule_first_resize_transfer(IdxOwner, NextOwner, CStateAcc)
+                                  %% indexes being abandoned in a shrinking ring have
+                                  %% no next owner
+                                  NextOwner = try riak_core_ring:index_owner(CStateAcc, Idx)
+                                              catch error:badarg -> none
+                                              end,
+                                  schedule_first_resize_transfer(Type,
+                                                                 IdxOwner,
+                                                                 NextOwner,
+                                                                 CStateAcc)
                           end,
                           CState0,
                           riak_core_ring:all_owners(Orig)),
@@ -590,12 +600,15 @@ compute_resize(Orig, Resized) ->
     riak_core_ring:set_pending_resize(CState1, Orig).
 
 %% @private
-schedule_first_resize_transfer({Idx, Owner}=IdxOwner, Owner, Resized) ->
-    %% index is not being moved during resize, use first predecessor
-    %% will require minor change for shrinking ring (use first successor?)
+schedule_first_resize_transfer(smaller, {Idx,_}=IdxOwner, none, Resized) ->
+    %% partition no longer exists in new ring, use first successor
+    Target = hd(riak_core_ring:preflist(<<Idx:160/integer>>, Resized)),
+    riak_core_ring:schedule_resize_transfer(Resized, IdxOwner, Target);
+schedule_first_resize_transfer(_Type,{Idx, Owner}=IdxOwner, Owner, Resized) ->
+    %% partition is not being moved during resizing, use first predecessor
     Target = hd(chash:predecessors(Idx-1, riak_core_ring:chash(Resized))),
     riak_core_ring:schedule_resize_transfer(Resized, IdxOwner, Target);
-schedule_first_resize_transfer({Idx, _Owner}=IdxOwner, NextOwner, Resized) ->
+schedule_first_resize_transfer(_,{Idx, _Owner}=IdxOwner, NextOwner, Resized) ->
     %% index is being moved, schedule this resize transfer first
     riak_core_ring:schedule_resize_transfer(Resized, IdxOwner, {Idx, NextOwner}).
 
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -27,6 +27,7 @@
          remove_member/1,
          force_replace/2,
          replace/2,
+         resize_ring/1,
          plan/0,
          commit/0,
          clear/0,
@@ -120,6 +121,12 @@ replace(Node, NewNode) ->
 %%      and does not yet own any partitions of its own.
 force_replace(Node, NewNode) ->
     stage(Node, {force_replace, NewNode}).
+
+%% @doc Stage a request to resize the ring %% TODO: moar doc
+-spec resize_ring(integer()) -> ok | {error, atom()}.
+resize_ring(NewRingSize) ->
+    %% use the node making the request. it will be ignored
+    stage(node(), {resize, NewRingSize}).
 
 %% @doc Clear the current set of staged transfers
 clear() ->
@@ -347,7 +354,9 @@ valid_request(Node, Action, Changes, Ring) ->
         {replace, NewNode} ->
             valid_replace_request(Node, NewNode, Changes, Ring);
         {force_replace, NewNode} ->
-            valid_force_replace_request(Node, NewNode, Changes, Ring)
+            valid_force_replace_request(Node, NewNode, Changes, Ring);
+        {resize, NewRingSize} ->
+            valid_resize_request(NewRingSize, Changes, Ring)
     end.
 
 %% @private
@@ -427,6 +436,17 @@ valid_force_replace_request(Node, NewNode, Changes, Ring) ->
     end.
 
 %% @private
+%% restrictions preventing resize to smaller ring and along
+%% with other operations are temporary
+valid_resize_request(NewRingSize, [], Ring) ->
+    case riak_core_ring:num_partitions(Ring) < NewRingSize of
+        true -> true;
+        false -> {error, smaller_or_equal_ring}
+    end;
+valid_resize_request(_, _Changes, _) ->
+    {error, non_empty_changes}.
+
+%% @private
 %% @doc Filter out any staged changes that are no longer valid. Changes
 %%      can become invalid based on other staged changes, or by cluster
 %%      changes that bypass the staging system.
@@ -436,6 +456,8 @@ filter_changes(Changes, Ring) ->
                    end, Changes).
 
 %% @private
+filter_changes_pred(_, {resize, _}, _, _) ->
+    true;
 filter_changes_pred(Node, {Change, NewNode}, Changes, Ring)
   when (Change == replace) or (Change == force_replace) ->
     IsMember = (riak_core_ring:member_status(Ring, Node) /= invalid),
@@ -529,13 +551,53 @@ compute_next_ring(Changes, Seed, Ring) ->
     Ring2 = apply_changes(Ring, Changes),
     {_, Ring3} = maybe_handle_joining(node(), Ring2),
     {_, Ring4} = do_claimant_quiet(node(), Ring3, Replacing, Seed),
-    Members = riak_core_ring:all_members(Ring4),
-    case riak_core_gossip:any_legacy_gossip(Ring4, Members) of
+    Ring5 = maybe_compute_resize(Ring, Ring4),
+    Members = riak_core_ring:all_members(Ring5),
+    case riak_core_gossip:any_legacy_gossip(Ring5, Members) of
         true ->
             {legacy, Ring};
         false ->
-            {ok, Ring4}
+            {ok, Ring5}
     end.
+
+%% @private
+maybe_compute_resize(Orig, MbResized) ->
+    OrigSize = riak_core_ring:num_partitions(Orig),
+    NewSize = riak_core_ring:num_partitions(MbResized),
+
+    %% restriction on expanding ring only is temprorary
+    case OrigSize < NewSize of
+        false -> MbResized;
+        true -> compute_resize(Orig, MbResized)
+    end.
+
+%% @private
+compute_resize(Orig, Resized) ->
+    %% work with the resized and balanced ring, the
+    %% original ring will be reset when changes are scheduled
+    CState0 = riak_core_ring:future_ring(Resized),
+
+    %% Each index in the original ring must perform several transfers
+    %% to properly resize the ring. The first transfer for each index
+    %% is scheduled here. Subsequent transfers are scheduled by vnode
+    CState1 = lists:foldl(fun({Idx, _} = IdxOwner, CStateAcc) ->
+                                  NextOwner = riak_core_ring:index_owner(CStateAcc, Idx),
+                                  schedule_first_resize_transfer(IdxOwner, NextOwner, CStateAcc)
+                          end,
+                          CState0,
+                          riak_core_ring:all_owners(Orig)),
+
+    riak_core_ring:set_pending_resize(CState1, Orig).
+
+%% @private
+schedule_first_resize_transfer({Idx, Owner}=IdxOwner, Owner, Resized) ->
+    %% index is not being moved during resize, use first predecessor
+    %% will require minor change for shrinking ring (use first successor?)
+    Target = hd(chash:predecessors(Idx-1, riak_core_ring:chash(Resized))),
+    riak_core_ring:schedule_resize_transfer(Resized, IdxOwner, Target);
+schedule_first_resize_transfer({Idx, _Owner}=IdxOwner, NextOwner, Resized) ->
+    %% index is being moved, schedule this resize transfer first
+    riak_core_ring:schedule_resize_transfer(Resized, IdxOwner, {Idx, NextOwner}).
 
 %% @private
 apply_changes(Ring, Changes) ->
@@ -571,7 +633,9 @@ change({{force_replace, NewNode}, Node}, Ring) ->
     Ring2 = riak_core_ring:add_member(NewNode, Ring, NewNode),
     Ring3 = riak_core_ring:change_owners(Ring2, Reassign),
     Ring4 = riak_core_ring:remove_member(Node, Ring3, Node),
-    Ring4.
+    Ring4;
+change({{resize, NewRingSize}, _Node}, Ring) ->
+    riak_core_ring:resize(Ring, NewRingSize).
 
 internal_ring_changed(Node, CState) ->
     {Changed, CState5} = do_claimant(Node, CState, fun log/2),
@@ -696,8 +760,9 @@ maybe_update_ring(Node, CState, Replacing, Seed, Log) ->
                     %% active nodes.
                     {false, CState};
                 _ ->
+                    Resizing = riak_core_ring:is_resizing(CState),
                     {Changed, CState2} =
-                        update_ring(Node, CState, Replacing, Seed, Log),
+                        update_ring(Node, CState, Replacing, Seed, Log, Resizing),
                     {Changed, CState2}
             end;
         _ ->
@@ -769,7 +834,7 @@ maybe_handle_joining(Node, Joining, CState) ->
     end.
 
 %% @private
-update_ring(CNode, CState, Replacing, Seed, Log) ->
+update_ring(CNode, CState, Replacing, Seed, Log, false) ->
     Next0 = riak_core_ring:pending_changes(CState),
 
     ?ROUT("Members: ~p~n", [riak_core_ring:members(CState, [joining, valid,
@@ -796,7 +861,8 @@ update_ring(CNode, CState, Replacing, Seed, Log) ->
     ?ROUT("Updating ring :: next2 : ~p~n",
           [riak_core_ring:pending_changes(CState4)]),
 
-    %% Rebalance the ring as necessary
+    %% Rebalance the ring as necessary. If pending changes exist ring
+    %% is not rebalanced
     Next3 = rebalance_ring(CNode, CState4),
     Log(debug,{"Pending ownership transfers: ~b~n",
                [length(riak_core_ring:pending_changes(CState4))]}),
@@ -818,6 +884,22 @@ update_ring(CNode, CState, Replacing, Seed, Log) ->
             {true, CState6};
         false ->
             {false, CState}
+    end;
+update_ring(CNode, CState, _Replacing, _Seed, _Log, true) ->
+    {Changed, CState1} = maybe_install_resized_ring(CState),
+    case Changed of
+        true ->
+            CState2 = riak_core_ring:increment_ring_version(CNode, CState1),
+            {true, CState2};
+        false ->
+            {false, CState}
+    end.
+
+maybe_install_resized_ring(CState) ->
+    case riak_core_ring:is_resize_complete(CState) of
+        true ->
+            {true, riak_core_ring:future_ring(CState)};
+        false -> {false, CState}
     end.
 
 %% @private
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -461,12 +461,14 @@ valid_resize_request(NewRingSize, [], Ring) ->
     %%            the cluster is capable. core knowing about search/kv is :(
     ControlRunning = app_helper:get_env(riak_control, enabled, false),
     SearchRunning = app_helper:get_env(riak_search, enabled, false),
-    case {ControlRunning, SearchRunning, Capable, IsResizing} of
-        {false, false, true, true} -> true;
-        {true, _, _, _} -> {error, control_running};
-        {_,  true, _, _} -> {error, search_running};
-        {_, _, false, _} -> {error, not_capable};
-        {_, _, _, false} -> {error, same_size}
+    NodeCount = length(riak_core_ring:all_members(Ring)),
+    case {ControlRunning, SearchRunning, Capable, IsResizing, NodeCount} of
+        {false, false, true, true, N} when N > 1 -> true;
+        {true, _, _, _, _} -> {error, control_running};
+        {_,  true, _, _, _} -> {error, search_running};
+        {_, _, false, _, _} -> {error, not_capable};
+        {_, _, _, false, _} -> {error, same_size};
+        {_, _, _, _, 1} -> {error, single_node}
     end.
 
 
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -453,10 +453,20 @@ valid_force_replace_request(Node, NewNode, Changes, Ring) ->
 valid_resize_request(NewRingSize, [], Ring) ->
     Capable = riak_core_capability:get({riak_core, resizable_ring}, false),
     IsResizing = riak_core_ring:num_partitions(Ring) =/= NewRingSize,
-    case {Capable, IsResizing} of
-        {true, true} -> true;
-        {false, _} -> {error, not_capable};
-        {_, false} -> {error, same_size}
+
+    %% NOTE/TODO: the checks below are a stop-gap measure to limit the changes
+    %%            made by the introduction of ring resizing. future implementation
+    %%            should allow applications to register with some flag indicating support
+    %%            for dynamic ring, if all registered applications support it
+    %%            the cluster is capable. core knowing about search/kv is :(
+    ControlRunning = app_helper:get_env(riak_control, enabled, false),
+    SearchRunning = app_helper:get_env(riak_search, enabled, false),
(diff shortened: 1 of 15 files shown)


src/riak_core_claimant.erl
```diff
@@ -436,6 +488,10 @@ filter_changes(Changes, Ring) ->
                    end, Changes).
 
 %% @private
+filter_changes_pred(_, {resize, _}, _, _) ->
+    true;
+filter_changes_pred(_, abort_resize, _, _) ->
+    true;
```
jtuple: Why do we need the two special cases for `resize` and `abort_resize` here? Is the "not invalid" catch-all case on [line 502](https://github.com/basho/riak_core/blob/7fc16113303b28b606eb68f07775e7c4a64247b6/src/riak_core_claimant.erl#L502) not work? Will we ever have invalid members involved in a transfer?

src/riak_core_claimant.erl
```diff
@@ -529,15 +585,69 @@ compute_next_ring(Changes, Seed, Ring) ->
...
+        false -> MbResized;
+        true -> compute_resize(Orig, MbResized)
+    end.
+
+%% @private
+compute_resize(Orig, Resized) ->
+    %% work with the resized and balanced ring, the
+    %% original ring will be reset when changes are scheduled
+    CState0 = riak_core_ring:future_ring(Resized),
+
```
jtuple, line 615: So, we have `Orig`, `Resized` and `CState0`. I believe this corresponds to existing/pre-resized ring, resized-but-not-yet-balanced ring, and balanced ring respectively, right? Perhaps a comment here would be helpful. Specifically, noting the difference between `Resized` and `future_ring(Resized)`.

src/riak_core_claimant.erl
```diff
@@ -529,15 +585,69 @@ compute_next_ring(Changes, Seed, Ring) ->
...
+    Target = hd(riak_core_ring:preflist(<<Idx:160/integer>>, Resized)),
+    riak_core_ring:schedule_resize_transfer(Resized, IdxOwner, Target);
+schedule_first_resize_transfer(_Type,{Idx, Owner}=IdxOwner, Owner, Resized) ->
+    %% partition is not being moved during resizing, use first predecessor
+    Target = hd(chash:predecessors(Idx-1, riak_core_ring:chash(Resized))),
+    riak_core_ring:schedule_resize_transfer(Resized, IdxOwner, Target);
+schedule_first_resize_transfer(_,{Idx, _Owner}=IdxOwner, NextOwner, Resized) ->
+    %% index is being moved, schedule this resize transfer first
+    riak_core_ring:schedule_resize_transfer(Resized, IdxOwner, {Idx, NextOwner}).
+
```
jtuple, line 658: I believe we talked about adding some comments to the `schedule_first_resize_transfer` function here about why successor vs predecessor vs moving/non-moving and how shrinking vs expanding fits into this.

src/riak_core_console.erl
```diff
@@ -447,9 +508,11 @@ output(Ring, NextRing) ->
             ok
     end,
 
-    case Next of
-        [] ->
+    case {Resizing, Next} of
+        {_, []} ->
             ok;
+        {true, _} ->
+            io:format("Ring is resizing. see riak-admin ring-status for transfer details.~n");
```
jtuple, line 518: A shame we can't print more details inline here in the output rather than punting to another command. Something to consider changing in a future release. Works for now.

src/riak_core_handoff_manager.erl
```diff
@@ -240,13 +249,13 @@ handle_cast({status_update, ModSrcTgt, StatsUpdate}, State=#state{handoffs=HS})
             {noreply, State#state{handoffs=HS2}}
     end;
 
-handle_cast({send_handoff, Mod, {Src, Target}, ReqOrigin,
+handle_cast({send_handoff, Type, Mod, {Src, Target}, ReqOrigin,
```
jtuple, line 252: As we discussed, changing the message signature here breaks repairs when there are mixed-clusters (eg. during an upgrade/downgrade). This is because [xfer/3](https://github.com/basho/riak_core/blob/7fc16113303b28b606eb68f07775e7c4a64247b6/src/riak_core_handoff_manager.erl#L107) does a direct cross-node cast that uses the signature from the old code. Could potentially fix by supporting a shim match here that just calls the other, but likely not worth it for 1.4.0.

src/riak_core_handoff_manager.erl
```diff
@@ -73,12 +74,20 @@ start_link() ->
...
 
-add_outbound(Module,Idx,Node,VnodePid) ->
+-spec add_outbound(ho_type(),atom(),integer(),term(),pid()) -> {ok, pid()} | {error, term()}.
+add_outbound(HOType,Module,Idx,Node,VnodePid) ->
+    add_outbound(HOType,Module,Idx,Idx,Node,VnodePid).
+
+-spec add_outbound(ho_type(),atom(),integer(),integer(),term(),pid()) ->
+                          {ok, pid()} | {error, term()}.
+add_outbound(HOType,Module,SrcIdx,TargetIdx,Node,VnodePid) ->
     case application:get_env(riak_core, disable_outbound_handoff) of
```
jtuple, line 84: Changing the `add_outbound` function signature breaks the use in [this eunit test](https://github.com/basho/riak_core/blob/7fc16113303b28b606eb68f07775e7c4a64247b6/src/riak_core_handoff_manager.erl#L612). Not sure if that test even matters (it's conditionally guarded after all). But, figured I'd note the issue.

src/riak_core_ring.erl
```diff
@@ -468,6 +527,85 @@ responsible_index(ChashKey, ?CHSTATE{chring=Ring}) ->
...
+            FuturePos =  ((CHashInt div NextInc) + 1),
+            NextOwner = FuturePos * NextInc,
+
+            %% Determine the partition that the key should be transferred to (has same position
+            %% in future preflist as source partition does in current preflist)
+            RingTop = trunc(math:pow(2,160)-1),
+            case NextOwner + (NextInc * OrigDist) of
+                FutureIndex when FutureIndex >= RingTop -> FutureIndex - RingTop;
+                FutureIndex -> FutureIndex
+            end
```
jtuple: Isn't this entire case statement the same as the more concise:

``` erlang
(NextOwner + (NextInc * OrigDist)) rem RingTop
```

No need to change or anything, just took me a second to realize it was just a simple wrap-around calculation.

src/riak_core_ring.erl
```diff
@@ -468,6 +527,85 @@ responsible_index(ChashKey, ?CHSTATE{chring=Ring}) ->
...
+check_invalid_future_index(OrigDist, NextCount, NValCheck) ->
+    OverRingSize = OrigDist >= NextCount,
+    OverNVal = case NValCheck of
+                   undefined -> false;
+                   _ -> OrigDist >= NValCheck
+               end,
+    OverRingSize orelse OverNVal.
+
+-spec is_future_index(chash:index(), integer(), integer(), chstate()) -> boolean().
+is_future_index(CHashKey, OrigIdx, TargetIdx, State) ->
```
jtuple, line 606: Short edoc comment would be great here.

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
 
+%% @doc Given a ring, `Resizing', that has been resized (and presumably rebalanced)
+%%      schedule a resize transition for `Orig'.
+-spec set_pending_resize(chstate(), chstate()) -> chstate().
+set_pending_resize(Resizing, Orig) ->
+    %% all existing indexes must transfer data when the ring is being resized
+    Next = [{Idx, Owner, '$resize', [], awaiting} ||
+               {Idx, Owner} <- riak_core_ring:all_owners(Orig)],
+    %% Resizing is assumed to have a modified chring, we need to put back
+    %% the original chring to not install the resized on pre-emptively. The
```
jtuple: "resized (on) pre-emptively" -> "resized (one) pre-emptively"

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+    PendingAbort = is_resize_aborted(State0),
+    case PendingAbort andalso Resizing andalso not PostResize of
+        true ->
+            State1 = State0?CHSTATE{next=[]},
+            State2 = clear_all_resize_transfers(State1),
+            State3 = remove_meta('$resized_ring_abort', State2),
+            {true, remove_meta('$resized_ring', State3)};
+        false ->
+            {false, State0}
+    end.
```
jtuple, line 891: Random style note, but using `State` rather than `State0` here would be great. I know we use `State0` when we want to do minimal changes for diffs, but since this is all new code, there's really no minimal change benefit here.

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+    State;
+schedule_resize_transfer(State, Source, Target) ->
+    Transfers = resize_transfers(State, Source),
+    case lists:keymember(Target, 1, Transfers) of
+        true -> State;
+        false ->
+            Transfers1 = lists:keystore(Target, 1, Transfers,
+                                        {Target, ordsets:new(), awaiting}),
+            set_resize_transfers(State, Source, Transfers1)
+    end.
```
jtuple, line 915: My notes say we talked about commenting this function wrt to the conditional check-before-adding logic.

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+
+-spec reschedule_resize_transfers(chstate(), [integer()], term(), term()) -> chstate().
+reschedule_resize_transfers(State=?CHSTATE{next=Next}, Node, NewNode) ->
```
jtuple, line 919: This entire `reschedule_resize_transfers` stuff is dense and confusing. We talked about it a bit and about adding some comments + possibly rewriting to make cleaner. Especially the distinction between rescheduling requests originating FROM the node that is being replaced vs rescheduling transfers being sent TO the the node being replaced.

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+                                      {integer(), term()} | undefined.
+awaiting_resize_transfer(State, Source, Mod) ->
+    ResizeTransfers = resize_transfers(State, Source),
+    Awaiting = [{Target, Mods, Status} || {Target, Mods, Status} <- ResizeTransfers,
+                                          Status =/= complete,
+                                          not ordsets:is_element(Mod, Mods)],
+    case Awaiting of
+        [] -> undefined;
+        [{Target, _, _} | _] -> Target
+    end.
```
jtuple, line 979: No big deal for now, but in the future maybe add early exit here. Ie. stop after first `awaiting` transfer found rather than iterating over the entire list and returning the first result. Doesn't really matter, but adding a PR comment is cheap. So, here ya go.

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+
+%% @doc return the status of a resize_transfer for `Source' (a index-node pair). undefined
```
jtuple: "(a) index" -> "(an) index"

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+%%      is returned if no such transfer is scheduled. complete is returned if the transfer
+%%      is marked as such or `Mod' is contained in the completed modules set. awaiting is
+%%      returned otherwise
+-spec resize_transfer_status(chstate(), {integer(), term()}, {integer(), term()}, atom()) ->
+                                    awaiting | complete | undefined.
+resize_transfer_status(State, Source, Target, Mod) ->
+    ResizeTransfers = resize_transfers(State, Source),
+    %% change to use keyfind
+    StatusMods = [{Status, Mods} || {T, Mods, Status} <- ResizeTransfers,
+                                    T =:= Target],
```
jtuple: Reminder that you have a note here to change to using `keyfind`. Perhaps after 1.4?

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+    IsComplete = case StatusMods of
+                     [{complete, _}] -> true;
+                     [{awaiting, Mods}] -> ordsets:is_element(Mod, Mods);
+                     _ -> undefined
+                 end,
```
jtuple, line 993: Are we confidence this will always return a single element list in normal operation? Any weird corner case?

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+    [Target || {Target, Mods, Status} <- resize_transfers(State, Source),
+               Status =:= complete orelse ordsets:is_element(Mod, Mods)].
+
+-spec deletion_complete(chstate(), integer(), atom()) -> chstate().
+deletion_complete(State, Idx, Mod) ->
+    transfer_complete(State, Idx, Mod).
+
+resize_transfers(State, Source) ->
+    {ok, Transfers} = get_meta({resize, Source}, [], State),
+    Transfers.
```
jtuple, line 1077: Add a `-spec` to `resize_transfers/2` to make dialyzer more useful, since metadata queries always return to `term()`.

src/riak_core_ring.erl
```diff
@@ -701,6 +860,263 @@ pending_changes(State) ->
...
+    update_meta('$resized_ring', FutureCHash, State).
+
+cleanup_after_resize(State) ->
+    update_meta('$resized_ring', '$cleanup', State).
+
+
+-spec vnode_type(chstate(),integer()) -> primary |
+                                         {fallback, term()} |
+                                         new_primary |
+                                         resized_primary.
```
jtuple, line 1110: Think spec might be off, don't think `new_primary` is something that we can ever return here.

src/riak_core_claimant.erl
jrwest: we don't need them. artifact of some original work when I was less familiar w/ the claimant. will remove.

src/riak_core_ring_handler.erl
```diff
@@ -36,6 +36,8 @@ init([]) ->
 handle_event({ring_update, Ring}, State) ->
     %% Make sure all vnodes are started...
     ensure_vnodes_started(Ring),
+    maybe_start_vnode_proxies(Ring),
+    maybe_stop_vnode_proxies(Ring),
```
jtuple, line 40: Any implications for normal Riak operation (ie. users that never use resizing) by adding this start/stop proxy code? There's nothing I could find after looking things over, but figured it was worth discussing.

src/riak_core_vnode.erl
```diff
@@ -103,8 +105,9 @@ behaviour_info(_Other) ->
           mod :: module(),
           modstate :: term(),
           forward :: node(),
```
jtuple: `forward` can now be a list as well

src/riak_core_vnode.erl
```diff
@@ -103,8 +105,9 @@ behaviour_info(_Other) ->
...
-          handoff_node=none :: none | node(),
+          handoff_target=none :: none | {integer(), node()},
           handoff_pid :: pid(),
+          handoff_type :: hinted_handoff | ownership_transfer | resize_transfer,
```
jtuple, line 110: So, there's no `repair` handoff type. Looking [here](https://github.com/basho/riak_core/blob/7fc16113303b28b606eb68f07775e7c4a64247b6/src/riak_core_vnode.erl#L864-L868) it looks like repair is flagged as a `ownership_transfer`. I guess I should test this. In any case, do we care about repair being a type?

src/riak_core_claimant.erl
jrwest, line 615: yes that is correct. will add a function comment with a bit more detail about the arguments and a body comment about the use of `future_ring(Resized)`

src/riak_core_vnode.erl
```diff
@@ -241,23 +250,53 @@ continue(State, NewModState) ->
...
+%% the ring, since several transfers with a single vnode as the source are necessary
+%% to complete the operation. A vnode will remain in the handoff state, for, potentially,
+%% more than one transfer and may be in the handoff state despite there being no active
+%% transfers with this vnode as the source. During this time requests that can be forwarded
+%% to a partition for which the transfer has already completed, are forwarded. All other
+%% requests are passed to handle_handoff_command.
+forward_or_vnode_command(Sender, Request, State=#state{forward=Forward,
+                                                       mod=Mod,
+                                                       index=Index}) ->
+    Resizing = is_list(Forward),
```
jtuple, line 264: The majority of `forward_or_vnode_command/3` is more or less the same as code in `active`. Compare [lines 264-292](https://github.com/basho/riak_core/blob/7fc16113303b28b606eb68f07775e7c4a64247b6/src/riak_core_vnode.erl#L264-L292) vs [lines 414-437](https://github.com/basho/riak_core/blob/7fc16113303b28b606eb68f07775e7c4a64247b6/src/riak_core_vnode.erl#L414-L437). Any way to refactor out and combine this logic into one place? Bonus points for it being a pure function that we can easily trace-on, run unit/EQC tests against, etc.

src/riak_core_vnode.erl
```diff
@@ -241,23 +250,53 @@ continue(State, NewModState) ->
...
+        %% implicit forwarding after ownership_transfer/hinted_handoff
+        {F, _} when not is_list(F) -> vnode_forward({Index, Forward}, Sender, Request, State);
+        %% during resize we can't forward a request w/o request hash, always handle locally
+        {_, undefined} -> vnode_command(Sender, Request, State);
+        %% possible forwarding during ring resizing
+        {_, _} ->
+            %% during ring resizing if we have completed a transfer to the index that will
+            %% handle request in future ring we forward to it. Otherwise we delegate
+            %% to the local vnode like other requests during handoff
+            %% TODO: how to handle all resize ops completed case (need an extra check here)
```
jtuple: We've talked about the above TODO and punting most likely. Just noting it officially in this PR. We did mention possibly modifying the TODO to be more clear about the specific problem: eg. N-value change during resize.

src/riak_core_vnode.erl
```diff
@@ -241,23 +250,53 @@ continue(State, NewModState) ->
...
+                {FutureIndex, FutureOwner} -> vnode_handoff_command(Sender, Request,
+                                                                    {FutureIndex, FutureOwner},
+                                                                    State)
+            end
+    end.
 
-vnode_command(Sender, Request, State=#state{index=Index,
-                                            mod=Mod,
+vnode_command(_Sender, _Request, State=#state{modstate={deleted,_}}) ->
+    continue(State);
```
jtuple, line 287: This "do nothing when deleted" case is new. Why do we need it? Does it impact anything else in Riak?

src/riak_core_vnode.erl
```diff
@@ -333,6 +385,13 @@ vnode_handoff_command(Sender, Request, State=#state{index=Index,
             {stop, Reason, State#state{modstate=NewModState}}
     end.
 
+vnode_forward(ForwardTo, Sender, Request, State) ->
+    lager:debug("Forwarding {~p,~p} -> ~p~n",
+                [State#state.index, node(), ForwardTo]),
+    riak_core_vnode_master:command(ForwardTo, Request, Sender,
+                                   riak_core_vnode_master:reg_name(State#state.mod)),
+    continue(State).
```
jtuple, line 385: Since `vnode_forward/4` now provides a single point for all forwarding -- traditional explicit, traditional implicit, and resize -- it seems like it would make sense to have a "forward type" parameter passed into this function that we could trace on if needed as well add to the `lager:debug` output. Thoughts?

src/riak_core_console.erl
jrwest, line 518: agreed. i originally played with printing the resize transfers but since we don't know all of them up front it was kind of funky. Printing that every partition will do a resize operation wasn't very helpful either. hoping further use will highlight what will be good to print here.

src/riak_core_vnode.erl
```diff
@@ -341,14 +400,57 @@ active(?COVERAGE_REQ{keyspaces=KeySpaces,
                      sender=Sender}, State) ->
     %% Coverage request handled in handoff and non-handoff.  Will be forwarded if set.
     vnode_coverage(Sender, Request, KeySpaces, State);
-active(?VNODE_REQ{sender=Sender, request=Request},
-       State=#state{handoff_node=HN}) when HN =:= none ->
+active(?VNODE_REQ{sender=Sender, request={resize_forward, Request}}, State) ->
     vnode_command(Sender, Request, State);
```
jtuple, line 409: So, what's the purpose behind the `resize_forward` prefix rather than just forwarding the request directly (like non-repair forwarding does)? Is it to prevent multiple forwarding hops, eg: `forward A -> forward B -> forward C`? Is there an existing bug in normal Riak we should be cared about (that still just forwards requests w/o a prefix)?

src/riak_core_console.erl
jrwest, line 518: also to be fair, ring status could have more useful information too. Right now its at the resize operation (next list) level not resize transfer level. more visibility w/o needing to `riak attach` will be something I want to focus on for next release.

src/riak_core_vnode.erl
```diff
@@ -356,14 +458,27 @@ active({send_manager_event, Event}, State) ->
...
+active(trigger_delete, State=#state{mod=Mod,modstate=ModState,index=Idx}) ->
+    case mark_delete_complete(Idx, Mod) of
+        {ok, _NewRing} ->
+            {ok, NewModState} = Mod:delete(ModState),
+            lager:debug("~p ~p vnode deleted", [Idx, Mod]);
+        _ -> NewModState = ModState
+    end,
+    maybe_shutdown_pool(State),
+    riak_core_vnode_manager:unregister_vnode(Idx, Mod),
+    continue(State#state{modstate={deleted,NewModState}});
```
jtuple, line 471: We talked about this a bit. Why the conditional delete but not conditional shutdown? Is this needed? Under what conditions? Do we care? Ie. not changing code is safer in case we forgot about some weird corner case that caused you to implement this in the first place.

src/riak_core_vnode.erl
```diff
@@ -449,10 +599,46 @@ finish_handoff(State=#state{mod=Mod,
...
+                       {_, Next, Status} = riak_core_ring:next_owner(Ring, Idx),
+                       case {Type, Next, Status} of
+                           {resized_primary, '$delete', awaiting} ->
+                               Ring3 = riak_core_ring:deletion_complete(Ring, Idx, Mod),
+                               %% Use local ring optimization like mark_handoff_complete
+                               {set_only, Ring3};
+                           {{fallback, _}, '$delete', awaiting} ->
+                               Ring3 = riak_core_ring:deletion_complete(Ring, Idx, Mod),
+                               %% Use local ring optimization like mark_handoff_complete
+                               {set_only, Ring3};
```
jtuple, line 632: Curious why `fallbacks` would have data to delete. I imagine there's a good reason, might as well throw a comment in. Annoying we have duplicate code in both case clauses, but there's plenty of bigger fish to fry.

src/riak_core_vnode_manager.erl
```diff
@@ -615,11 +620,49 @@ update_handoff(AllVNodes, Ring, State) ->
 
 should_handoff(Ring, Mod, Idx) ->
     {_, NextOwner, _} = riak_core_ring:next_owner(Ring, Idx),
-    Owner = riak_core_ring:index_owner(Ring, Idx),
+    Type = riak_core_ring:vnode_type(Ring, Idx),
     Ready = riak_core_ring:ring_ready(Ring),
-    case determine_handoff_target(Ready, Owner, NextOwner) of
+    IsResizing = riak_core_ring:is_resizing(Ring),
+    Me = node(),
+    case {Type, NextOwner, Ready, IsResizing} of
```
jtuple: Suggest breaking out this case statement into a function similar to what `determine_handoff` used to be. Makes it possible to trace if needed + easy to write eunit tests against if we ever care.

jtuple:
Overall comment about using `try/catch` in tons of places, many of which catch directly on `error:badarg`. Think the approach works for now, but it would be great if we could change this in a future revision. I'm mainly worried about errors from things other than we expect (esp. as code evolves) as well as generating errors that don't match the specific `error:badarg` type crashing.

src/riak_core_ring.erl
jrwest, line 891: will change. wasn't aware we have a preference. I can never decide so I tend to go back and forth depending on nearby code. Now its easy :).

src/riak_core_ring.erl
jrwest: tentatively changed it to a real `TODO` at least. will try to just fix it up after other changes but if not at least its more than a straggling comment.

src/riak_core_ring.erl
jrwest, line 993: should always be a single element b/c of the check that you mentioned needs a comment in `schedule_resize_transfer` [1]. Originally was going to say, lets just change it to pull the head of the list but actually i think I'd prefer it crash because if we have multiple resize transfers scheduled for the same target we a have a problem and this may be way less subtle than some other funky things i can imagine happening. Also, thats where the keyfind comment originated from fwiw (not really smart use of LC here on my part imo) 

[1] https://github.com/basho/riak_core/pull/301#discussion_r4371567

src/riak_core_ring.erl
jrwest, line 1110: heh, wrote the spec before the function then realized `new_primary` made no sense and forgot to remove. will do so.

jrwest, line 1110: ah wait, i see the problem. `new_primary` should be `future_primary`. sorry, commented before looking at the function beforehand.

src/riak_core_ring_handler.erl
jrwest, line 40: I couldn't think of anything, it should basically be some extra useless work that has no effects. I don't love this but I figure it will be better once its merged with the changes in #319.

src/riak_core_vnode.erl
jrwest, line 110: hrm, i was under the impression this was not the code path taken by repair which is why I don't have the type for the vnode state but i do have it [here](https://github.com/basho/riak_core/blob/7fc16113303b28b606eb68f07775e7c4a64247b6/include/riak_core_handoff.hrl#L16). iirc since repair operates on primaries it just sends the fold request and the primaries continue to call `handle_command` instead of doing the `handle_handoff_command` and forwarding dance. I remember having a discussion w/ @rzezeski about this briefly a while back but I could have totally misunderstood what he told me, need to look at the code as well.

src/riak_core_vnode.erl
jrwest, line 287: This just prevents some annoying log messages from occurring during the cleanup stage of resizing when vnodes are being shut down. [This commit](https://github.com/basho/riak_core/commit/134e3ac9af6f76904e1a1580751445c59e8d8509) has more details (the messages mentioned are things like `merge_callback` that fired while the shutdown messages were being sent). If you remove it and run the `riak_test` it should still pass but you may see some (harmless) `riak_kv_vnode` crashes because `{deleted, ModState}` is passed to `handle_command`.

src/riak_core_vnode.erl
jrwest, line 385: agreed. i have to look at the call sites but i think it would be easy to add. will do.

src/riak_core_vnode.erl
jrwest, line 409: yep it prevents that case exactly. there is not an existing bug in riak because only primaries forward and we only forward to non-primaries. With resizing, especially shrinking, we commonly forward from primary to primary.

jrwest, line 409: but...perhaps adding something similar for normal forwarding could be useful for tracing?

src/riak_core_vnode.erl
jrwest, line 632: [this comment in the vnode manager](https://github.com/basho/riak_core/blob/7fc16113303b28b606eb68f07775e7c4a64247b6/src/riak_core_vnode_manager.erl#L640-L642) explains why, i can copy it over. agree the duplication is unfortunate, perhaps more an issue with the return values of `riak_core_ring:vnode_type` which make it difficult to pattern match on multiple at once if fallback is one of them?

src/riak_core_vnode_manager.erl
jrwest: i actually took `determine_handoff` out because I found it weird to follow the code at the time. in hindsight i think you are correct. will change.

jrwest:
yea not a fan of the `try/catch` either. the reasoning behind it was to change as little of the existing semantics as possible. In a future version I'd like to move the error handling into the respective functions.

src/riak_core_handoff_manager.erl
jrwest, line 84: changed the test so if you add the define it compiles but the test fails anyways because it requires `riak_core_handoff_receiver/sender_sup`:

```
ring-resizing/riak_core $ ./rebar skip_deps=true eunit suite=riak_core_handoff_manager
==> riak_core (eunit)
src/riak_core_handoff_manager.erl:612: function add_outbound/4 undefined
ring-resizing/riak_core $ ./rebar skip_deps=true eunit suite=riak_core_handoff_manager
==> riak_core (eunit)
Compiled src/riak_core_handoff_manager.erl
riak_core_handoff_manager:603: handoff_test_...*skipped*
module 'riak_core_handoff_manager'
*unexpected termination of test process*
::{noproc,{gen_server,call,
                      [riak_core_handoff_receiver_sup,count_children,
                       infinity]}}
....
```

src/riak_core_ring.erl
jrwest: it should be what you have above. I actually noticed that when I was reviewing large ring work but forgot to change. Went ahead and made the change since `riak_core_ring_eqc:prop_future_index()` still passes.

src/riak_core_ring.erl
jrwest, line 919: while addressing this, noticed the spec above is for the totally wrong function (facepalm).

src/riak_core_vnode.erl
jrwest, line 471: so looking at the code again, i believe this was written to be defensible to "unpredictable" corner cases. The `Mod:delete` is not run if either the operation has already been marked as complete (in which case running it again would be fine) or if the vnode type is not what we expected (in which case we may have hit some weird case where we shouldn't delete the data now). Initially i thought to get rid of the case clause but I think now it should remain. I think its still safe to shutdown the pool and such because we will unregister the vnode and it can be restarted if needed but at least the data is still there. alternatively, could differentiate between the wrong vnode type case and the already marked complete case to be extra clear/safe.

src/riak_core_ring.erl
jrwest, line 979: agreed. going to leave it for now since this is always going to be a small list (not ring size dependent and always much smaller than ring size) but have a note to change along w/ future work.

jtuple:
Latest commits address my review concerns. Re-ran `riak_test`, manually played with code both before and after latest commits, etc. Going to +1 this for merge for 1.4 with the knowledge that we're shipping this feature "dark".

Overall, resizing works as intended. Both scaling up and down in ring size. Tested various resize operations while loading a cluster, shutting nodes down, marking nodes down, etc. All worked without any major issue.

There are a few minor issues that would be great to fix for a future release or at least are useful to note for the time being:
1. If a node is offline and marked as down, Riak will keep trying to initiate resize transfers to that node and immediately fail. This will spam the logs over and over/etc. For normal Riak handoff, marking a node as down forces the claimant to remove transfers scheduled to said node, and therefore we never attempt to handoff to said node. Thus, the behavior is different for resizing. This is a fact of life given that resize doesn't use the claimant for scheduling transfers after the first. However, we can improve the situation in the future by looking at node watcher and not attempting handoff to nodes that are offline.
2. A known limitation of this resize work is that we only delete entire vnodes. If a resize causes some (by not all) data owned by a vnode to be rehomed to a new vnode, we don't delete the data on the original vnode. This was a limitation we decided to allow for this iteration of ring resizing. However, a related problem is that the AAE trees for vnodes will still include hashes for the old data. Exchanges with siblings that correctly don't have that data will trigger repairs for keys that shouldn't be owned. These repairs won't actually trigger new writes, however, and will actually cause the AAE tree to realize it shouldn't own said key. However, this will trigger additional, unnecessary read repairs for a period of time. An alternative approach is to clear all AAE trees after a resize, but then that would require rebuilding all trees -- potentially more expensive than extra read repairs. A smarter approach that actually cleans up moved data both in the K/V backend and the AAE hashtree should be a top priority for the next major release.
3. Resizing with Bitcask while there are incoming writes is rather painful. More so if performing other operations that perform folds (such as AAE tree building or full-sync replication). When there are incoming writes, Bitcask only allows a single fold to occur at a time. For normal Riak, there's only ever a single outbound handoff for a given vnode. However, with resize, it's possible for a vnode to have multiple outbound handoffs running concurrently to transfer different portions of it's keyspace to different target vnodes. Bitcask will allow only 1 to happen a time, all others will block. Resize therefore takes much longer to complete while also (in practice) preventing any other folds from occurring during the entire resize operation.
4. I'd love to see better insight/observability into how far along we are on a resize operation. `member_status` is completely useless, `ring_status` provides only coarse-grained details (eg. have no idea how many sub-transfers are outstanding for each shown awaiting index).

jrwest:
merged w/ #329

PR #302, opened 2013-04-18
rework coverage fsm timeouts.

evanmcc:
Current coverage fsm timeouts arent' consistent with the way other timeouts in riak work.  This change makes it so they're bounded by the total time given by the init functions, rather than applying their timeout to their waiting phases, which could be unbounded if there are multiple calls to waiting_results.


seancribbs:
I'm generally :+1: but there should be a review from another person, e.g. @jtuple @jrwest @engelsanchez @rzezeski

kellymclaughlin:
+1 to merge

PR #307, opened 2013-04-30
Backport issue 300 to 1.3 branch

beerriot:
It was desired to have the riak_core_vnode_worker_pool race from #300 backported to the 1.3 branch. The parent commit containing the regression tests for this race has been elided, since it requires changes to poolboy that are not otherwise required to fix the issue (so there is no need to roll a new poolboy version).

To run the worker pool test against this branch, also cherry-pick commit 7a60b6e onto this branch. Then follow the instructions in #300 (poolboy master now has PULSE annotations).


evanmcc:
all of the related riak_tests finish fine, code inspection looks clean, after some drama, the pulse tests pass.

Issue #309, opened 2013-05-02
Fix riak_core_connection_mgr_stats timestamp formatting


bookshelfdave:
riak-repl clusterstats
RPC to 'riak@brawl' failed: {'EXIT',
                                                           {function_clause,
                                                            [{riak_core_connection_mgr_stats,
                                                              format_stat,
                                                              [{riak_conn_mgr_stats_stat_ts,
                                                                1365529127}],
                                                              [{file,
                                                                "src/riak_core_connection_mgr_stats.erl"},
                                                               {line,61}]},
                                                             {riak_core_connection_mgr_stats,
                                                              '-get_consolidated_stats/0-lc$^0/1-0-',
                                                              1,
                                                              [{file,
                                                                "src/riak_core_connection_mgr_stats.erl"},
                                                               {line,54}]},
                                                             {riak_core_connection_mgr_stats,
                                                              get_consolidated_stats,
                                                              0,
                                                              [{file,
                                                                "src/riak_core_connection_mgr_stats.erl"},
                                                               {line,54}]},
                                                             {riak_repl_console,
                                                              clusterstats,1,
                                                              [{file,
                                                                "src/riak_repl_console.erl"},
                                                               {line,203}]},
                                                             {rpc,
                                                              '-handle_call_call/6-fun-0-',
                                                              5,
                                                              [{file,
                                                                "rpc.erl"},
                                                               {line,203}]}]}}

bookshelfdave:
see https://github.com/basho/riak_core/pull/310

slfritchie:
Time for some belated repo curating ... Dave, it seems to my bleary eyes that this ticket can be closed?

bookshelfdave:
thanks @slfritchie

PR #310, opened 2013-05-02
render _stat_ts without blowing up

bookshelfdave:
also, try to catch other stats that may have slipped in

You can see the output in an Advanced repl cluster via
`riak-repl clusterstats`

cc @aggress


bookshelfdave:
see https://github.com/basho/riak_core/issues/309

bookshelfdave:
on hold, I have to fix the format of the stat.

PR #314, opened 2013-05-04
Implement vnode overload protection

jtuple:
This pull-request provides overload protection for vnodes, limiting the size of a given vnode's mailbox and dropping message if the limit is reached. The overload logic is implemented by the vnode proxy, which decides to either forward the message to the relevant vnode or drop the message -- potentially informing a waiting `riak_client` about the overload condition.

Every time the proxy process sends a message to the vnode, it increments a counter that estimates the size of the vnode's mailbox. If the counter hits the limit, messages are dropped. This limit is configured via `riak_core/vnode_overload_threshold` setting. This estimated mailbox size is periodically adjusted to make it match the actual mailbox size.

After sending a configurable number of messages (determined by `riak_core/vnode_check_request_interval`), the proxy will send a ping message to the vnode. Whenever a pong arrives, the proxy knows all messages sent before the ping have been handled and can adjust the estimate accordingly. In the common case, this ping/pong approach keeps the mailbox estimate up-to-date without ever falling back to the more expensive approach described below.

After sending even more messages (determined by `riak_core/vnode_check_interval`), the proxy will directly query the size of the vnode's mailbox using `process_info`. This call is potentially expensive. If the vnode is currently running, the proxy will be suspended until the vnode is descheduled.

This pull-request also provides the `dropped_vnode_requests_total` stat, which is incremented anytime a request is dropped.

I will edit this pull-request with a link to the relevant riak_test for this work, which also tests the the related FSM overload protection feature. As well as add benchmark results.

This pull-request requires related basho/riak_kv#547

### DIFF
--- a/src/riak_core_stat.erl
+++ b/src/riak_core_stat.erl
@@ -106,6 +106,9 @@ update1(gossip_received) ->
 update1(rings_reconciled) ->
     folsom_metrics:notify_existing_metric({?APP, rings_reconciled}, 1, spiral);
 
+update1(dropped_vnode_requests) ->
+    folsom_metrics:notify_existing_metric({?APP, dropped_vnode_requests_total}, {inc, 1}, counter);
+
 update1(converge_timer_begin) ->
     folsom_metrics:notify_existing_metric({?APP, converge_delay}, timer_start, duration);
 update1(converge_timer_end) ->
@@ -123,6 +126,7 @@ stats() ->
      {gossip_received, spiral},
      {rejected_handoffs, counter},
      {handoff_timeouts, counter},
+     {dropped_vnode_requests_total, counter},
      {converge_delay, duration},
      {rebalance_delay, duration}].
 
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -514,6 +514,9 @@ handle_sync_event(core_status, _From, StateName, State=#state{index=Index,
         end,
     {reply, {Mode, Status}, StateName, State, State#state.inactivity_timeout}.
 
+handle_info({'$vnode_proxy_ping', From, Msgs}, StateName, State) ->
+    riak_core_vnode_proxy:cast(From, {vnode_proxy_pong, self(), Msgs}),
+    {next_state, StateName, State, State#state.inactivity_timeout};
 
 handle_info({'EXIT', Pid, Reason},
             _StateName,
--- a/src/riak_core_vnode_proxy.erl
+++ b/src/riak_core_vnode_proxy.erl
@@ -21,14 +21,26 @@
          unregister_vnode/3, command_return_vnode/2]).
 -export([system_continue/3, system_terminate/4, system_code_change/4]).
 
--record(state, {mod, index, vnode_pid, vnode_mref, check_interval,
-                check_threshold, check_counter=0, check_mailbox=0}).
-
 -include("riak_core_vnode.hrl").
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 -endif.
+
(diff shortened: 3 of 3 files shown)


engelsanchez:
I have a general question about the approach.  So, with default values, we'll start dropping messages once we get to the point were we have forwarded 5k messages without hearing a pong back (ping sent at 2.5k messages).  Once we reach this state, we ignore the next pong.  The next event that may update the mailbox to be under the threshold is reaching another 2.5k of incoming messages, sending the ping and either receiving it or reach the 5k incoming (dropped) messages again and querying directory.  During that time, if for any reasons we don't receive enough messages to trigger that, the vnode could be in a state where its mailbox has already gone under the 10k threshold, but the proxy will not get an update on that for an unknown amount of time.  Could you guys check my reasoning?

If correct, then maybe we need another way to bound the amount of time we spend in that "it could handle the messages, but unfortunately we don't know".  Perhaps send a single ping to the vnode when we know it has > 10k messages, so that as soon as all the messages are cleared the proxy finds out ASAP?

Another option (again, only needed if I'm correct above): is querying your own msg queue length as bad as querying it from another process? If not, would it make sense to change the pong response to return the at-that-moment msg queue length, so that the proxy can read that number + msgs sent since the ping to get an upper bound on the vnode's real msg queue length and use that?

Thoughts?

/cc @jtuple @Vagabond @jonmeredith

jtuple:
We send a ping after 2.5k messages. Any time the pong comes back, we adjust our mailbox. We don't ignore it. The only time we set the flag to ignore the pong is _after_ we've already done a hard query of the mailbox to directly reset it based on the actual current value.

So, the vnode mailbox eventually drains to the ping message and sends the pong response back adjusting the mailbox, or it doesn't get to it in time and we do the hard query.

As for the second option. Querying local mailbox count is very, very cheap. But, that doesn't help us here. If the vnode queries it's mailbox at T0, sends the response to the proxy where it arrives at time T1, but in between T0 and T1 the proxy sends another 500 messages to the vnode, then the mailbox count the proxy sees at T1 is off by 500 messages. This was one of my initial implementations of this approach, until I realized that problem.

This is why the ping/pong approach operates as an adjustment rather than an overwrite. Proxy sends 2500 messages, sends {ping, 2500}, sends another 1000 messages, sees {pong, 2500}. At this moment, the proxy's estimate is 3500, the pong claims at minimum 2500 messages had to have been handled, so we subtract that to get 1000 message estimate. If the vnode proxy had send 10, 1500, etc messages in between the ping/pong, we still adjust correctly.

Of course, once we do a hard reset of the estimate based on process_info, we can't adjust the current mailbox estimate anymore because that number may not include all the messages that a given ping/pong represent. Thus, we set the ignore flag.

The original implementation here didn't have this ping/pong logic. We always did a hard query after `check_interval` messages. This means that in the worst case, we can always drop up to `check_interval` messages needlessly. It also meant that we always did a hard query even when not overloaded. The ping/pong approach still has the same theoretical worse case bounds, but avoids this worse case in practice. It also only ever does the hard query if pong messages aren't arriving fast enough, which only happens in overload scenarios.

In the end, we're aiming at a conservative estimate. We'd rather drop some messages that we could of been handled as long as we don't ever send a message that is over the limit.

engelsanchez:
This code has been reviewed and tested along with the other overload protection PRs. It seems to work as expected. Joe and I have discussed some ways to simplify and protect from theoretical worst cases when a vnode enters the overload state and what it takes for it to leave that space. Specifically, the fact that it needs a number of incoming messages before it will take it out of that state. Improvements should follow, but the situation is very unlikely to really cause problems in our opinion.  So :+1: :dancer: :boat:

Issue #320, opened 2013-05-16
riak-admin transfers exits with status 1 when there are no transfers.


fprimex:
```
$./bin/riak start
$./bin/riak-admin transfers
No transfers active

Active Transfers:


$echo $?
1
```

PR #321, opened 2013-05-16
proper return value for riak_core_console:transfers/1

jrwest:
addresses #320.


evanmcc:
+1, as far as I can tell nothing consumes the output of that function other than nodetool.

PR #323, opened 2013-05-20
Permanently disable legacy gossip

jtuple:
This commit hardcodes all checks for legacy gossip to return false. This is the first step towards removing legacy gossip from riak_core entirely. A future commit will completely remove all relevant code as well as adjust the callsites that currently still call into these now hardcoded functions.

### DIFF
--- a/src/riak_core.erl
+++ b/src/riak_core.erl
@@ -81,7 +81,7 @@ join(false, _, Node, Rejoin, Auto) ->
         pang ->
             {error, not_reachable};
         pong ->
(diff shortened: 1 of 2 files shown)


rzezeski:
+1 to merge
- Code compiles, eunit passes, EQC passes
- The riak test basic-upgrade and staged-clustering pass
- Lightly ran through areas that call legacy gossip, couldn't find anything that might be harmful.
- Discussed PR with @jtuple for 30 minutes, as long as someone doesn't mix a 0.14.x and 1.x node in the same cluster we are fine.

PR #325, opened 2013-05-22
Fix a typo in tcp_mon init

Vagabond:


bookshelfdave:
+1, eunit passes, EE starts

PR #327, opened 2013-05-26
Coverage plan nif

zinid:
I have partially rewritten find_coverage in C using bitarrays (bitmasks), as the original version is slow as hell and consumes 50% of time according to fprof. It is now not a bottle-neck anymore and a node can handle significantly more workload.

NOTE: the code is improved for ring_creation_size = 64 only.

### DIFF
--- a/c_src/riak_core_coverage_plan.c
+++ b/c_src/riak_core_coverage_plan.c
@@ -12,6 +12,7 @@
 struct state {
   unsigned int offset;
   unsigned int nval;
+  uint64_t coverage[BASE];
 };
 
 typedef unsigned char vnode_t;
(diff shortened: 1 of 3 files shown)


evanmcc:
Since this isn't general (that is, it's limited to a fixed ring size) and we're trying to avoid NIFs more generally, we aren't going to accept this.  Recently, there have been some Erlang-side optimizations to this which should make it less of a bottleneck in your case.  You may want to rerun your benchmarks against a post 1.4 riak.

Thank you for your efforts.

zinid:
I rerun them actually and still have the same picture. It's really too bad you have strange policy of avoiding NIFs as the target language should be choosen by the task.
It is also quite easy to re-write it for any size rings using bignum libraries for example.

PR #328, opened 2013-05-28
Fix overload test time outs

engelsanchez:
Testing slow vnodes tends to time out, so upping the eunit timeout
values.
The gen:call was timing out with its 5 second default instead of obeying
the eunit test timeout, so it was changed to infinity.


russelldb:
Looks sane, test passes for me from the console, weird meck error is I run just that test from rebar, which I'm digging into. But I get that on master too, so +1

russelldb:
Oh, finally got it running. Who knows what was up before. +1 all round!

PR #330, opened 2013-05-30
dont start coverage timeout timer if timeout is infinite

jrwest:
When the timeout is `infinity` the coverage fsm crashes w/o this change. To verify run `mapred_verify_rt` with `riak_test` w/ and w/o this change (or any mapreduce request that uses an `infinity` timeout, e.g. basho_bench)


evanmcc:
:+1: Change looks good, it fixes mapred_verify_rt and doesn't break verify_api_timeouts or verify_listkeys

PR #331, opened 2013-06-03
fix forced_ownership_handoff during resize

jrwest:
All resize operations remain in the ring's list of pending
changes until all complete. Prior to this change transfers would
only be triggered for the first forced_ownership_handoff operations.
Subsequent operations would only be triggered by vnode _inactivity_.

This commit modifies the use of forced_ownership_handoff during resize
to ensure that only resize operations that are still pending are in
the throttled transfer list.

This addresses the feature blocking issue in the second to last paragraph [here](https://github.com/basho/riak_pipe/pull/76#issuecomment-18715817)

<hr />

This PR can be verified using basho_bench w/ configs [here](https://gist.github.com/jrwest/e534ebc60f6a91456d4b): 
1. Build a 64-partition, 4-node cluster and run `mapred_populate.config`. verify everything is ok w/ `mapred_verify.config`.
2. Kick off a resize with `riak-admin cluster resize-ring 128`, then `plan` and `commit`.
3. Run `mapred_verify.config` over and over w/ something like `while 1; do ./basho_bench mapred_verify.config; done`.

w/o this change during 3. the resize will stall (no transfer output in logs is an easy way to verify this, also `ring-status` will not progress). w/ this change, you will see the resize progress despite the mapreduce traffic.

Note: mapreduce traffic is not necessary, any sufficient load that prevents all vnodes from reaching their inactivity timeout will do

### DIFF
--- a/src/riak_core_vnode_manager.erl
+++ b/src/riak_core_vnode_manager.erl
@@ -421,7 +421,7 @@ handle_info(management_tick, State0) ->
             _ ->
                 Repairs = State#state.repairs,
                 kill_repairs(Repairs, ownership_change),
-                trigger_ownership_handoff(Transfers, Mods, State2),
+                trigger_ownership_handoff(Transfers, Mods, Ring, State2),
                 State2#state{repairs=[]}
         end,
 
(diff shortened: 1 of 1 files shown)


jrwest:
forgot to mention, when running `mapred_verify.config` after the resize completes, you will see errors from basho_bench because the expected key counts will be wrong depending on the resize operation. For more information see the last paragraph [here](https://github.com/basho/riak_pipe/pull/76#issuecomment-18715817). This is expected behaviour for now and is not the subject of this PR.

jtuple:
Reviewed code, nothing of consequence to note, all looks good there.

Retested against master to verify transfer stall. Then tested again using this branch merged to master, and no stall occurred. Awesomeness.

+1 merge away

PR #332, opened 2013-06-03
update bad value protection for timer value

evanmcc:
Since there are many paths to spawn a coverage FSM, it would be a lot of code duplication to validate the timeout at each of those entry points.  Instead, expand may_start_timeout_timer to guard against more categories of bad input.

Fixes https://github.com/basho/riak_kv/issues/571 

cc @jrwest @seancribbs

### DIFF
--- a/src/riak_core_coverage_fsm.erl
+++ b/src/riak_core_coverage_fsm.erl
@@ -199,8 +199,6 @@ maybe_start_timeout_timer(infinity) ->
     ok;
 maybe_start_timeout_timer(Bad) when not is_integer(Bad) ->
     maybe_start_timeout_timer(?DEFAULT_TIMEOUT);
-maybe_start_timeout_timer(undefined) ->
-    maybe_start_timeout_timer(?DEFAULT_TIMEOUT);
 maybe_start_timeout_timer(Timeout) ->
     gen_fsm:start_timer(Timeout, {timer_expired, Timeout}).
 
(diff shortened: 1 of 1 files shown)


src/riak_core_coverage_fsm.erl
```diff
@@ -195,6 +197,10 @@ init({test, Args, StateProps}) ->
 %% @private
 maybe_start_timeout_timer(infinity) ->
     ok;
+maybe_start_timeout_timer(Bad) when not is_integer(Bad) ->
+    maybe_start_timeout_timer(?DEFAULT_TIMEOUT);
+maybe_start_timeout_timer(undefined) ->
```
jrwest: this clause is covered by the previous one

src/riak_core_coverage_fsm.erl
```diff
@@ -97,6 +97,8 @@ behaviour_info(callbacks) ->
 behaviour_info(_) ->
     undefined.
 
+-define(DEFAULT_TIMEOUT, 60000*8).
```
jrwest, line 100: what's the reasoning behind this default? is it just supposed to be greater than all the other timeouts in layers above?

jrwest:
`client_java_verify` passes for me w/ these changes

src/riak_core_coverage_fsm.erl
evanmcc, line 100: It's the same default that's defined in the client.

jrwest, line 100: ok cool. just wanted to know its origin. thanks

jrwest:
+1

Issue #335, opened 2013-06-12
riak_core_stat:produce_stats() is never rescheduled to be called after a crash


bsparrow435:
If there is an error during the calculation of riak_core stats, it dies as expected but stats are never re-calculated(until node restart).
#### Example from the field

Error from log on 6/10:

```
Error in process <0.14823.27> on node 'OMMITTED' with exit value: {badarg,[{riak_core_stat,'-vnodeq_stats/0-lc$^0/1-0-',1,[{file,"src/riak_core_stat.erl"},{line,168}]},{riak_core_stat,'-vnodeq_stats/0-lc$^0/1-0-',1,[{file,"src/riak_core_stat.erl"},{line,169}]},{riak_core_stat,vnodeq_stats...
```

Report of last time stats were calculated(Run on 6/12):

```
 {'OMMITTED',{{2013,6,10},{20,31,40}}}]
Reported running_vnodes: 631
Actual running_vnodes: 83
```

jtuple:
This has been fixed for both 1.3.2 and master. Merged into 1.3.2 in #338 and merged to master in #336 

Closing issue.

PR #336, opened 2013-06-12
Fix crashing stat mod never getting rescheduled

russelldb:
See issue #335 

1.3.1 updated the cache to fetch stats in the background rather than
on demand. A new bug was added. If the stat mod crashes during
production of stats, it is never rescheduled.

Fix by rescheduling when crash is detected. Exponentially backoff
the schedule after an error so as not to spam the log.


russelldb:
Fixes https://github.com/basho/riak_core/issues/335

evanmcc:
is this the master version of #338?

evanmcc:
Also :+1:, sorry I didn't review it at the same time.

PR #338, opened 2013-06-14
Fix crashing stat mod never getting rescheduled 

russelldb:
1.3.1 updated the cache to fetch stats in the background rather than
on demand. A new bug was added. If the stat mod crashes during
production of stats, it is never rescheduled.

Fix by rescheduling when crash is detected. Exponentially backoff
the schedule after an error so as not to spam the log.

Also fix the case that led to bug discovery:     Fix crash when vnode Pid vanishes during stat calculation.


evanmcc:
:+1: can no longer easily get stats to hang forever.

PR #339, opened 2013-06-15
Fix repair handoff crash, missing not sent fun

engelsanchez:
/cc @jrwest @javajolt 
Partition repair was crashing on an undefined 'not sent function' in the handoff accumulator record. Jordan should know the details of the recent change that introduced him. Jordan: please review, feel free to replace with cleaner version of the fix.

Without this fix, the partition_repair riak_test fails every time. It passes with the fix.

### DIFF
--- a/src/riak_core_handoff_sender.erl
+++ b/src/riak_core_handoff_sender.erl
@@ -334,14 +334,16 @@ visit_item(K, V, Acc) ->
             end;
 
         false ->
-            NewNotSentAcc = case NotSentFun of
-                undefined -> undefined;
-                _ -> NotSentFun(K, NotSentAcc)
-            end,
+            NewNotSentAcc = handle_not_sent_item(NotSentFun, NotSentAcc, K),
             Acc#ho_acc{error=ok,
                        total_objects=TotalObjects+1,
                        notsent_acc=NewNotSentAcc}
     end.
+
+handle_not_sent_item(undefined, _, _) ->
+    undefined;
+handle_not_sent_item(NotSentFun, Acc, Key) when is_function(NotSentFun) ->
+    NotSentFun(Key, Acc).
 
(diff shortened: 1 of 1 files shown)


src/riak_core_handoff_sender.erl
```diff
@@ -334,9 +334,13 @@ visit_item(K, V, Acc) ->
             end;
 
         false ->
+            NewNotSentAcc = case NotSentFun of
```
jrwest: think it would be nice to pull this out into a separate function

jrwest:
@engelsanchez code changes make sense. Left one style comment but I can address that while testing this change if you are busy. Will run the riak_test's later this evening when I get home.

jrwest:
@engelsanchez trying to run the test but getting the error below. Pretty sure its a `riak_test` config thing but I have `spam_dir` set in `default` section of config to existing directory and was getting different error until i untarrred the search-corpus tar. Any ideas?

```
20:20:14.029 [warning] partition_repair failed: {undef,[{rt,config_or_os_env,[spam_dir],[]},{partition_repair,confirm,0,[{file,"tests/partition_repair.erl"},{line,32}]}]}
20:20:14.030 [error] Error in process <0.94.0> on node 'riak_test@127.0.0.1' with exit value: {undef,[{rt,config_or_os_env,[spam_dir],[]},{partition_repair,confirm,0,[{file,"tests/partition_repair.erl"},{line,32}]}]}
```

EDIT: not that i think it affects things but this happens w/ or w/o this branch

jrwest:
pushed a small cleanup commit for the comment I left (re: refactoring). `partition_repair`, with recent `riak_test` fixes,  passes with this change and fails w/o. Thanks for finding & fixing this @engelsanchez. +1.

PR #334, opened 2013-06-07
Reporting 'normal' events is spammy, don't do it

slfritchie:
Avoid log spamminess like this:

```
14:42:43.877 [info] 1141798154164767904846628775559596109106197299200 riak_kv_vnode ignored handle_info {'DOWN',#Ref<0.0.1.47104>,process,<0.9841.1>,normal} - vnode unregistering
14:42:47.720 [info] 1324485858831130769622089379649131486563188867072 riak_kv_vnode ignored handle_info {'DOWN',#Ref<0.0.1.42852>,process,<0.9416.1>,normal} - vnode unregistering
14:42:48.050 [info] 502391187832497878132516661246222288006726811648 riak_kv_vnode ignored handle_info {'DOWN',#Ref<0.0.1.49271>,process,<0.10111.1>,normal} - vnode unregistering
14:43:14.399 [info] 936274486415109681974235595958868809467081785344 riak_kv_vnode ignored handle_info {'DOWN',#Ref<0.0.1.52833>,process,<0.10501.1>,normal} - vnode unregistering
14:43:27.347 [info] 685078892498860742907977265335757665463718379520 riak_kv_vnode ignored handle_info {'DOWN',#Ref<0.0.1.49219>,process,<0.10092.1>,normal} - vnode unregistering
14:43:57.701 [info] 1027618338748291114361965898003636498195577569280 riak_kv_vnode ignored handle_info {'DOWN',#Ref<0.0.1.52781>,process,<0.10482.1>,normal} - vnode unregistering
```

PR #340, opened 2013-06-19
only silently drop DOWN-normal messages in deleted modstate

beerriot:
This is a restriction of the modification made in PR #334.

Dropping all `{'DOWN',_,process,_,normal}` messages on the floor instead of passing them to vnode `handle_info` functions causes `riak_pipe` vnodes to missing messages that it uses to cleanup workers for pipes that shutdown unexpectedly.

This commit restricts the DOWN-normal message dropping to the case that the vnode's modstate is `{deleted, _}`. PR #334 suggests the original modification was made only to quiet the log spam generated by the following clause, which also only operates in modstate-deleted.

Before this commit, the `riak_test` `pipe_verify_exceptions` would fail during its `verify_middle_fitting_normal` test, because workers would be left running after the fitting exited `normal`. After this commit, workers are once again terminated correctly, so the test passes again.

This PR resolves basho/riak_test#316.

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -784,7 +784,12 @@ handle_info({'EXIT', Pid, Reason},
             continue(State#state{pool_pid=NewPoolPid})
         end;
 
-handle_info({'DOWN',_Ref,process,_Pid,normal}, _StateName, State) ->
+handle_info({'DOWN',_Ref,process,_Pid,normal}, _StateName,
+            State=#state{modstate={deleted, _}}) ->
+    %% these messages are produced by riak_kv_vnode's aae tree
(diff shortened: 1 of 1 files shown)


beerriot:
@slfritchie and @jrwest - it would be great if you could take a look here to ensure this still fulfills the goals of #334

joedevivo:
ran tests on this

before: fail
after: pass
+1 from me, but still want eyes on the code from @jrwest or @slfritchie

jrwest:
Adding the `deleted` restriction makes sense. The original message fired while vnode was "unregistering" which is when `deleted` is set. Verified quickly that the messages are still suppressed.

PR #341, opened 2013-06-19
Refactor and better organization of the coverage plan module

kellymclaughlin:


jtuple:
Haven't looked at this code in detail yet, but did glance over to see how it played with the large ring changes. Sadly, this code needlessly reverts changes that were made during the large ring work. This needs to be addressed. Likewise, I wouldn't be surprised if much of the new logic also has similar problems as the code that was previous changed.

A few rules of thumb. 
1. Code that runs in the Riak critical path (basically, anything that effects user-visible request latency) shouldn't use `get_my_ring`. After the large ring work, the `get_my_ring` call is slower during periods of ring churn than it used to be. Prefer using `get_chashbin` or other equivalents. When in doubt, look at the commits in #319 to see how things were changed during the large ring work.
2. Likewise, prefer using `chashbin.erl` equivalents over `riak_core_ring` equivalents where appropriate: eg. `index_owner`, etc. The chashbin approach is constant-time, whereas `riak_core_ring` scales with the size of the ring.
3. Don't use `ordsets` or `orddict` if the size of such depends on the size of the ring. Prefer `sets` and `dict`.
4. All new code introduced into Riak after the large ring changes (ie. from this point on) should _always_ be tested, benchmarked, and profiled using a ring size of at least 16384. It's too easy to ship code that appears "fast enough" with 64-1024 size rings but than falls over on larger rings. Supporting up to 16k rings is the new standard for Riak.

Again, I haven't had the time to test or play with this myself (would love to, but it'd be after 1.3.2 and 1.4 ship), but wanted to note that on a quick glance I'm worried about this code on large rings. Especially since it reintroduced code that was already previously found to be less than desirable on large rings. Who ever does a formal review definitely needs to test this on a 16k ring/etc.

kellymclaughlin:
Good feedback. This is definitely a WIP PR and not one I viewed as ready to be merged. I simply rebased the branch of changes I had started a long time ago. I'll have a look at changes that may have been overwritten during rebase and use this info as a guide as we work through this.

zinid:
@jtuple the code from riak_core_coverage_plan.erl will NEVER be fast if written in Erlang. Forget about 16384. It's unreal: it will take tens of milliseconds to compute the find_coverage() function.

slfritchie:
Time for some belated repo curating ... is there any plan to un-WIP-ify this for Riak 2.0?

rzezeski:
I'd like to as it can help Yokozuna.  But it's not on the official list right now :/

kellymclaughlin:
Sadly I have not found any time to work on this and I doubt that will change before the 2.0 deadline. Unless someone else wants to claim this I will just close the PR.

slfritchie:
More curating ... Kelly, do you want to close this or give it a post-2.0 milestone?

Issue #267, opened 2013-01-14
gen_server call review


slfritchie:
While trying to start Riak under Valgrind, I encountered a gen_server:call() timeout exception which causes an application startup failure which prevents Riak from starting successfullly.

The culprit for my particular case is `riak_core_stat_cache.erl`:

```
 register_app(App, {M, F, A}, TTL) ->
    gen_server:call(?SERVER, {register, App, {M, F, A}, TTL}).
```

This is likely one of many uses of `gen_server:call/2` when we really ought to be using the 3-arity version with an infinity timeout.

slfritchie:
Time for some belated repo curating ... waiting for movement on PR #346.

jrwest:
#346 merged in some form. closing.

PR #346, opened 2013-07-03
gen_server:call() timeout fixes

slfritchie:
Use 'infinity' (or at least a longer timeout) when we know that the server pid is on the local node.

### DIFF
--- a/src/riak_core_tracer.erl
+++ b/src/riak_core_tracer.erl
@@ -54,15 +54,15 @@ start_link() ->
     gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
 
 stop() ->
-    gen_server:call(?SERVER, stop).
+    gen_server:call(?SERVER, stop, infinity).
 
 reset() ->
-    gen_server:call(?SERVER, reset).
+    gen_server:call(?SERVER, reset, infinity).
 
 %% Set up a filter on trace messages to the list of [{M, F}]s.  The
 %% tracer function should return a list of events to log
 filter(MFs, Filter) ->
-    gen_server:call(?SERVER, {filter, MFs, Filter}).
+    gen_server:call(?SERVER, {filter, MFs, Filter}, infinity).
 
 %% Collect traces
 collect() ->
@@ -72,15 +72,15 @@ collect(Duration) ->
     collect(Duration, nodes()).
 
 collect(Duration, Nodes) ->
-    gen_server:call(?SERVER, {collect, Duration,  Nodes}).
+    gen_server:call(?SERVER, {collect, Duration,  Nodes}, infinity).
 
 %% Stop collection
 stop_collect() ->
-    gen_server:call(?SERVER, stop_collect).
+    gen_server:call(?SERVER, stop_collect, infinity).
 
 %% Return the trace
 results() ->
-    gen_server:call(?SERVER, results).
+    gen_server:call(?SERVER, results, infinity).
 
 all_events({trace, Pid, call, {M,F,A}}) ->
     [{node(Pid), {M,F,A}}].
@@ -128,7 +128,7 @@ handle_call({collect, Duration, Nodes}, _From, State) ->
                                          {Mega, Secs, Micro} = os:timestamp(),
                                          Ts = 1000000 * (1000000 * Mega + Secs) + Micro,
                                          TsEntries = [{Ts, E} || E <- Entries],
-                                         gen_server:call(Pid,  {traces, TsEntries})
+                                         gen_server:call(Pid,  {traces, TsEntries}, infinity)
                                  end,
                                  Pid
                          end, self()}),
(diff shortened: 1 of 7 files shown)


slfritchie:
See also: #267

rzezeski:
I'm -1 on such a large number gen_server:call changes.  Just because a call is being made to a local process doesn't mean that local process might not get blocked on something thus causing a chain effect rippling through the system.  

Given a server process that has many client processes that talk to it, if that server process now uses infinity calls then one client process can lock-up the server process blocking all other clients when they don't require that the call be completed to do their work.

Or the repair calls for example.  I specifically designed those to not use infinity timeout.  Repair is designed in an idempotent manner.  If something fails retry it later.  I did this beacuse while reviewing @jtuple and I found many issues with blocking calls and timeouts, but using infinity would have introduced potential indefinite blocking on long-lived processes where we don't want that to happen (I think one was vnode_manager).  Changing to idempotent/async made all the issues and edge cases go away.

Also, not all these calls are local.

slfritchie:
@rzezeski So, if 'infinity' is bad, and the 5000 millisecond timeout is bad, then what is not bad?

IMO there are nearly zero uses in Riak where 5000 milliseconds makes sense. We should use infinity everywhere (all other gen_server calls and gen_fsm sync calls) and find any inter-process deadlocks and fix them.

reiddraper:
> Given a server process that has many client processes that talk to it, if that server process now uses infinity calls then one client process can lock-up the server process blocking all other clients when they don't require that the call be completed to do their work.

Any blocking call in the shared server has the chance of doing this. If the server needs to churn through a message queue of 100 messages, each taking 5 seconds, each client is still going to be blocked. If the server can actually handle concurrent work, then, as you mention, calls should be replaced with cast/async.

> Also, not all these calls are local.

Are non-local infinity calls actually dangerous? I was under the impression that the process you call is monitored, so if the node or process goes down, you'll be unstuck blocked.

rzezeski:
@slfritchie My point is neither is always good, and it should be a case-by-base decision.  This seems like a blanket commit with no reasons behind all the individual changes.

@reiddraper My point about the non-local calls was that Scott's original message mentions that he only changed local calls.  His reasons for that I do not know.  But not all these calls are local.

slfritchie:
Hrm, well, if I knew which patches were for non-local calls, then I could edit the patch to be more acceptable.

buddhisthead:
I don't know if this helps, but the way we handled this for replication was I went through all the changes and decided, based on knowledge of how the repl system works, if the call was actually local. It didn't take long and did uncover some issues where we were timing out and crashing (the cluster manager comes to mind). I think this is a really good idea; sorry I don't know enough about core to resolve the locals.

slfritchie:
Time for some belated repo curating ... IMO it's insane to let something that fixes nasty-obvious-stupid timeout bugs sit to rot.  @rzezeski What non-local calls are you objecting to?

src/riak_core_tracer.erl
```diff
@@ -128,7 +128,7 @@ handle_call({collect, Duration, Nodes}, _From, State) ->
                                          {Mega, Secs, Micro} = os:timestamp(),
                                          Ts = 1000000 * (1000000 * Mega + Secs) + Micro,
                                          TsEntries = [{Ts, E} || E <- Entries],
-                                         gen_server:call(Pid,  {traces, TsEntries})
+                                         gen_server:call(Pid,  {traces, TsEntries}, infinity)
```
rzezeski, line 131: This isn't always local.  But maybe we don't care since it's just for tracing?

rzezeski:
On second look it seems the things I was worried about were the calls related to vnode repair.  Which was designed to timeout + retry.  However, looking at the calls touched I wonder if they shouldn't be infinity.  I've been away from that code for too long.  I'd have to spend at least an hour swapping it all back into memory.  Given that you just increased the timeout for those calls I'm inclined to say +1.  I also noticed that for all but one non-local call you still have a timeout, it's just longer.  So that's good, I guess (I have no idea why non-local infinity calls are a bad idea, I just remember you saying something about them in the past @slfritchie ).

TL;DR - I'm +1 to merge.

Might be good to have one person from core/kv team +1.

src/riak_core_vnode_master.erl
```diff
@@ -120,7 +120,7 @@ command_return_vnode({Index,Node}, Msg, Sender, VMaster) ->
     Req = make_request(Msg, Sender, Index),
     case riak_core_capability:get({riak_core, vnode_routing}, legacy) of
         legacy ->
-            gen_server:call({VMaster, Node}, {return_vnode, Req});
+            gen_server:call({VMaster, Node}, {return_vnode, Req}, ?LONG_TIMEOUT);
```
rzezeski, line 123: This isn't local.  But it still has timeout so okay I guess.

slfritchie:
I rebased this branch on `develop` and merged with commit e1a5f31ed0f8, closing.  Thanks everyone!

PR #348, opened 2013-07-09
R16B01 compatibility changes.

evanmcc:
update hash functions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,12 +3,23 @@ APPS = kernel stdlib sasl erts ssl tools os_mon runtime_tools crypto inets \
 COMBO_PLT = $(HOME)/.riak_core_combo_dialyzer_plt
 PULSE_TESTS = worker_pool_pulse
 
+# The code below figures out the OTP release and introduces a macro at 
+# build and test time to tell later released to use the new hash
+# functions introduced in R15B02.  Older versions still use the old
+# hash functions.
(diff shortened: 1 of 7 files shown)


jaredmorrow:
This breaks compatibility with R15 I think.  There is no crypto:hash/2 function in R15B01.  The crypto:hash/2 function was added in R15B02.

evanmcc:
Bleah, forgot to push the commit fixing that here.  Will update.

evanmcc:
Updated.

joedevivo:
+1 to the changes in this PR.

Unfortunately, this PR leaves the eunit test suite broken due to a failing `core_vnode_eqc`. That test passes when run on it's own, but some test that runs before it seems to be causing some kind of side effect. We have noticed that R15B01 runs tests in a different order than R16B01. 

Here's the failure output: https://gist.github.com/joedevivo/fe82f31d25b3281a542b

The eunit suite on this branch runs successfully when using R15B01.

I would like two members of the core/kv cabal to +1 that the suite can remain in such a state until @evanmcc or I can track down the test causing the issue.

engelsanchez:
My thoughts here: If the change passes review and unit tests in R15B01, there's no need to hold back the merge. We will have to address the unit test order and probably many other issues as we test R16. So :+1: from me.

Issue #350, opened 2013-07-29
vnode_q len calculation error is not caught


russelldb:
Syntax problem https://github.com/basho/riak_core/blob/1.4/src/riak_core_stat.erl#L181

slfritchie:
Hrm, this looks merged, so I'm closing.  Please reopen if I'm wrong.

PR #351, opened 2013-07-29
Fix catch pattern to match all errors

russelldb:
Addresses https://github.com/basho/riak_core/issues/350


rzezeski:
+1 to merge

PR #355, opened 2013-08-09
Remove webmachine dependency

seancribbs:
As discussed on the [riak-core mailing list](http://lists.basho.com/pipermail/riak-core_lists.basho.com/2013-August/000050.html).  Additional changes:
1. Removed the `riak_core_config` module that was mostly dead code. Its only used export was fixed (see `riak_core_bucket:default_bucket_props/0`).
2. All references to `mochiglobal` were changed to the `riak_core_mochiglobal`, which is simply a copy/rename of the original module. This removes the transitive dependency on [mochiweb](/basho/mochiweb).
3. `json_pp` module used only in printing statistics in `riak_kv_wm_stats` was moved to [riak_kv](/basho/riak_kv/blob/sdc-wm-move/src/json_pp.erl).


rebar.config
```diff
@@ -11,8 +11,6 @@
                               {branch, "master"}}},
   {basho_stats, ".*", {git, "git://github.com/basho/basho_stats", "HEAD"}},
```
evanmcc, line 12: should protobuffs also get yanked?

seancribbs, line 12: It should be a separate PR. `riak_core.proto` is still in the `src` directory, even though it is only used by KV.

seancribbs:
Closing so I can rebase atop develop.

PR #356, opened 2013-08-15
Add protection against folsom stat errors

engelsanchez:
Folsom may sometimes return an error tuple if something goes wrong (see
folsom_ets.erl), but our code was only catching exceptions. So the error
would end up being used as a valid value and crash the riak_kv_stat
process later. This fixes that problem and gives us better protection
from folsom funkiness.

### DIFF
--- a/src/riak_core_stat_q.erl
+++ b/src/riak_core_stat_q.erl
@@ -90,6 +90,10 @@ get_stat(Stat) ->
     Pid = riak_core_stat_calc_sup:calc_proc(Stat),
     riak_core_stat_calc_proc:value(Pid).
 
+throw_folsom_error({error, _, _} = Err) ->
+    throw(Err);
+throw_folsom_error(Other) -> Other.
+
 %% Encapsulate getting a stat value from folsom.
 %%
 %% If for any reason we can't get a stats value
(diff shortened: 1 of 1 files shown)


src/riak_core_stat_q.erl
```diff
@@ -90,6 +90,9 @@ get_stat(Stat) ->
     Pid = riak_core_stat_calc_sup:calc_proc(Stat),
     riak_core_stat_calc_proc:value(Pid).
 
+err2unavailable({error, _, _}) -> unavailable;
```
evanmcc: is there a reason that we're not logging what the folsom error actually is here?  would it make sense to just throw an exception so that it gets dumped into the normal, logged error path in the catch below?

engelsanchez: Yes, it makes sense to change it so that all errors go through the same path. I'll update it soon.

engelsanchez:
Changed to throw folsom error tuples, so all errors go down the same code path. The astute reader might notice a rebase...

src/riak_core_stat_q.erl
```diff
@@ -90,6 +90,10 @@ get_stat(Stat) ->
     Pid = riak_core_stat_calc_sup:calc_proc(Stat),
     riak_core_stat_calc_proc:value(Pid).
 
+throw_folsom_error({error, _, _} = Err) ->
```
evanmcc, line 93: in an ideal world, I feel that this would be called maybe_throw_folsom_error, but that is deep in the nit-picky weeds.

engelsanchez, line 93: Well, if it's a folsom error, it will absolutely throw it. Not maybe. That's my argument and I'm sticking to it :)

evanmcc:
:+1:, all folsom errors hit the logging path now, neither crash kills stats, everything seems happy.

PR #357, opened 2013-08-15
Teach riak_core Bucket Types

jrwest:
Like buckets, we must teach core this because a few subsystems use them. Bucket Types provide a method for grouping buckets logically (see https://github.com/basho/riak/issues/362).

This PR requires Cluster Metadata which can be found on https://github.com/basho/riak_core/tree/corekv-unstable and https://github.com/basho/riak_core/tree/jrw-cluster-meta. This PR should not be merged until Cluster Metadata has the +1 and is merged as well.

### DIFF
--- a/src/riak_core_bucket_props.erl
+++ b/src/riak_core_bucket_props.erl
@@ -1,0 +1,232 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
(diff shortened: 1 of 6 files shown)


src/riak_core_bucket_props.erl
```diff
@@ -0,0 +1,232 @@
...
+    LinkFun1; %% arbitrary choice
+resolve({n_val, N1}, {n_val, N2}) ->
+    case N1 > N2 of
+        true -> N1;
+        false -> N2
+    end;
+resolve({notfound_ok, NF1}, {notfound_ok, NF2}) ->
+    NF1 orelse NF2;
+resolve({old_vclock, Old1}, {old_vclock, Old2}) ->
+    case Old1 > Old2 of
```
engelsanchez, line 114: This and all the other resolution cases would be more readable long term if we used good ol' max() instead IMO

PR #359, opened 2013-08-17
Corruption filtering changes for core.

evanmcc:
The visit_item function in handoff needs to be made aware that encode_handoff_data can possibly fail.


evanmcc:
dependency of: https://github.com/basho/riak_kv/pull/638

engelsanchez:
Tested this change along with basho/riak_kv#638, observed handoffs failing when corrupted objects were found, then pass with this patch applied to detect them. :+1: :dancer: :boat:

PR #362, opened 2013-08-27
Add bucket type defaults and reset

Vagabond:
Non-default bucket-types are unaffected by changes to app.config and
bucket types can have their properties reset to the hardcoded defaults.

cc @jrwest


Vagabond:
basho/riak#362

engelsanchez:
:+1:

PR #363, opened 2013-08-28
Cluster Metadata (Part 1/2)

jrwest:
This PR is the first of two that comprises the initial version of `riak_core`'s new cluster-wide meta data store. In the future, it is intended for this store to be used in place of metadata currently stored in the ring. 

**[part 2](https://github.com/basho/riak_core/pull/370) should be merged into this PR before merging into develop**

This PR includes the following:
- in-memory (and for-now on-disk) representation of metadata objects
- per-node local storage in DETS with full in-memory copy in ETS
- API for local access to metadata (get/put/iterators/folds)
- Initial implementation of broadcast support for replication (see more below)

This PR does not yet include, but may before merge:
- subsystem documentation (in docs/)

This PR does not include (see [part 2](https://github.com/basho/riak_core/pull/370)):
- ensured at-least once delivery in `riak_core_broadcast` (this is what will be included in part 2)
- local tracking of changes for groups of keys

In addition, this PR includes an implementation of [Dotted Version Vectors](https://github.com/ricardobcl/Dotted-Version-Vectors#how-to-use) which are used as the logical clock for metadata objects. The implementation is a direct copy of the implementation [built by Carlos](https://github.com/ricardobcl/riak_core/blob/dvvset/src/dvvset.erl) including the MIT license. This license, at a minimum, must be addressed pre-merge.

### DIFF
--- a/src/riak_core_broadcast.erl
+++ b/src/riak_core_broadcast.erl
@@ -1,0 +1,473 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(riak_core_broadcast).
+
+-behaviour(gen_server).
+
+%% API
+-export([start_link/0,
+         start_link/3,
+         broadcast/2,
+         ring_update/1,
+         broadcast_members/0,
+         set_peers/3,
+         neighbors_down/1]).
+
+%% Debug API
+-export([debug_get_peers/2,
+         debug_get_tree/2]).
+
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+-define(SERVER, ?MODULE).
+
+-type nodename()        :: any().
+-type message_id()      :: any().
+-type message_round()   :: non_neg_integer().
+-type outstanding()     :: {message_id(), module(), message_round(), nodename()}.
+
+-record(state, {
+          %% Initially trees rooted at each node are the same.
+          %% Portions of that tree belonging to this node are
+          %% shared in this set.
+          common_eagers :: ordset:ordset(nodename()),
+
+          %% Initially trees rooted at each node share the same lazy links.
+          %% Typically this set will contain a single element. However, it may
+          %% contain more in large clusters and may be empty for clusters with
+          %% less than three nodes.
+          common_lazys  :: ordset:ordset(nodename()),
+
+          %% A mapping of sender node (root of each broadcast tree)
+          %% to this node's portion of the tree. Elements are
+          %% added to this structure as messages rooted at a node
+          %% propogate to this node. Nodes that are never the
+          %% root of a message will never have a key added to
+          %% `eager_sets'
+          eager_sets    :: [{nodename(), ordset:ordset(nodename())}],
+
+          %% A Mapping of sender node (root of each spanning tree)
+          %% to this node's set of lazy peers. Elements are added
+          %% to this structure as messages rooted at a node
+          %% propogate to this node. Nodes that are never the root
+          %% of a message will never have a key added to `lazy_sets'
+          lazy_sets     :: [{nodename(), ordset:ordset(nodename())}],
+
+          %% Lazy messages that have not been acked. Messages are added to
+          %% this set when a node is sent a lazy message (or when it should be
+          %% sent one sometime in the future). Messages are removed when the lazy
+          %% pushes are acknowleged via graft or ignores. Entries are keyed by their
+          %% destination
+          outstanding   :: [{nodename(), outstanding()}],
+
+          %% Set of modules that handle messages that have been broadcast
+          handling_mods :: ordset:ordset(module()),
+
+          %% Set of all known members. Used to determine
+          %% which members have joined and left during a membership update
+          all_members   :: ordset:ordset(nodename())
+         }).
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+%% @doc Starts the broadcast server on this node. The initial membership list is
+%% fetched from the ring. If the node is a singleton then the initial eager and lazy
+%% sets are empty. If there are two nodes, each will be in the others eager set and the
+%% lazy sets will be empty. When number of members is less than 5, each node will initially
+%% have one other node in its eager set and lazy set. If there are more than five nodes
+%% each node will have at most two other nodes in its eager set and one in its lazy set, initally.
+%% In addition, after the broadcast server is started, a callback is registered with ring_events
+%% to generate membership updates as the ring changes.
+-spec start_link() -> {ok, pid()} | ignore | {error, term()}.
+start_link() ->
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    Members = all_broadcast_members(Ring),
+    {InitEagers, InitLazys} = init_peers(Members),
+    Res = start_link(Members, InitEagers, InitLazys),
+    riak_core_ring_events:add_sup_callback(fun ?MODULE:ring_update/1),
+    Res.
+
+%% @doc Starts the broadcast server on this node. `InitMembers' must be a list
+%% of all members known to this node when starting the broadcast server.
+%% `InitEagers' are the initial peers of this node for all broadcast trees.
+%% `InitLazys' is a list of random peers not in `InitEagers' that will be used
+%% as the initial lazy peer shared by all trees for this node. If the number
+%% of nodes in the cluster is less than 3, `InitLazys' should be an empty list.
+%% `InitEagers' and `InitLazys' must also be subsets of `InitMembers'.
+%%
+%% NOTE: When starting the server using start_link/2 no automatic membership update from
+%% ring_events is registered. Use start_link/0.
+-spec start_link([nodename()], [nodename()], [nodename()]) -> {ok, pid()} | ignore | {error, term}.
+start_link(InitMembers, InitEagers, InitLazys) ->
+    gen_server:start_link({local, ?SERVER}, ?MODULE, [InitMembers, InitEagers, InitLazys], []).
+
+%% @doc Broadcasts a message originating from this node. The message will be delivered to
+%% each node at least once. The `Mod' passed is responsible for handling the message on remote
+%% nodes as well as providing some other information both locally and and on other nodes.
+%% `Mod' must be loaded on all members of the clusters and implement the
+%% `riak_core_broadcast_handler' behaviour.
+-spec broadcast(any(), module()) -> ok.
+broadcast(Broadcast, Mod) ->
+    {MessageId, Payload} = Mod:broadcast_data(Broadcast),
+    gen_server:cast(?SERVER, {broadcast, MessageId, Payload, Mod}).
+
+
+%% @doc Notifies broadcast server of membership update given a new ring
+-spec ring_update(riak_core_ring:riak_core_ring()) -> ok.
+ring_update(Ring) ->
+    CurrentMembers = ordsets:from_list(all_broadcast_members(Ring)),
+    BroadcastMembers = ?MODULE:broadcast_members(),
+    New = ordsets:subtract(CurrentMembers, BroadcastMembers),
+    Removed = ordsets:subtract(BroadcastMembers, CurrentMembers),
+    case ordsets:size(New) > 0 of
+        false -> ok;
+        true ->
+            {EagerPeers, LazyPeers} = init_peers(CurrentMembers),
+            ?MODULE:set_peers(CurrentMembers, EagerPeers, LazyPeers)
+    end,
+    ?MODULE:neighbors_down(Removed).
+
+%% @doc Returns the broadcast servers view of full cluster membership.
+-spec broadcast_members() -> ordset:ordset(nodename()).
+broadcast_members() ->
+    gen_server:call(?SERVER, broadcast_members).
+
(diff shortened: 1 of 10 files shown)


src/riak_core_broadcast.erl
```diff
@@ -0,0 +1,473 @@
...
+
+%% @doc Broadcasts a message originating from this node. The message will be delivered to
+%% each node at least once. The `Mod' passed is responsible for handling the message on remote
+%% nodes as well as providing some other information both locally and and on other nodes.
+%% `Mod' must be loaded on all members of the clusters and implement the
+%% `riak_core_broadcast_handler' behaviour.
+-spec broadcast(any(), module()) -> ok.
+broadcast(Broadcast, Mod) ->
+    {MessageId, Payload} = Mod:broadcast_data(Broadcast),
+    gen_server:cast(?SERVER, {broadcast, MessageId, Payload, Mod}).
```
jtuple, line 134: As a cast, this message isn't guaranteed to actually reach the server. Eg. call `broadcast` right as the server is crashing for some reason. We never acknowledge the broadcast actually made it to the server, that it was merged into local state, that any eager messages were triggered, etc. Is this fine? Would we prefer changing to a call that returns after some level of guarantee was met? Something else?

jtuple:
I would really like to see `riak_core_broadcast:ring_update` changed before merging. Currently, this is called in the ring event handler and performs a series of calls + casts to the `riak_core_broadcast` server. This is multiple back and forth messages. Likewise, both `set_members/3` and `neighbors_down/1` are casts that can theoretically be dropped leaving the server partially updated.

Why not move all of the `ring_update` logic inside the `riak_core_broadcast` server itself? Then just have `ring_update` send either a single cast or call? This leads to a single message, that the server handles at once, with full access to it's entire state. No need to do multiple queries or send separate messages after the fact.

After said change, you can also remove `set_members/3` and `neighbors_down/1` as they are solely used to implement the current `ring_update` approach.

jtuple:
In both `riak_core_broadcast` and `riak_core_metadata_manager` we have local `gen_server:calls` that are using the default 5 second timeout. I imagine these should all be changed to infinity, unless there is a good reason for a timeout (even then, is 5 seconds the right value?).

src/riak_core_broadcast.erl
jrwest, line 134: This is the same problem as if we were to process the message while partitioned and then crash I believe (we lose the lazy queue and at that point have no record of the message). This case should be covered by the hashtree work. Changing it to a call probably makes it a bit less likely but then means the metadata write path blocks on briefly broadcast for not much extra guarantee. 

Regarding local state, the local process is required to "deliver" the message first locally before calling broadcast. This ensures a future exchange will include the broadcast e.g.: https://github.com/basho/riak_core/pull/363/files#L5R272

src/riak_core_broadcast.erl
```diff
@@ -0,0 +1,473 @@
...
+
+%% @private
+-spec handle_cast(term(), #state{}) -> {noreply, #state{}} |
+                                       {noreply, #state{}, non_neg_integer()} |
+                                       {stop, term(), #state{}}.
+handle_cast({broadcast, MessageId, Message, Mod}, State) ->
+    State1 = eager_push(MessageId, Message, Mod, State),
+    State2 = schedule_lazy_push(MessageId, Mod, State1),
+    {noreply, State2};
+handle_cast({broadcast, MessageId, Message, Mod, Round, Root, From}, State) ->
```
jtuple, line 224: Just a minor comment, not something that needs to be changed unless you agree (and even then, can be in a later PR), but I'm not a huge fan of having two "broadcast" messages: one being local to implement the `broadcast/2` API, and the other being a remote message sent by other nodes as part of the actual protocol. Wonder if "send_broadcast" would be more appropriate for local message, or perhaps keep that "broadcast" but have "remote_broadcast" or "broadcast_message" for the remove message. Dunno. Again, just clarity point.

jrwest:
re: https://github.com/basho/riak_core/pull/363#issuecomment-23977787. agreed. not sure why it was written this way. will change. (addressed in https://github.com/basho/riak_core/commit/744c6cea39a678ac5080edc27a8b45d53febbd3b)

re: https://github.com/basho/riak_core/pull/363#issuecomment-23977869 they are 5 seconds right now because my lazy brain prefers to type one less word :). will change. (addressed for riak_core_broadcast.erl in https://github.com/basho/riak_core/commit/82942f2a56429945e389873ca431b81937f8ddcc and for riak_core_metadata_manager.erl in https://github.com/basho/riak_core/commit/c8cb75f649b6e042bdb20097c3d1ef621f0e0716)

src/riak_core_broadcast.erl
```diff
@@ -0,0 +1,473 @@
...
+handle_broadcast(true, MessageId, Message, Mod, Round, Root, From, State) -> %% valid msg
+    State1 = add_eager(From, Root, State),
+    State2 = eager_push(MessageId, Message, Mod, Round+1, Root, From, State1),
+    schedule_lazy_push(MessageId, Mod, Round+1, Root, From, State2).
+
+handle_ihave(true, MessageId, Mod, Round, Root, From, State) -> %% stale i_have
+    send({ignored_i_have, MessageId, Mod, Round, Root, node()}, From),
+    State;
+handle_ihave(false, MessageId, Mod, Round, Root, From, State) -> %% valid i_have
+    %% TODO: don't graft immediately
```
jtuple, line 290: Future plan is delayed grafting with some timers? Planning this for 2.0 or punting? What are we losing without?

src/riak_core_broadcast.erl
```diff
@@ -0,0 +1,473 @@
...
+    end.
+
+send(Msg, Peers) when is_list(Peers) ->
+    [send(Msg, P) || P <- Peers];
+send(Msg, P) ->
+    %% TODO: add debug logging
+    gen_server:cast({?SERVER, P}, Msg).
+
+schedule_lazy_tick() ->
+    TickMs = app_helper:get_env(riak_core, broadcast_lazy_timer, 1000),
```
jtuple, line 435: Reminder that need to decide on a good default value before we actually ship. If we have a todo list somewhere, good thing to add.

src/riak_core_broadcast.erl
jrwest, line 224: I'm fine changing the local one to something like `send_broadcast` (the remote one is already long enough being a 7-tuple) -- i thought about making them records but meh. For me, i see them as distinct messages because a 4-tuple and a 7-tuple are different "types". But again, not opposed to it if you found it awkward while reading.

src/riak_core_broadcast.erl
```diff
@@ -0,0 +1,473 @@
...
+    State#state{
+      common_eagers = ordsets:del_element(node(), ordsets:from_list(EagerPeers)),
+      common_lazys  = ordsets:del_element(node(), ordsets:from_list(LazyPeers)),
+      eager_sets    = orddict:new(),
+      lazy_sets     = orddict:new(),
+      all_members   = ordsets:from_list(AllMembers)
+     }.
+
+all_broadcast_members(Ring) ->
+    riak_core_ring:members(Ring, [joining, valid, leaving, exiting, down]).
```
jtuple: Any reason we're not using `riak_core_ring:all_members` here? Do we ever envisioning we'll want some smaller subset than "all members"? Prefer not to leak node status outside of ring/claimant logic unless absolutely necessary.

jrwest: nope I just kept looking at `active_members` instead and missing `all_members`. will change

src/riak_core_broadcast.erl
jrwest, line 290: Yes. I'd like to try to get this in for 2.0 but it can wait. The basic gist is that the tree healing is more eager without the delay. There is also the optimization mentioned in the plumtree paper (which uses the rounds that are currently just passed around) that I would like to squeeze in as well.

src/riak_core_broadcast.erl
jrwest, line 435: I have my personal TODO list and we have some TODO items scattered in these PRs and code but it would probably be good to organize them all under once place. Perhaps github issues w/ a cluster metadata tag or something?

src/riak_core_broadcast_handler.erl
```diff
@@ -0,0 +1,36 @@
...
+-callback merge(any(), any()) -> boolean().
+
+%% @doc return true if the message (given the message id) has already been received.
+%% `false' otherwise
+-callback is_stale(any()) -> boolean().
+
+%% @doc return the message associated with the given message id. In some cases a message
+%% has already been sent with information that subsumes the message associated with the given
+%% message id. In this case, `stale' is returned.
+-callback graft(any()) -> stale | {ok, any()}.
```
jtuple: Can also return `{error, term()}`.

src/riak_core_broadcast.erl
jrwest: addressed in https://github.com/basho/riak_core/commit/744c6cea39a678ac5080edc27a8b45d53febbd3b

src/riak_core_broadcast_handler.erl
jrwest: addressed in https://github.com/basho/riak_core/commit/ee0d41f848e26df7daceeeaf156fb6f6cb2d24b5

jtuple:
With recent code changes, looks like all my earlier review concerns have been met.

Manual testing of cluster metadata looks reasonable, although I think there's one thing we need definitely need to fix in a future PR. I tested the entire metadata API while always doing various things like randomly killing broadcast and metadata managers, deleting keys while iterating/folding, etc.

The main issue is that iterators and folds are not safe to concurrent updates/deletion.

Consider the following:

``` erlang
riak_core_metadata:fold(fun(X, Acc) ->
                                io:format("X: ~p~n", [X]),
                                timer:sleep(1000),
                                Acc
                        end, ok, {abc, xyz}, []).

X: {test2,[testing]}
X: {test3,[testing]}
X: {test4,[testing1234]}
X: {test,[testing]}

%% Alright, that's good. Now, let's be evil and delete keys in our fold.
%% A concurrent fold has similar effect, but this seems even more evil so why not?

riak_core_metadata:fold(fun(X, Acc) ->
                                io:format("X: ~p~n", [X]),
                                timer:sleep(1000),
                                riak_core_metadata:delete({abc, xyz}, X, []),
                                Acc
                        end, ok, {abc, xyz}, []).

X: {test2,[testing]}
X: {test3,[testing]}
X: {test4,[testing1234]}
X: {test,[testing]}
X: {{test4,[testing1234]},['$deleted']}
X: {{test2,[testing]},['$deleted']}
X: {{test,[testing]},['$deleted']}

%% Huh, that's unexpected output....
```

Do we want read snapshot isolation? Is this a job for [`safe_fixtable/2`](http://www.erlang.org/doc/man/ets.html#safe_fixtable-2)? Do we have other plans/thoughts?

Also, some other "obvious" things. If the metadata manager dies, all existing iterators become invalid. Or rahter, the ETS continuations become invalid. I don't think we care, and it's a classic "let it crash" issue, but something to at least think about if we ever use metadata iterators in sensitive code (ie. should wrap in try/catch). Of course, we could build try/catch into metadata API itself too. Again, something to think about. Some examples of errors that can happen when manager dies while concurrently iterating:

``` erlang
** exception exit: {{badarg,
                        [{ets,match_object,[{3145763,130,1,<<>>,[],0}],[]},
                         {riak_core_metadata_manager,next_iterator,1,
                             [{file,"src/riak_core_metadata_manager.erl"},{line,305}]},
                         {riak_core_metadata_manager,handle_call,3,
                             [{file,"src/riak_core_metadata_manager.erl"},{line,258}]},
                         {gen_server,handle_msg,5,
                             [{file,"gen_server.erl"},{line,588}]},
                         {proc_lib,init_p_do_apply,3,
                             [{file,"proc_lib.erl"},{line,227}]}]},
                    {gen_server,call,
                        [riak_core_metadata_manager,
                         {iterate,
                             {metadata_iterator,
                                 {abc,xyz},
                                 undefined,
                                 {3145763,130,1,<<>>,[],0},
                                 {test2,
                                     {metadata,
                                         {[{'dev1@127.0.0.1',1,[{testing,{1378,526792,418220}}]}],
                                          []}}},
                                 false,3145763}},
                         infinity]}}
     in function  gen_server:call/3 (gen_server.erl, line 188)
     in call from riak_core_metadata:itr_next/1 (src/riak_core_metadata.erl, line 156)

** exception exit: {killed,
                       {gen_server,call,
                           [riak_core_metadata_manager,
                            {iterate,
                                {metadata_iterator,
                                    {abc,xyz},
                                    undefined,
                                    {36962339,130,1,<<>>,[],0},
                                    {test2,
                                        {metadata,
                                            {[{'dev1@127.0.0.1',1,[{testing,{1378,526792,418220}}]}],
                                             []}}},
                                    false,36962339}},
                            infinity]}}
     in function  gen_server:call/3 (gen_server.erl, line 188)
     in call from riak_core_metadata:itr_next/1 (src/riak_core_metadata.erl, line 156)
     in call from riak_core_metadata:fold_it/3 (src/riak_core_metadata.erl, line 121)
```

Finally, a known issue is cases where dets isn't closed properly and needs to be repaired:

```
dets: file "./data/cluster_meta/manifest.dets" not properly closed, repairing ...
```

And, of course, all known issues that are addressed in the follow-up PR #370

This is good enough to merge for pre1, but we have work before 2.0:
1. Review/merge in #370 in order to handle known metadata data loss scenarios.
2. Consider hardening metadata against the issues mentioned above. Especially the issue of deletes/updates during iteration. That seems bad. The iterator invalidation seems less bad, but is something to think about.
3. Consider if we want to move away from dets to something else.
4. My review over the `dvvsets.erl` code was minimal, and something we still need to test and review in more detail. Worst case, cluster metadata can be changed to using `vclock.erl` instead. We may want to consider making the choice between dvvsets and vector clocks something that can be toggled via `app.config` plus a node restart.
5. TODO items discussed in this review: determining a good lazy tick internal and adding delayed grafting.

In any case, +1 merge for pre1.

jrwest:
Addressing the delete/update issue during iteration I noticed that the example has a bug. Its deleting a different key (namely the key and the value as a 2-tuple). Quick scans by me and joe before missed that it was spitting out slightly different keys. This became evident when you rant it over and over again (pesky erlang tuples). Anyways, when you change the example to the version below the "issue" no longer exists:

``` erlang
riak_core_metadata:fold(fun({K, X}, Acc) ->
                                io:format("K: ~p X: ~p~n", [K, X]),
                                timer:sleep(1000),
                                riak_core_metadata:delete({abc, xyz}, K, []),
                                Acc
                        end, ok, {abc, xyz}, []).
```

That said this doesn't mean there isn't an issue lurking in a more complex example. more research is necessary but this is also less critical of an issue now. It does however, highlight that the iterators are not read isolated (but for cluster metadata we wonder if thats actually necessary -- you may want to see new keys or updates as long as you don't see a key more than once).

PR #364, opened 2013-08-29
Background Manager (Locks + Tokens)

buddhisthead:
Background Manager = Lock Manager + Token Manager

Token manager is broken out to a separate gen_server. Someday, I will break lock manager out as well. Not time now.
Token calls are available in the background manager and proxied to the token manager server.
Changed enabled/disabled in BG manager to affect BOTH subsystems.
Renamed enable/disable for locks to enable_lock/disable_lock to match API for tokens: enable_tokens/disable_tokens

The token manager has a separate set of eunit tests that have good coverage:
rebar skip_deps=true eunit suite=token_manager

This PR also includes a new module "riak_core_table_manager", which creates and owns tables until a process claims them. It remains the heir of a table so that a crashed table user process can get it back again, without the table going away. The table manager is supervised by riak_core_sup, as is a new riak_core_bg_manager_sup which oversees the lock and token managers.

Theory of operation...
The table manager is started by the riak_core_sup and is given the names and options for the new lock and token manager servers. These tables persist across crashed lock/token managers so that locks and tokens are not lost.

The "background" manager is really locks + tokens, but they don't interact except for the master on/off switch. Lock manager allows locks to be taken and handles release when processes die with locks held. Also, if the maximum count for a lock is lowered, processes will be killed to enforce that new limit. Tokens are dispensed as a token bucket scheduling mechanism. Tokens are replenished at some specified rate. Both lock limits and token rates must be registered ahead of their use.

As an example, the AAE entropy manager from riak_kv has been ported to use the locks and tokens from bg manager. It is on a separate branch named "cet-bg-mgr-aae": https://github.com/basho/riak_kv/pull/645

It is anticipated that other subsystems will start using the background manager. We might even add locks and tokens to vnode folds to avoid thrashing the backends.

### DIFF
--- a/include/riak_core_bg_manager.hrl
+++ b/include/riak_core_bg_manager.hrl
@@ -1,0 +1,23 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-define(LM_ETS_TABLE, lock_mgr_table).   %% name of private lock manager ETS table
+-define(LM_ETS_OPTS, [private, bag]).    %% creation time properties of lock manager ETS table
+
+-type concurrency_limit() :: non_neg_integer() | infinity.
--- a/src/riak_core_bg_manager.erl
+++ b/src/riak_core_bg_manager.erl
@@ -71,8 +71,8 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
--record(state, {held    :: ordict:orddict(),
-                info    :: orddict:orddict(),
+-record(state, {table_id:: ets:tid(),            %% TableID of ?LM_ETS_TABLE
+                held    :: ordict:orddict(),
                 enabled :: boolean()}).
 
 -record(lock_info, {concurrency_limit :: non_neg_integer(),
@@ -84,6 +84,8 @@
 -define(limit(X), (X)#lock_info.concurrency_limit).
 -define(enabled(X), (X)#lock_info.enabled).
 -define(DEFAULT_LOCK_INFO, #lock_info{enabled=true, concurrency_limit=?DEFAULT_CONCURRENCY}).
+-define(LM_ETS_TABLE, lock_mgr_table).   %% name of private lock manager ETS table
+-define(LM_ETS_OPTS, [private, bag]).    %% creation time properties of lock manager ETS table
 
 -type concurrency_limit() :: non_neg_integer() | infinity.
 
@@ -312,7 +314,7 @@ concurrency_limit_reached(Type) ->
                   ignore |
                   {stop, term()}.
 init([]) ->
-    {ok, #state{info=orddict:new(),
+    {ok, #state{
                 held=orddict:new(),
                 enabled=true}}.
 
@@ -338,8 +340,9 @@ handle_call({lock_limit_reached, LockType}, _From, State) ->
     HeldCount = held_count(LockType, State),
     Limit = ?limit(lock_info(LockType, State)),
     {reply, HeldCount >= Limit, State};
-handle_call(lock_types, _From, State=#state{info=Info}) ->
-    Types = [{Type, ?enabled(LI), ?limit(LI)} || {Type, LI} <- orddict:to_list(Info)],
+handle_call(lock_types, _From, State=#state{table_id=TableId}) ->
+    Infos = [{Type,Info} || {{info, Type},Info} <- ets:match_object(TableId, {{info, '_'},'_'})],
+    Types = [{Type, ?enabled(LI), ?limit(LI)} || {Type, LI} <- Infos],
     {reply, Types, State};
 handle_call({query_locks, Query}, _From, State) ->
     Results = query_locks(Query, State),
@@ -378,6 +381,11 @@ handle_cast(disable, State) ->
 -spec handle_info(term(), #state{}) -> {noreply, #state{}} |
                                        {noreply, #state{}, non_neg_integer()} |
                                        {stop, term(), #state{}}.
+%% Handle transfer of ETS table from table manager
+handle_info({'ETS-TRANSFER', TableId, Pid, _Data}, State) ->
+    lager:debug("table_mgr (~p) -> bg_mgr (~p) receiving ownership of TableId: ~p", [Pid, self(), TableId]),
+    State2 = State#state{table_id=TableId},
+    {noreply, State2};
 handle_info({'DOWN', Ref, _, _, _}, State) ->
     State2 = release_lock(Ref, State),
     {noreply, State2};
@@ -498,13 +506,21 @@ update_concurrency_limit(LockType, Limit, State) ->
                      ?DEFAULT_LOCK_INFO#lock_info{concurrency_limit=Limit},
                      State).
 
-update_lock_info(LockType, Fun, Default, State=#state{info=Info}) ->
-    NewInfo = orddict:update(LockType, Fun, Default, Info),
-    State#state{info=NewInfo}.
+update_lock_info(LockType, Fun, Default, State=#state{table_id=TableId}) ->
+    Key = {info, LockType},
+    NewInfo = case ets:lookup(TableId, Key) of
+                  [] -> Default;
+                  [{_Key,LockInfo}] -> Fun(LockInfo)
+              end,
+    ets:insert(TableId, {Key, NewInfo}),
+    State.
 
-
-lock_info(LockType, #state{info=Info}) ->
-    case orddict:find(LockType, Info) of
-        error -> ?DEFAULT_LOCK_INFO;
-        {ok, LockInfo} -> LockInfo
+lock_info(LockType, #state{table_id=TableId}) ->
+    Key = {info,LockType},
+    case ets:lookup(TableId, Key) of
+        [] -> ?DEFAULT_LOCK_INFO;
+        [{_Key,LockInfo}] -> LockInfo;
+        [First | _Rest] ->
+            lager:error("Unexpected multiple instances of key ~p in table", [{info, LockType}]),
+            First %% try to keep going
     end.
--- a/src/riak_core_bg_manager.erl
+++ b/src/riak_core_bg_manager.erl
@@ -1,0 +1,210 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(riak_core_bg_manager).
+
+-behaviour(gen_server).
+
+%% API
+-export([start_link/0,
+         get_lock/1,
+         get_lock/2,
+         get_lock/3,
+         lock_count/0,
+         lock_count/1,
+         enable/1,
+         disable/1,
+         concurrency/1,
+         set_concurrency/2
+        ]).
+
+%% gen_server callbacks
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2,
+         terminate/2, code_change/3]).
+
+-define(DEFAULT_CONCURRENCY, 2).
+
+-define(SERVER, ?MODULE). 
+
+-record(state, { locks  :: ordict:orddict(),
+                 limits :: orddict:orddict() }).
+
+%%%===================================================================
+%%% API
+%%%===================================================================
+
+%% @doc Starts the server
+-spec start_link() -> {ok, pid()} | ignore | {error, term}.
+start_link() ->
+    gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
+
+%% @doc Acquire a concurrency lock of the given type, if available,
+%%      and associate the lock with the calling process.
+-spec get_lock(any()) -> ok | max_concurrency.
+get_lock(Type) ->
+    get_lock(Type, self()).
+
+%% @doc Acquire a concurrency lock of the given type, if available,
+%%      and associate the lock with the provided pid.
+-spec get_lock(any(), pid()) -> ok | max_concurrency.
+get_lock(Type, Pid) ->
+    get_lock(Type, Pid, []).
+
+%% @doc Acquire a concurrency lock of the given type, if available,
+%%      and associate the lock with the provided pid.
+%%      TODO: info on options, but there are none right now
+-spec get_lock(any(), pid(), [{atom(), any()}]) -> ok | max_concurrency.
+get_lock(Type, Pid, Opts) ->
+    gen_server:call(?MODULE, {get_lock, Type, Pid, Opts}, infinity).
+
+%% @doc Return the current concurrency count for all lock types
+-spec lock_count() -> integer().
+lock_count() ->
+    gen_server:call(?MODULE, lock_count, infinity).
+
+%% @doc Return the current concurrency count of the given lock type.
+-spec lock_count(any()) -> integer().
+lock_count(Type) ->
+    gen_server:call(?MODULE, {lock_count, Type}, infinity).
+
+%% @doc Enable handing out of locks of the given type.
+-spec enable(any()) -> ok.
+enable(Type) ->
+    gen_server:call(?MODULE, {enable, Type}, infinity).
+
+%% @doc Disable handing out of locks of the given type.
+-spec disable(any()) -> ok.
+disable(Type) ->
+    gen_server:call(?MODULE, {disable, Type}, infinity).
+
+%% @doc Get the current maximum concurrency for the given lock type.
+-spec concurrency(any()) -> integer().
+concurrency(Type) ->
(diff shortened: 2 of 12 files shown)


jrwest:
the lock manager seems to have some issues stemming from its use of a `bag` table to store both locks and lock info. The first two functions below will crash the lock manager. The third returns unexpected results. I'll comment on some specific places in the code where its related to this issue.

``` erlang
lock_info_issue() ->
    Args = [{lock_mgr_table, [private, bag]}],
    riak_core_table_manager:start_link(Args),
    riak_core_bg_manager:start_link(),
    riak_core_bg_manager:set_concurrency_limit(a, 2),
    false = riak_core_bg_manager:concurrency_limit_reached(a),
    riak_core_bg_manager:set_concurrency_limit(a, 3),
    %% this call crashes
    riak_core_bg_manager:concurrency_limit_reached(a).

lock_info_update_issue() ->
    Args = [{lock_mgr_table, [private, bag]}],
    riak_core_table_manager:start_link(Args),
    riak_core_bg_manager:start_link(),
    riak_core_bg_manager:set_concurrency_limit(a, 2),
    riak_core_bg_manager:set_concurrency_limit(a, 3),
    riak_core_bg_manager:set_concurrency_limit(a, 4).

lock_types_issue() ->
    Args = [{lock_mgr_table, [private, bag]}],
    riak_core_table_manager:start_link(Args),
    riak_core_bg_manager:start_link(),
    riak_core_bg_manager:set_concurrency_limit(a, 2),
    riak_core_bg_manager:set_concurrency_limit(a, 3),
    Types = riak_core_bg_manager:lock_types(),
    {Types, length(Types) =:= 1}.
```

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+              end,
+    ets:insert(TableId, {Key, NewInfo}),
+    State.
+
+lock_info(LockType, #state{table_id=TableId}) ->
+    Key = {info,LockType},
+    case ets:lookup(TableId, Key) of
+        [] -> ?DEFAULT_LOCK_INFO;
+        [{_Key,LockInfo}] -> LockInfo;
+        [First | _Rest] ->
```
jrwest, line 526: `First` should be `{Key, First}` or else this clause returns the whole object not just the `lock_info` record. I wonder though if we should move away from using `bag` for the info (see above comment) -- the primary cause is inserting the same key with two different values, e.g. updating a type, means we hit this case.

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+    update_lock_info(LockType,
+                     fun(LockInfo) -> LockInfo#lock_info{concurrency_limit=Limit} end,
+                     ?DEFAULT_LOCK_INFO#lock_info{concurrency_limit=Limit},
+                     State).
+
+update_lock_info(LockType, Fun, Default, State=#state{table_id=TableId}) ->
+    Key = {info, LockType},
+    NewInfo = case ets:lookup(TableId, Key) of
+                  [] -> Default;
+                  [{_Key,LockInfo}] -> Fun(LockInfo)
```
jrwest, line 516: this doesn't handle the case where there are multiple lock info values and can cause a crash

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+            ok
+    end;
+maybe_honor_limit(false, _LockType, _Limit, _State) ->
+    ok.
+
+held_count(LockType, State) ->
+    length(held_locks(LockType, State)).
+
+held_locks(#state{table_id=TableId}) ->
+    [Lock || {{held, _Type},Lock} <- ets:match_object(TableId, {{held, '_'},'_'})].
```
jrwest, line 489: instead of iterating the table and then the match list take a look at `ets:select/2`:

``` erlang
held_locks_match() ->
    T = ets:new(undefined, [bag]),
    ets:insert(T, [{{held, type1}, a}, {{held, type1}, b}, {{held, type2}, a}, {{held, type1}, c}]),
    Res1 = [Lock || {{held, _Type},Lock} <- ets:match_object(T, {{held, '_'},'_'})],
    Res2 = ets:select(T, [{{{held, '_'}, '$1'},[],['$1']}]),
    ets:delete(T),
    {Res1, Res2, lists:usort(Res1) =:= lists:usort(Res2)}.
```

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+add_lock(LockType, Lock, State=#state{table_id=TableId}) ->
+    Key = {held, LockType},
+    ets:insert(TableId, {Key, Lock}),
+    State.
+
+release_lock(Ref, State=#state{table_id=TableId}) ->
+    %% There should only be one instance of the object, but we'll zap all that match.
+    Pattern = {{held, '_'}, {'_', Ref, '_'}},
+    Matches = [Lock || {{held, _Type},Lock} <- ets:match_object(TableId, Pattern)],
+    [ets:delete_object(TableId, Obj) || Obj <- Matches],
```
jrwest, line 468: think you can use `match_delete` here instead

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+    {reply, held_count(LockType, State), State};
+handle_call(lock_count, _From, State) ->
+    Count = length(held_locks(State)),
+    {reply, Count, State};
+handle_call({lock_limit_reached, LockType}, _From, State) ->
+    HeldCount = held_count(LockType, State),
+    Limit = ?limit(lock_info(LockType, State)),
+    {reply, HeldCount >= Limit, State};
+handle_call(lock_types, _From, State=#state{table_id=TableId}) ->
+    Infos = [{Type,Info} || {{info, Type},Info} <- ets:match_object(TableId, {{info, '_'},'_'})],
```
jrwest, line 343: because of the `bag` for info, this can return the same type multiple times

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+%% @doc Queries the currently held locks returning any locks that match the given criteria.
+%%      If no criteria is present then all held locks are returned. The query is a proplist of
+%%      2-tuples. The keys 'pid' and 'type', have special meaning. If they are keys in the
+%%      query proplists, only locks matching the corresponding pid or lock type are
+%%      returned. All other pairs are compared for equality against the proplist passed as the third
+%%      argument to get_lock/3. The returned value is a list of 4-tuples. The first element
+%%      is the lock type; the second, the pid holding the lock; the third, the lock refernce,
+%%      and the fourth is the metadata passed to get_lock/3.
+-spec query_locks([{atom(), any()}]) -> [{any(), pid(), reference(), [{atom(), any()}]}].
+query_locks(Query) ->
```
jrwest, line 250: for things like `query_locks` and `lock_types`, since we are using ets and these are read-only, any reason not to just go directly to the ets table (besides having to have a named table, but we do this in many places)?

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+    ok.
+
+held_count(LockType, State) ->
+    length(held_locks(LockType, State)).
+
+held_locks(#state{table_id=TableId}) ->
+    [Lock || {{held, _Type},Lock} <- ets:match_object(TableId, {{held, '_'},'_'})].
+
+held_locks(LockType, #state{table_id=TableId}) ->
+    [Lock || {{held, _Type},Lock} <- ets:match_object(TableId, {{held, LockType},'_'})].
```
jrwest, line 492: this can also use `ets:select`

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+
+-record(lock_info, {concurrency_limit :: non_neg_integer(),
+                    enabled           :: boolean()}).
+
+-define(SERVER, ?MODULE).
+-define(DEFAULT_CONCURRENCY, 0). %% DO NOT CHANGE. DEFAULT SET TO 0 TO ENFORCE "REGISTRATION"
+-define(limit(X), (X)#lock_info.concurrency_limit).
+-define(enabled(X), (X)#lock_info.enabled).
+-define(DEFAULT_LOCK_INFO, #lock_info{enabled=true, concurrency_limit=?DEFAULT_CONCURRENCY}).
+-define(LM_ETS_TABLE, lock_mgr_table).   %% name of private lock manager ETS table
```
jrwest, line 86: this is duplicated in `riak_core_bg_manager.hrl`

include/riak_core_bg_manager.hrl
```diff
@@ -0,0 +1,23 @@
...
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-define(LM_ETS_TABLE, lock_mgr_table).   %% name of private lock manager ETS table
+-define(LM_ETS_OPTS, [private, bag]).    %% creation time properties of lock manager ETS table
+
+-type concurrency_limit() :: non_neg_integer() | infinity.
```
jrwest, line 23: is it necessary for this type to be in the `.hrl` a quick grep shows its only used in `riak_core_bg_manager.erl`. If it is to be used, maybe make it an exported type from the bg manager.

src/riak_core_bg_manager_sup.erl
```diff
@@ -0,0 +1,47 @@
...
+%%
+%% Lock manager (currently in bg_manager) and Token manager both use ETS
+%% tables to manage persistent state. The table_manager is started from
+%% our parent supervisor (riak_core_sup) and creates the tables used by
+%% background managers. So, riak_core_table_manager needs to be started
+%% before this supervisor.
+
+%% @private
+init ([]) ->
+    {ok,{{one_for_all,10,10},
```
jrwest, line 45: why `one_for_all`. aren't the two processes pretty independent (NOTE: haven't looked much at token manager yet, so let me know if I'm way off)

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+%% @private
+%% @doc Convert process state when code is changed
+-spec code_change(term() | {down, term()}, #state{}, term()) -> {ok, #state{}}.
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+query_locks(FullQuery, State) ->
```
jrwest, line 413: I wonder if some `select/match`-fu would reduce the amount of iteration here.

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+%% @private
+%% @doc Initializes the server
+-spec init([]) -> {ok, #state{}} |
+                  {ok, #state{}, non_neg_integer() | infinity} |
+                  ignore |
+                  {stop, term()}.
+init([]) ->
+    lager:debug("Background Manager starting up."),
+    %% claiming the table will result in a handle_info('ETS-TRANSFER', ...) message.
+    ok = riak_core_table_manager:claim_table(?LM_ETS_TABLE),
```
jrwest, line 317: not sure if this will be a problem but just wanted to get your opinion. There seems to be a very slight chance of race between the delivery of the transfer message and a message from an operation, e.g. `get_lock`. If the latter beats the former the process will crash. Seems to me its highly unlikely and even less likely it would happen two starts in a row.

jrwest:
Finished most of the initial code review of the lock API of the `bg_manager`. Have yet to look at tokens. Some other general comments before I move on:
- I wonder if we should remove the handoff changes and put them in a seperate PR. It makes this PR easier to merge, being all new code then, and there are some changes that may be necessary there (like falling back to the old handoff concurrency mechanism just in case (tm)).
- the commit history got messed up somewhere along the way. there are a bunch of duplicate commits. since most of the code is new, should be pretty quick to make a new branch w/ a fresh history.

buddhisthead:
Jordan, I will look into the issues in the lock manager re: crashes. And before I change any code, I might try and build a new branch. The double commits probably happened as a result of a rebase and the pain around moving the PR to be against develop instead of master.

jrwest:
should the following succeed against the token manager? 

``` erlang
%% set really high refresh rate. div 1000 here b/c sep. issue
riak_core_token_manager:set_token_rate(a, {4294967295 div 1000, 2}) -> {0, 0}
riak_core_token_manager:get_token_sync(a, self(), []) -> ok
riak_core_token_manager:get_token_sync(a, self(), []) -> ok
riak_core_token_manager:get_token_sync(a, self(), []) -> ok
```

I expected that even though the same process was asking for tokens since there was no refresh that `max_tokens` would be returned in the 3rd call.

Dumping the token manager's ets table it looks like after the three calls there is only one `given` entry in the table:

``` erlang
[{{given,a},{token_entry,a,<0.723.0>,[],undefined,given}},
 {{blocked,a},{[],[]}},
 {{info,a},{token_info,{4294967,2},true}}]
```

src/riak_core_token_manager.erl
```diff
@@ -0,0 +1,705 @@
...
+%% This is needed because we just reloaded our saved persisent state data
+%% after a crash.
+reschedule_token_refills(State) ->
+    Types = do_token_types(State),
+    [schedule_refill_tokens(Type, State) || Type <- Types].
+    
+%% Schedule a timer event to refill tokens of given type
+schedule_refill_tokens(Type, State) ->
+    {Period, _Count} = ?rate(token_info(Type, State)),
+    erlang:send_after(Period*1000, self(), {refill_tokens, Type}).
```
jrwest, line 554: should `Period` multiplied by 1000 here? the comment near the definition of `tm_period()` says its in milliseconds.

src/riak_core_bg_manager.erl
```diff
@@ -0,0 +1,529 @@
...
+%%      Returns "max_tokens" if empty.
+-spec get_token(any(), pid() | [{atom(), any()}]) -> ok | max_tokens.
+get_token(Type, Pid) when is_pid(Pid) ->
+    get_token(Type, Pid, []);
+get_token(Type, Meta) ->
+    get_token(Type, self(), Meta).
+
+-spec get_token(any(), pid(), [{atom(), any()}]) -> ok | max_concurrency.
+get_token(Type, Pid, Meta) ->
+    riak_core_token_manager:get_token_async(Type, Pid, Meta).
```
slfritchie, line 140: The use of "async" here is a bit different than I expect.  I expect "async" to mean that there's a `gen_server:cast/2` or `!` used, and we don't wait for a reply.  AFAICT, this ends up being a `gen_server:call/2` to the token server and thus a full round-trip and thus feeling pretty synchronous.  :-)

slfritchie, line 140: Also, the -spec here appear to be a bit wonky: ok | max_concurrency doesn't match ok | max_tokens for the /1 and /2 arity flavors.

slfritchie:
I've more general comments, based on what I've read in research papers and seen in other projects.

Token-based or ticket-based or admission-based rate control systems typically have the notion of a ticket or token.  If you want to do X, you first need to ask for a ticket that gives you permission to do X.  (Whether or not the system will allow you to do X without a ticket isn't a concern here.)  If the intent is rate limiting, e.g. for network bandwidth use & control, then the ticket/token/thingie is disposable: the ticket agent/server/manager/service doesn't require the ticket be returned or be notified when the ticket was actually "used" or anything like that.

For other applications, the ticket/token/chit is long-lived and is returned to the agent/server/manager/service when the requestor has finished doing X.  In these cases, the agent/server/manager/service is acting like a library that loans books: the books are loaned to various parties repeatedly.  If a party misbehaves (e.g. crashes), it's nice if the ticket/token/chit/library-book is returned to the agent/.../.  If the party misbehaves by remaining alive forever but not returning the ticket/.../, then the system may/may not be able to do something about the misbehavior.

Both types of systems have a manager thingie that manages the supply of ticket/.../ things and manages requests for them.  In the first use case, because tickets are never returned, there's a replenishment every so often.  In the second use case, there's never a replenishment because the full set of ticket/.../ things is created at the start of the system.

It seems to me that the "token"s in the PR fit the first use case that I describe, and "locks" in the PR fit the second use case?  Other implementations I've seen & read about unify these two, whereas they're separate in this PR.  I think it's worthwhile to consider whether unification is useful here.

In the "lock" case, I think that it's also useful to have the option of having more than 1 simultaneous acquirer: this allows semaphore-style use instead of simple lock-style use.

buddhisthead:
Hi Scott,

I did consider the choice of giving/expecting/taking-back tickets and finally settled on the give/don't-expect-them-back/replenish semantics and implementation. Locks and Tokens are really doing two very different things here and do have different semantics. I think for our use in riak, we want to keep the current semantics.

However, I agree that we can unify the implementation. This work initially started with @jrwest building the lock manager and myself building the token manager. Then, we realized that we needed to persist the state to handle crashes of the managers, which lead to the managed ETS table. I think it's a good idea to re-work this code so that we have only a single table and common code to handle the "entries". This will also make it easier to unify the query and history stats, which is a good thing.

Also, I will be making a new PR that smashes the commit logs and tidies things up, which I'll reference from here. I will also reference back to this one, so we don't lose our discussion notes.

src/riak_core_bg_manager.erl
buddhisthead, line 526: Resolved in refactored version.

src/riak_core_bg_manager.erl
buddhisthead, line 343: Yep. Addressed in the insert; we always make sure to delete existing record first.

src/riak_core_bg_manager.erl
buddhisthead, line 250: This is handled differently now. See query_resource()

src/riak_core_bg_manager.erl
buddhisthead, line 86: Now only present in .hrl file.

include/riak_core_bg_manager.hrl
buddhisthead, line 23: Now that it's all one module, it's needed here.

src/riak_core_bg_manager.erl
buddhisthead, line 413: Maybe it would. I'm in the mode of "make it right" now. Perhaps in another PR, I can do some optimizations. This code path is now critical.

src/riak_core_bg_manager.erl
buddhisthead, line 317: Yeah, I suppose it's possible. I could guard all API calls with a check for table ownership. It's probably a good thing to do, but I'm really feeling the heat of time. I'll make an issue for this and come back to it. Maybe an EQC test is in order :-)

src/riak_core_token_manager.erl
buddhisthead, line 554: Fixed comment to show seconds.

src/riak_core_bg_manager.erl
buddhisthead, line 140: Agreed. Changed API to just 'get_token()' and 'get_token_blocking()'. Same for 'get_lock()' and 'get_lock_blocking()'.

Also changed spec and API to use 'max_concurrency' for both kinds of resources: token and lock.

buddhisthead:
I've refactored the two modules into one, and they share lots of common code and types. However, I have not unified tokens and locks in any bigger sense. If you have any ideas on how to do that, I'd be interested to hear more. In the mean time, I think this API/semantics will serve us well for keeping background jobs from stomping each other's performance.

jrwest:
closing as this work has merged in a different form

PR #367, opened 2013-09-04
Fix meck related unit test problems

engelsanchez:
Meck now needs a couple of flags in meck:new for some of our tests:
- non_strict so we can create a fresh non-existent module
- no_link so it doesn't link to the processes created with a meck
- module. Bad when we kill things on purpose.


rzezeski:
Tests fail before.  Tests pass after.

+1 to merge.

PR #368, opened 2013-09-05
Merge 1.4.2 to develop

engelsanchez:
Testing status: most tests should pass, but occasionally I'm seeing the vnode proxy overload and stat cache tests fail, unfortunately.


evanmcc:
:+1: code looks good, eunit passes except for some known brokenness

Issue #373, opened 2013-09-09
Add Tests for Cluster Metadata


jrwest:
Cluster Metadata is woefully missing eunit/eqc/riak_tests. Of course, it was not built untested, but the tests were not built into a framework. Do so.

some tests:
- broadcast/eventual consistency/exchanges
- iterators
- hashtree_tree
- riak_tests that add and remove nodes

jrwest:
some work on broadcast/eventual consistency tests can be found in https://github.com/basho/riak_core/tree/jrw-cluster-meta-eqc-rebased

jrwest:
More testing work can be found in https://github.com/basho/riak_test/tree/jrw-more-cluster-meta-tests

jrwest:
testing work that needs to be done for `2.0-RC`

PR #380, opened 2013-09-10
QuickCheck tests for riak_core_broadcast

UlfNorell:
My git skills are somewhat lacking. Hopefully I didn't mess anything up... :)

### DIFF
--- a/test/broadcast_eqc.erl
+++ b/test/broadcast_eqc.erl
@@ -1,0 +1,189 @@
+%%% File        : broadcast_eqc.erl
+%%% Author      : Ulf Norell
+%%% Description : 
+%%% Created     : 20 Aug 2013 by Ulf Norell
+-module(broadcast_eqc).
+
+-compile(export_all).
+
+-include_lib("eqc/include/eqc.hrl").
+-include_lib("eqc/include/eqc_statem.hrl").
+-include_lib("eqc/include/eqc_temporal.hrl").
+
+-record(state, {nodes = []}).
+
+-define(LAZY_TIMER, 20).
+
+initial_state() -> #state{}.
+
+node_list() ->
+  [a, b, c, d, e, f, g, h, i, j, k, l].
+
+%% -- Generators -------------------------------------------------------------
+
+key() -> elements([k1, k2, k3, k4, k5]).
+val() -> elements([v1, v2, v3, v4, v5]).
+msg() -> {key(), val()}.
+
+%% -- Commands ---------------------------------------------------------------
+
+%% -- init --
+init_pre(S) -> S#state.nodes == [].
+
+init_args(_S) ->
+  [?LET(N, choose(1, length(node_list())), shrink_list(lists:sublist(node_list(), N)))].
+
+init(Names) ->
+  Nodes = [ mk_node(N) || N <- Names ],
+  [ rpc:call(mk_node(Name), ?MODULE, start_server, [Name, Nodes]) || Name <- Names ],
+  %% log_tree(Names),
+  ok.
+
+start_server(_Name, Nodes) ->
+  %% TODO: deterministic lazy_peers!
+  {Eager, Lazy} = riak_core_broadcast:init_peers(Nodes),
+  {ok, Pid} = riak_core_broadcast:start_link(Nodes, Eager, Lazy),
+  unlink(Pid),
+  (catch metadata_manager_mock:stop()),
+  timer:sleep(1),
+  {ok, Mgr} = metadata_manager_mock:start_link(),
+  unlink(Mgr),
+  ok.
+
+init_next(S, _, [Names]) ->
+  S#state{ nodes  = Names }.
+
+%% -- broadcast --
+broadcast_pre(S) -> S#state.nodes /= [].
+broadcast_pre(S, [Node, _Msg]) -> lists:member(Node, S#state.nodes).
+
+broadcast_args(S) -> [elements(S#state.nodes), msg()].
+
+broadcast(Node, Msg={Key, Val}) ->
+  %% event_logger:event({broadcast, Node, Msg}),
+  rpc:call(mk_node(Node), metadata_manager_mock, put, [Key, Val]),
+  rpc:call(mk_node(Node), riak_core_broadcast, broadcast, [Msg, metadata_manager_mock]).
+    %% Parameterize on manager
+
+%% -- sleep --
+sleep_pre(S) -> S#state.nodes /= [].
+sleep_args(_) -> [choose(1, ?LAZY_TIMER)].
+sleep(N) -> timer:sleep(N).
+
+weight(_, sleep)     -> 1;
+weight(_, broadcast) -> 4;
+weight(_, init)      -> 1.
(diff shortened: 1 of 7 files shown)


test/broadcast_eqc.erl
```diff
@@ -0,0 +1,319 @@
...
+
+weight(_, sleep)     -> 1;
+weight(_, broadcast) -> 4;
+weight(_, init)      -> 1.
+
+%% -- Property ---------------------------------------------------------------
+
+prop_test() ->
+  ?SETUP(fun() -> setup(), fun() -> ok end end,
+  ?FORALL(Cmds, ?SIZED(N, resize(N div 2, commands(?MODULE))),
```
jrwest, line 206: just want to confirm my assumption here that the reason for resizing commands is to shorten the length of the test since tests can run pretty long.

test/broadcast_eqc.erl
```diff
@@ -0,0 +1,319 @@
...
+  ?LET(Shrinking, parameter(shrinking, false),
+  ?ALWAYS(if Shrinking -> 1; true -> 1 end,
```
jrwest, line 208: wondering about the `?LET/?ALWAYS` here. I know what parameter generates (although i'm not sure how to set it using `eqc_gen:with_parameter/3`). Is using `?ALWAYS(1, ...)` different from removing it?

test/broadcast_eqc.erl
```diff
@@ -0,0 +1,319 @@
...
+    event_logger:event(reset),
+    {messages, Mailbox} = process_info(global:whereis_name(event_logger), messages),
+    timer:sleep(10),
+    Tree = (catch get_tree(S#state.nodes)),
+    stop_servers(),
+    timer:sleep(5),
+    Views = [ {Node, get_view(Node)} || #node{name = Node} <- S#state.nodes ],
+    timer:sleep(10),
+    MoreTrace = event_logger:get_events(),
+    ?IMPLIES(length(MoreTrace) == 2,
```
jrwest, line 224: this is to ensure that trace is valid?

test/riak_core_broadcast_eqc.erl
```diff
@@ -0,0 +1,233 @@
...
+eager_push_callouts(S, [Node, Id, Body, Round, Root, From]) ->
+  ?PAR([ ?CALLOUT(peer, cast, [P, {broadcast, Id, Body, ?MODULE, Round, slave_node(Root), slave_node(Node)}], ok)
+        || P <- eager_peers(S, Root), P /= From ]).
+
+lazy_push_next(S, _, [Node, Id, _Body, Round, Root, From]) ->
+  S#state{ lazy_queue = lists:umerge(S#state.lazy_queue,
+            lists:sort([ {P, {i_have, Id, ?MODULE, Round, slave_node(Root), slave_node(Node)}}
+                        || P <- lazy_peers(S, Root), P /= From ])) }.
+
+dispatch_callouts(_S, []) -> ?EMPTY.
```
jrwest, line 133: is this needed? I couldn't find where it was used and removing does not seem to affect anything.

test/riak_core_broadcast_eqc.erl
```diff
@@ -0,0 +1,233 @@
...
+  begin
+    timer:sleep(2),
+    HSR={_H, S, Res} = run_commands(?MODULE, Cmds),
+    stop_peers(S),
+    pretty_commands(?MODULE, Cmds, HSR,
+      Res == ok)
+  end)).
+
+api_spec() ->
+  #api_spec{
```
jrwest, line 159: the docs seem to suggest that api specs need a `language`. Is `undefined` valid because this is the only spec? Also, I don't see any calls to `init_lang`. Is this because the test is checking that the callouts are actually called during execution as expected the the mocking of `data` is primarily for generation of values using `?CALLOUT` and the mocking of `data` is primarily of validating that `riak_core_broadcast_peer`s were sent the messages we expect?

test/broadcast_eqc.erl
UlfNorell, line 206: Yes that's correct.

test/broadcast_eqc.erl
UlfNorell, line 208: ?LET-bound things don't show up in the counterexample. In this case we're looking at the special shrinking parameter to possibly try harder falsifying a test case when shrinking.
?ALWAYS(1, P) is equivalent to P, so at the moment it doesn't do anything, but one might want something
other than 1 in the case of hard to shrink bugs.

test/broadcast_eqc.erl
UlfNorell, line 224: We had some problems with messages between nodes being delayed by the VM, resulting in part of the trace arriving after we had decided the test was complete. This happens sufficiently rarely that we could use an ?IMPLIES to simply discard such cases.

test/riak_core_broadcast_eqc.erl
UlfNorell, line 133: I forgot we still had riak_core_broadcast_eqc in there. We started on a component model, but it's not the best fit for this example. Component models describe quite precisely the interaction of a component with the rest of the system and in this case I don't think we want to model the broadcast protocol in such detail.

Bottom line: this model isn't finished and I don't think it's what we want to have.

test/riak_core_broadcast_eqc.erl
UlfNorell, line 159: I believe the default mocking language is erlang.

Yes, the component library takes care of driving the mocking framework, checking that you get the calls specified by your ?CALLOUTS.

test/broadcast_eqc.erl
jrwest, line 208: thanks for the explanation.

test/riak_core_broadcast_eqc.erl
jrwest, line 133: ah thanks for the heads up. was still a good way to dip my toe in the component waters.

slfritchie:
Jordan: what's the status of this PR?

Also, how the heck is it supposed to be used?  I created a kludge by the following:

```
mkdir /tmp/p
cp .eunit/*beam /tmp/p
cp deps/*/ebin/*beam /tmp/p
erl -sname foo@localhost -pz .eunit deps/*/ebin
broadcast_eqc:start_nodes().
rpc:multicall(code, add_pathz, ["/tmp/p"]).
eqc:quickcheck(broadcast_eqc:prop_test()).
```

jrwest:
@slfritchie this is still a work in progress but I'm going to close this specific PR since the work has been continued [1] and is tracked elsewhere [2]. The goal is to move the eqc to `riak_test` and extend it during the testing cycle. 

As for running it, I believe if you just make the `.eunit` directory the working directory and run erlang from there (with a `-sname`) then everything will run as expected. This is one of the reasons I plan to move it to `riak_test`. Once we extend the eqc to include the hashtree work dealing w/ all the path additions, etc will be easier in that environment.

[1]  https://github.com/basho/riak_core/compare/jrw-cluster-meta-eqc-rebased
[2] https://github.com/basho/riak_core/issues/373

PR #383, opened 2013-09-11
Move tcp-mon, connection and service manager supervisors from replication to core.

buddhisthead:
This is the matching PR for replication below. This is the final move to migrate the connection and service
managers from replication into their proper home in core. The riak_core_tcp_mon supervisor also belongs here.

If you build/test this on riak_ee, you will need the following PR as well, which is recently rebased on develop.
https://github.com/basho/riak_repl/pull/398

### DIFF
--- a/src/riak_core.app.src
+++ b/src/riak_core.app.src
@@ -12,6 +12,7 @@
                   stdlib,
                   lager,
                   sasl,
+                  ranch,
                   crypto,
                   riak_sysmon,
                   webmachine,
--- a/src/riak_core_conn_services_sup.erl
+++ b/src/riak_core_conn_services_sup.erl
@@ -1,8 +1,8 @@
 %% -------------------------------------------------------------------
 %%
-%% riak_core: Core Riak Application
+%% riak_core_conn_services_sup: supervise connection mananger services
 %%
-%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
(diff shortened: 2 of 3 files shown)


seancribbs:
@buddhisthead Looks like you missed adding the module.

buddhisthead:
@seancribbs You were so right! I had to re-create this module because, of course, I had blown away the local version. If you don't mind having a look, I would really appreciate it. Thanks.

cmeiklejohn:
I've built this and ran the eunits and it looks good.  Also, I've also verified that this new change works in the Riak ETL client code, where we manually started the supervisor.  However, this branch should be squashed because it contains a bunch of intermediary commits that don't build independently and would be better as a single commit into Riak Core.

buddhisthead:
Thanks, Chris. I will visit sifu for git squash training next.

buddhisthead:
Ah, squash is cool. Much better, but had to force push this branch. Please update if you have a local copy.

cmeiklejohn:
Looks great to me; waiting for a Core/KV :+1:

buddhisthead:
@cmeiklejohn rebased on latest develop

slfritchie:
Ping?

seancribbs:
@slfritchie Seems like since riak_core_connection_mgr and friends moved back to repl, this won't be needed. Confirm @cmeiklejohn ?

cmeiklejohn:
Correct; closing.

PR #370, opened 2013-09-05
Cluster Metadata (Part 2/2) - Hashtrees & Exchanges

jrwest:
This PR is the second of two comprising the initial version of Cluster Metadata. It builds on [part 1](https://github.com/basho/riak_core/pull/363), ensuring that all updates are delivered at-least once.

**This PR is targeted at `jrw-cluster-meta` not `develop` it should be merged there before part 1 is merged.**

This PR includes:
- hashtree.erl - the original hash tree from riak_kv (unmodified)
- hashtree_tree.erl - hashtree used by cluster metadata
- metadata hashtree exchanges and repair of divergent cluster metadata
- generalized exchange mechanism for riak_core_broadcast

(see commits for more details)

This PR does not yet include, but should before merge (into develop but to be safe lets say part 1):
- [x] [hashtree documentation](https://github.com/basho/riak_kv/blob/develop/docs/hashtree.md) from basho/riak_kv
- [x] riak_kv pull request to remove hashtree.erl and documentation (NOTE: need careful merge here to not break develop for people) - https://github.com/basho/riak_kv/pull/652
- [x] change references from `riak_kv` to `riak_core` in hashtree.erl
- [x] look at failing builds (failed tests fixed by rebase)

This PR does not include the following but i think it can be done concurrently with review and added to this PR or to new PRs:
- [x] if a node crashes abnormally hashtrees stored in leveldb are not properly cleaned up (included in #385)
- [ ] there is some minor refactoring that could be done here and there (may also come up during review)
- [ ] `riak_core_broadcast` knows which modules will broadcast but does not validate it
- add tests (need to convert a utilty test module into proper eunit/r_t/etc) and have some eqc tests in the works (issue: https://github.com/basho/riak_core/issues/373)
- [x] module and function edocs
- [x] logging (included in #385)

This PR will not include:
- memory backed hashtree.erl (which was originally planned)

Issue #372, opened 2013-09-09
Harden Metadata Iterators


jrwest:
As mentioned in the [review](https://github.com/basho/riak_core/pull/363#issuecomment-23981530) of basho/riak_core#363 cluster metadata iterators suffer from a few issues:

a) iterators are not safe to concurrent updates/deletion
b) iterators become stale when metadata manager crashes and then subsequent use of the iterators will cause further crashes

there are several possible solutions to a) depending on the semantics we want. Surely, at a minimum, using [`safe_fixtable/2`](http://www.erlang.org/doc/man/ets.html#safe_fixtable-2) or similar to ensure keys are only seen once is necessary.

for b) the remote iterators added in #370 may provide a basis for a solution. Although we don't want all local iterator operations serialized through the metadata manager, exposing the iterators ets table may do the trick (need to investigate more before committing to this as a solution)

jrwest:
As mentioned [here](https://github.com/basho/riak_core/pull/363#issuecomment-24296163) this issue is not as critical as we originally thought (or at least the issue we thought existed doesn't). However, there may be a more complex issue hidden deep in ets and so more research is necessary. It should also be discussed whether or not we care about read-isolated snapshots for cluster metadata iterators.

jrwest:
So after further research I think there are some minimal changes to be made to the code in #370 to address the iterator concurrency concerns. Assuming that what we mean by "safety" for iterators is the following:
- Iterating while there are concurrent updates does not crash the iterator or the metadata manager
- A key should appear at most once during a full iteration. The iterator may see updates newer than when the iterator was opened and may see keys inserted after the iterator was opened (its not read-isolated)

Then I think the current implementation mostly satisfies this. From the [ets docs](http://www.erlang.org/doc/man/ets.html) on concurrency:

```
...the safe_fixtable/2  function can  be used  to guarantee  that a
sequence of first/1 and next/2 calls will traverse the table
without errors and that each existing object in the table is
visited exactly once,  even if another process  (or the same
process) simultaneously deletes or  inserts objects into the
table.  Nothing more  is guaranteed;  in particular  objects
that are inserted or deleted  during such a traversal may be
visited  once  or  not  at all.  Functions  that  internally
traverse over a table, like  select and match, will give the
same guarantee as safe_fixtable.
```

[...]

PR #385, opened 2013-09-15
Cluster Metadata Part 2 Rebased

jrwest:
Replaces #370. Rebased on develop since jrw-cluster-meta branch was merged. In addition, this PR adds a few things that were not included in #370. If they were mentioned in the original PR they are noted there.


jtuple:
In `hashtree_tree.erl`, it seems possible to leak snapshot ETS if [`update_snapshot/1`](https://github.com/basho/riak_core/blob/ded140d8948a71fb155d30617afe8992b84a7b78/src/hashtree_tree.erl#L245) is called but [`update_perform/1`](https://github.com/basho/riak_core/blob/ded140d8948a71fb155d30617afe8992b84a7b78/src/hashtree_tree.erl#L257) never is. Obviously, this is all performed pretty straightline in [`riak_core_metadata_hashtree:update_async/3`](https://github.com/basho/riak_core/blob/ded140d8948a71fb155d30617afe8992b84a7b78/src/riak_core_metadata_hashtree.erl#L271), but given that `update_snapshot` is called in a spawned pid, you never really know.

In the normal `hashtree.erl`, we keep track of the snapshot state in both the returned snapshot copy of the tree, and the primary copy of the tree. The next time the primary tree is snapshotted again, we close the prior snapshot (if one exists), before creating the new one: see [here](https://github.com/basho/riak_core/blob/ded140d8948a71fb155d30617afe8992b84a7b78/src/hashtree.erl#L694). Should we do similar with the ETS snapshot table for `hashtree_tree` as a "just to be safe" change?

src/hashtree_tree.erl
```diff
@@ -0,0 +1,564 @@
...
+    SnapTree = UpdatedTree#hashtree_tree{snapshot=Snapshot},
+    {SnapTree, UpdatedTree}.
+
+
+%% @doc Update the tree with a snapshot obtained by {@link
+%% update_snapshot/1}. This function may be called by a process other
+%% than the one managing the tree.
+-spec update_perform(tree()) -> ok.
+update_perform(Tree=#hashtree_tree{snapshot=Snapshot}) ->
+    DirtyParents = ets:foldl(fun(DirtyKey, DirtyParents) ->
```
jtuple, line 256: The name `DirtyKey` here is unfortunate since the term is actually `{DirtyKey, SnapNode}`, and you use the name `DirtyKey` elsewhere to refer to just the bare key. Also, while we're talking about `SnapNode` you switch back and forth between `DirtyNode` and `SnapNode` throughout this module. Probably good to be consistent. Something to fix by pre3.

src/hashtree_tree.erl
```diff
@@ -0,0 +1,564 @@
...
+set_node(NodeName, Node, Tree) when is_list(NodeName) orelse NodeName =:= ?ROOT ->
+    set_node(node_key(NodeName, Tree), Node, Tree);
+set_node(NodeKey, Node, #hashtree_tree{nodes=Nodes}) when is_tuple(NodeKey) ->
+    ets:insert(Nodes, [{NodeKey, Node}]),
+    Node.
+
+%% @private
+destroy_node(Node) ->
+    Node1 = hashtree:close(Node),
+    hashtree:destroy(Node1).
```
jtuple: `eleveldb:destroy` used by `hashtree:destroy` just needs the path, the entire open/close/destroy/reopen sequence could probably just be destroy/open if you added `hashtree:destroy(Path) where is_list(Path)` to `hashtree.erl`. Unclear if that's good/bad given plans to extend hashtree to have different backends in the future, but something to consider.

src/hashtree_tree.erl
```diff
@@ -0,0 +1,564 @@
...
+    end.
+
+%% @private
+node_path({_, <<NodeInt:176/integer>>}, #hashtree_tree{data_root=DataRoot}) ->
+    NodeMD5 = riak_core_util:integer_to_list(NodeInt, 16),
+    filename:join(DataRoot, NodeMD5).
+
+%% @private
+node_key(NodeName, #hashtree_tree{id=TreeId}) ->
+    {TreeId, NodeName}.
```
jtuple, line 503: Why does the `node_key` include the `TreeId`? Isn't there an independent ETS table per `TreeId`, making this redundant? Is this to enable future changes such as mapping multiple trees to same ETS table or some other related thing? Or am I missing something?

src/riak_core_broadcast.erl
```diff
@@ -353,6 +371,62 @@ send_lazy(Peer, Messages) ->
...
+%% picks random peer favoring peers not in eager or lazy set and ensuring
+%% peer is not this node
+random_peer(Root, State=#state{all_members=All}) ->
+    Eagers = all_eager_peers(Root, State),
+    Lazys  = all_lazy_peers(Root, State),
+    Union  = ordsets:union([Eagers, Lazys]),
+    Other  = ordsets:del_element(node(), ordsets:subtract(All, Union)),
+    case ordsets:size(Other) of
+        0 ->
+            random_other_node(ordsets:del_element(node(), All));
```
jtuple, line 416: Technically, this violates the function documentation since this can be an eager/lazy peer.

jtuple:
In `riak_core_broadcast`, I wonder if we should have a debug API to list current exchanges as well as ability to cancel ongoing exchanges. Perhaps something to add by pre3.

src/riak_core_metadata.erl
```diff
@@ -156,6 +158,11 @@ itr_next({It, Opts}) ->
     It1 = riak_core_metadata_manager:iterate(It),
     {It1, Opts}.
 
+%% @doc Closes the iterator
+-spec itr_close(iterator()) -> ok.
+itr_close({It, _Ots}) ->
+    riak_core_metadata_manager:iterator_close(It).
```
jtuple, line 164: Is `itr_close` actually used anywhere? Don't see any current code/tests using it. Is this something bucket types/security should be using somewhere?

src/riak_core_metadata_hashtree.erl
```diff
@@ -0,0 +1,366 @@
...
+-spec compare(hashtree_tree:remote_fun(), hashtree_tree:handler_fun(X), X) -> X.
+compare(RemoteFun, HandlerFun, HandlerAcc) ->
+    gen_server:call(?SERVER, {compare, RemoteFun, HandlerFun, HandlerAcc}).
+
+%%%===================================================================
+%%% gen_server callbacks
+%%%===================================================================
+
+init([DataRoot]) ->
+    process_flag(trap_exit, true),
```
jtuple, line 168: Seems better to use monitors rather than trapping exits, especially since this module spawns quite a few different types of children, which have different "when this process exits, do this" semantics. Perhaps a pre3 change?

src/riak_core_metadata_hashtree.erl
```diff
@@ -0,0 +1,366 @@
...
+%% @private
+update_async(From, Lock, State=#state{tree=Tree}) ->
+    {Snap, Tree2} = hashtree_tree:update_snapshot(Tree),
+    Pid = spawn(fun() ->
+                        hashtree_tree:update_perform(Snap),
+                        case From of
+                            undefined -> ok;
+                            _ -> gen_server:reply(From, ok)
+                        end
+                end),
```
jtuple: The caller represented by `From` is monitoring the `riak_core_metadata_hashtree` process, not this spawned process. If this spawned process for some reason crashes and never sends the reply, the caller will wait indefinitely. The easiest solution would be to change this to a `spawn_link`. That's what `riak_kv_index_hashtree` [does](https://github.com/basho/riak_kv/blob/1.4.2/src/riak_kv_index_hashtree.erl#L255).

jtuple:
Lots of `gen_server:calls` in `riak_core_metadata_hashtree.erl` and `riak_core_metadata_manager.erl` that should have explicit timeouts / infinity.

src/riak_core_metadata_manager.erl
```diff
@@ -293,21 +408,55 @@ code_change(_OldVsn, State, _Extra) ->
...
     end.
 
 next_iterator(It=#metadata_iterator{done=true}) ->
     It;
+next_iterator(It=#metadata_iterator{prefix=undefined,match=undefined,tab=Tab,pos=Pos}) ->
+    next_iterator(It, ets:next(Tab, Pos));
+next_iterator(It=#metadata_iterator{prefix=undefined,pos=Pos}) ->
+    next_iterator(It, ets:select(Pos));
 next_iterator(It=#metadata_iterator{pos=Pos}) ->
     next_iterator(It, ets:match_object(Pos)).
```
jtuple, line 450: Might be helpful to have comments about the different classes of iterators (full-prefix, sub-prefix, key) based on the input arguments and how this interacts with `next_iterator/2`.

jtuple, line 450: Similar comments apply for `new_iterator` below.

jtuple:
Overall, the different iterator types in `riak_core_metadata_manager` are somewhat confusing since it all maps to the same record, the same `next_iterator` and `new_iterator` functions, etc. As noted in inline diff comments, I think adding some comments would help. Alternatively, for `new_iterator`, might even make sense to have different functions entirely: `new_prefix_iterator` or something. 

Anyway, just a minor comment. Not important for pre2.

src/riak_core_metadata_object.erl
```diff
@@ -58,6 +59,11 @@ context({metadata, Object}) ->
 -spec empty_context() -> metadata_context().
 empty_context() -> [].
 
+%% @doc returns a hash representing the metadata objects contents
+-spec hash(metadata_object()) -> binary().
+hash({metadata, Object}) ->
+    term_to_binary(erlang:phash2(term_to_binary(Object))).
```
jtuple: Probably should use something stronger than `phash2`, such as MD5 or SHA1. We really need to change K/V trees to not use `phash2` as well for 2.0, but that's a bit more complicated since we already shipped with `phash2`. Probably good idea to avoid shipping with a weak hash in the first place here in metadata land.

src/riak_core_broadcast.erl
jrwest, line 416: perhaps this function documentation is unclear. Its ok if it returns an eager/lazy peer but only if there are no other peers. In small clusters this is necessary.

src/riak_core_metadata.erl
jrwest, line 164: Yea its now required for every iterator you open (this is the sol'n for #372). Of course none of them are using it now. This was rebased on pretty recent develop but don't believe it includes security (believe that is only thing using iterators right now). Rebase again and make the changes here or seperate PR?

src/riak_core_metadata_hashtree.erl
jrwest, line 168: agreed. https://github.com/basho/riak_core/issues/389

jrwest:
replies to non-commit comments:

re: `update_snapshot/1` leak. Agreed. I originally had this in my lost version and forgot to add it the second time around. ~~will fix.~~ addressed in https://github.com/basho/riak_core/commit/27292c3b1b2e1acd60cc76cab39a15d6086839d7

re: exchange debug API. made an issue for this: https://github.com/basho/riak_core/issues/390

re: gen_server timeouts. will fix. addressed in https://github.com/basho/riak_core/commit/b5efcc52d32db0d34578fb2cbb3ebb317b8d2a19

re: iterators. will add comments. i'd like to try and keep the function names the same but if its still not so clear with comments i'm open to change. addressed in https://github.com/basho/riak_core/commit/1e7fd124e5b98ce26c0114373ec8fbb198cdf1c4

src/hashtree_tree.erl
jrwest, line 503: yea the idea was they could share ets tables and maybe, also, leveldb (or ets) backends. Not sure we will ever have use for it. Do you have a concern about needlessly copying it around right now?

src/riak_core_metadata_hashtree.erl
jrwest: will change to `spawn_link`. as mentioned in chat, trapping exists complicates things, but since we don't have a catch-all for `'EXIT'` when processes fail abnormally the affect is the same. #389 will make this better.

src/hashtree_tree.erl
jrwest, line 256: https://github.com/basho/riak_core/issues/391

src/hashtree_tree.erl
jrwest: will change. think its safe to change `hashtree` in this way. If we have pluggable backends we can just call out to `Mod:destroy` and the ets one can fail on paths. metadata hashtree would explicitly deciede to use ets, so the call could change at that point.

src/riak_core_metadata_exchange_fsm.erl
```diff
@@ -0,0 +1,295 @@
...
+    HandlerFun = fun(Diff, Acc) ->
+                         repair(Peer, Diff),
+                         track_repair(Diff, Acc)
+                 end,
+    Res = riak_core_metadata_hashtree:compare(RemoteFun, HandlerFun,
+                                              #exchange{local=0,remote=0,keys=0}),
+    #exchange{local=LocalPrefixes,
+              remote=RemotePrefixes,
+              keys=Keys} = Res,
+    lager:info("completed metadata exchange with ~p. repaired ~p missing local prefixes, "
```
jrwest: i noticed while testing changes based on review that this is really spammy. Changed to debug log when sum of repaired differences is 0 and info log otherwise.

src/riak_core_broadcast.erl
```diff
@@ -353,6 +371,62 @@ send_lazy(Peer, Messages) ->
...
+    FreeMod = lists:keyfind(Mod, 1, Exchanges) =:= false,
+    case BelowLimit and FreeMod of
+        true -> exchange(Peer, State);
+        false -> State
+    end.
+
+exchange(Peer, State=#state{mods=[Mod | Mods],exchanges=Exchanges}) ->
+    State1 = case Mod:exchange(Peer) of
+                 {ok, Pid} ->
+                     lager:info("started ~p exchange with ~p (~p)", [Mod, Peer, Pid]),
```
jrwest: also noticed this was really spammy. changing to `debug`. https://github.com/basho/riak_core/commit/0c28ac2ce12a9c533dbaf3f8f74a0f7376fdabae

src/hashtree_tree.erl
jrwest: addressed in https://github.com/basho/riak_core/commit/9f278163a27250b7b2b3721f9a27754a57bc587f

src/riak_core_metadata_hashtree.erl
jrwest: addressed in https://github.com/basho/riak_core/commit/fae31739b8c6f38d8c5376842b75848957346db3

src/riak_core_metadata_object.erl
jrwest: changed to sha in https://github.com/basho/riak_core/commit/2416eedff6f7d56ef7248c357d366e8ad18ccb67

src/riak_core_metadata_exchange_fsm.erl
jrwest: addressed in https://github.com/basho/riak_core/commit/0c28ac2ce12a9c533dbaf3f8f74a0f7376fdabae

jrwest:
pushed up the `riak_test` i've been running as one of my sanity checks before I push things up: https://github.com/basho/riak_test/pull/388. Doesn't cover very many cases but it requires exchanges to pass.

src/hashtree_tree.erl
```diff
@@ -0,0 +1,554 @@
...
+-opaque tree()         :: #hashtree_tree{}.
+-type prefix()         :: atom() | binary().
+-type prefixes()       :: [prefix()].
+-opaque tree_node()    :: prefixes() | ?ROOT.
+-type prefix_diff()    :: {missing_prefix, local | remote, prefixes()}.
+-type key_diffs()      :: {key_diffs, prefixes(),[{missing |
+                                                   remote_missing |
+                                                   different, binary()}]}.
+-type diff()           :: prefix_diff() | key_diffs().
+-opaque handler_fun(X) :: fun((diff(), X) -> X).
```
jtuple: Please change this from `-opaque` to `-type` since Dialyzer currently doesn't support polymorphic opaque types:

```
dialyzer: Analysis failed with error:
hashtree_tree.erl:286: Polymorphic opaque types not supported yet
```

jtuple:
The good news:
- Recent commits + filed issues address concerns raised in this review.
- `riak_core` eunits remain passing.
- The linked basho/riak_test#388 passes on this branch, but fails without as expected.
- The existing bucket types riak_test still passes.
- Basic manual smoke testing looks good.

Other news:
- I didn't personally test/try to break the ETS iterator stuff, other than basic usage. I know we had some concerns about this. We should probably write some tests to really exercise things.
- We need more detailed/involved test of the entire cluster metadata subsystem. We should also do coverage analysis and make sure we're hitting all the new code, including some of the corner cases we've coded for.

Things aren't dialyzer clean. Looking over the errors, I think things are just spec issues, but who knows. Please look over and make sure nothing is obviously broken. Also, create an issue to make metadata code dialyzer clean before pre3. Here's the relevant output from my run: https://gist.github.com/jtuple/6658224

All in all, good enough for pre2: Please fix the noted dialyzer opaque-type issue I mentioned in the most recent diff comment, then feel free to merge. +1.

src/hashtree_tree.erl
jrwest: addressed in https://github.com/basho/riak_core/commit/e5b3a41ae687ff6606a5177f26242ec47e9f7bc3

jrwest:
re: dialyzer https://github.com/basho/riak_core/issues/392

re: iterators, i have added a note to https://github.com/basho/riak_core/issues/373. I was planning to write an iterator_eqc, time permitting, or at a minimum a riak_test that covers all of them. The issues in https://github.com/basho/riak_core/issues/372 have been addressed in this PR (including stale iterators not crashing the metadata manager -- they will only crash the calling process).

thanks for the review. merging.

jrwest:
merged manually to resolve `rebar.config` conflict in https://github.com/basho/riak_core/commit/288bf4951ae133ca11a6e26c99f42aa15cc67cb7

PR #386, opened 2013-09-16
Add security to Riak

Vagabond:
This PR implements all the riak_core bits of security, that is, a place security information is stored and modified and APIs for checking a user has a particular set of permissions. riak_core applications are expected to use these APIs themselves, which is why there's no actual security checking done anywhere in this PR. The riak_kv and riak_api repos are where that happens.

See also basho/riak#355

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -13,5 +13,5 @@
   {riak_sysmon, ".*", {git, "git://github.com/basho/riak_sysmon", {branch, "develop"}}},
   {folsom, ".*", {git, "git://github.com/basho/folsom.git", {tag, "0.7.4p3"}}},
   {ranch, "0.4.0-p1", {git, "git://github.com/basho/ranch.git", {tag, "0.4.0-p1"}}},
-  {pbkdf2, ".*", {git, "git://github.com/basho/erlang-pbkdf2", {branch, "master"}}}
+  {pbkdf2, ".*", {git, "git://github.com/basho/erlang-pbkdf2", {branch, "adt-cleanups"}}}
 ]}.
--- a/rebar.config
+++ b/rebar.config
@@ -12,5 +12,6 @@
   {basho_stats, ".*", {git, "git://github.com/basho/basho_stats", {branch, "develop"}}},
   {riak_sysmon, ".*", {git, "git://github.com/basho/riak_sysmon", {branch, "develop"}}},
   {folsom, ".*", {git, "git://github.com/basho/folsom.git", {tag, "0.7.4p3"}}},
-  {ranch, "0.4.0-p1", {git, "git://github.com/basho/ranch.git", {tag, "0.4.0-p1"}}}
-       ]}.
+  {ranch, "0.4.0-p1", {git, "git://github.com/basho/ranch.git", {tag, "0.4.0-p1"}}},
+  {pbkdf2, ".*", {git, "git://github.com/basho/erlang-pbkdf2", {branch, "master"}}}
+]}.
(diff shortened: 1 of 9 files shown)


src/table.erl
```diff
@@ -0,0 +1,84 @@
+-module(table).
```
engelsanchez, line 1: So this is really the definitive table in the Erlang world? :)

engelsanchez, line 1: Licence header or it didn't happen

src/table.erl
```diff
@@ -0,0 +1,84 @@
...
+    lists:reverse([BottomBorder | IoList]);
+create_table(_Spec, [], _Length, IoList) ->
+    lists:reverse(IoList);
+create_table(Spec, [Row | Rows], Length, IoList) ->
+    create_table(Spec, Rows, Length, [row(Spec, Row) | IoList]).
+
+-spec get_row_length(list(tuple())) -> non_neg_integer().
+get_row_length(Spec) ->
+    lists:foldl(fun({_Name, Size}, Total) ->
+                    Total + Size + 2
```
engelsanchez, line 33: I think 7 is a prettier number

rebar.config
```diff
@@ -12,5 +12,6 @@
   {basho_stats, ".*", {git, "git://github.com/basho/basho_stats", {branch, "develop"}}},
   {riak_sysmon, ".*", {git, "git://github.com/basho/riak_sysmon", {branch, "develop"}}},
   {folsom, ".*", {git, "git://github.com/basho/folsom.git", {tag, "0.7.4p2"}}},
-  {ranch, "0.4.0-p1", {git, "git://github.com/basho/ranch.git", {tag, "0.4.0-p1"}}}
-       ]}.
+  {ranch, "0.4.0-p1", {git, "git://github.com/basho/ranch.git", {tag, "0.4.0-p1"}}},
+  {pbkdf2, ".*", {git, "git://github.com/basho/erlang-pbkdf2", {branch, "adt-cleanups"}}}
```
jaredmorrow, line 16: Mental note that we should merge adt-cleanups to master, give pdkdf2 the 'develop' treatment and then update this line.

src/riak_core.erl
```diff
@@ -350,10 +350,15 @@ register(App, [{bucket_validator, ValidationMod}|T]) ->
...
+register(App, [{permissions, Permissions}|T]) ->
+    register_mod(App, Permissions, permissions),
+    register(App, T);
+register(App, [{auth_mod, {AuthType, AuthMod}}|T]) ->
+    register_proplist({AuthType, AuthMod}, auth_mods),
     register(App, T).
 
-
-register_mod(App, Module, Type) when is_atom(Module), is_atom(Type) ->
+register_mod(App, Module, Type) when is_atom(Type) ->
```
jaredmorrow, line 361: Why this change?

Vagabond, line 361: Because when we register permissions, it is a list. not an atom.

src/riak_core_security.erl
```diff
@@ -0,0 +1,673 @@
...
+
+-record(context,
+        {username,
+         grants,
+         epoch}).
+
+prettyprint_users([all]) ->
+    "all";
+prettyprint_users(Users0) ->
+    %% my kingdom for an iolist join...
```
jaredmorrow, line 39: R17 PR's accepted

jaredmorrow:
Anything else need to be done now that we decided on R16 with conditional support for R15?

jaredmorrow:
+1 from testing on the command line the functionality and code reading with the understanding that there are some `%TODO` we need to get rid of before this hits staging.

src/table.erl
Vagabond, line 1: This will be cleaned up and renamed for pre3.

PR #387, opened 2013-09-17
Expose the resolution of capabilities for general use in protocol negotiations

buddhisthead:
The general mechanism of capabilities' preferred mode resolution can be exported to enable non-gossip usage,
such as over socket connections, e.g. protocols built on top of the connection and service managers (also in core). The first step is to refactor the resolver of preferred modes and export API such that the gen_server's state is not entangled in the core functions.

A separate PR will use the exported API to negotiate capabilities over a socket based protocol.

Tagged @jtuple for possible review as well.

### DIFF
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -182,7 +182,9 @@ update_ring(Ring) ->
             add_supported_to_ring(node(), Supported, Ring)
     end.
 
-%% @doc Make a capbility from
+%% @doc
+%% Make a capbility from a capability atom, a list of supported modes,
+%% the default mode, and a mapping from a legacy var to it's capabilities.
 -spec make_capability(capability(), [mode()], mode(), term())
                      -> {capability(), #capability{}}.
 make_capability(Capability, Supported, Default, Legacy) ->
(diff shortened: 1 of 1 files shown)


src/riak_core_capability.erl
```diff
@@ -179,6 +182,12 @@ update_ring(Ring) ->
             add_supported_to_ring(node(), Supported, Ring)
     end.
 
+%% @doc Make a capbility from
```
cmeiklejohn: Missing end of sentence?

src/riak_core_capability.erl
```diff
@@ -393,6 +402,23 @@ add_supported_to_ring(Node, Supported, Ring) ->
             {true, Ring2}
     end.
 
+%% Given my node's capabilities, my node's registered default modes, the
```
cmeiklejohn, line 408: Missing `@doc`.

src/riak_core_capability.erl
buddhisthead: Addressed by c4bfd01

src/riak_core_capability.erl
buddhisthead, line 408: Addressed by c4bfd01

cmeiklejohn:
This looks good to me, but a core/kv reviewer would be best.

slfritchie:
+1, looks like handy functions to expose to the outside world.

PR #393, opened 2013-09-22
update hashtree_eqc. get it passing on eqc 1.29.1

jrwest:
- get rid of make_treevars since the {call, erlang,'*', _} was the
  cause of unknown precondition failures and the same generation can be
  done without it
- remove initial_state function and just pass the rest of the
  initial state to commands/2
- use pretty_commands/4 so failures are easier to track down


jtuple:
Changes look good and work fine for me on EQC 1.28.2. Running via eunit seems to time out on occasion though, while you're here might want to modify `hashtree_test_` to have slighter wider gap between eunit timeout (currently 30s) and test duration (currently 29s).

Otherwise, +1 merge.

PR #394, opened 2013-09-23
Fix rebar.config to use {branch, "name"}

engelsanchez:
The rebar.config file currently is using "master" by itself, which does
not make rebar actually update that branch on update-deps.


Vagabond:
+1

Issue #392, opened 2013-09-22
Make Cluster Metadata Dialyzer Clean


jrwest:
see https://gist.github.com/jtuple/6658224

jrwest:
fixed in #397

PR #397, opened 2013-09-24
dialyzer fixes for cluster metadata

jrwest:
addresses issue #392
- fix some typespecs in dvvset, riak_core_metadata and metadata manager
- change some opaque types to regular ones
- don't reuse `remote_iterator` as message to manager process since its
  a record too
- fix typo in exchange_fsm that didn't harm anything

This PR does not fix the issues found by dialyzer in
riak_core_metadata_exchange_fsm (see [here](https://gist.github.com/jrwest/760d9934c5d64d30252e)). Dialyzer isn't able to figure
out we are doing the "right thing" with the iterators here.

### DIFF
--- a/src/dvvset.erl
+++ b/src/dvvset.erl
@@ -82,7 +82,7 @@
 
 %% @doc Constructs a new clock set without causal history,
 %% and receives a list of values that gos to the anonymous list.
--spec new([value()]) -> clock().
+-spec new(value() | [value()]) -> clock().
 new(Vs) when is_list(Vs) -> {[], Vs};
 new(V) -> {[], [V]}.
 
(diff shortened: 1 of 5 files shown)


beerriot:
Commit e6fff5b addresses the issue about riak_core_metadata_exchange_fsm mentioned in the PR description. Dialyzer was correct that a `metadata_context()` was not being passed. The atom `undefined` is passed instead. The value in question is ignored, so `any()` would have been as correct as anything, but preserving the information about what would normally be expected in that tuple seemed appropriate.

beerriot:
Verified that dialyzer errors are fixed, and no new ones are added. Also verified that `cluster_meta_basic` test still passes.

Looks good. +1

PR #398, opened 2013-09-24
Remove merkle

rzezeski:


slfritchie:
+1, nuke it

ancheremukhin:
What do you think about merkle tree?
Is it worse than hash-tree implementation.
Have you removed merkle because it is only on RAM?

rzezeski:
@ancheremukhin This module was no longer used by any of the Riak subsystems and we have a new merkle tree implementation in [hashtree.erl](https://github.com/basho/riak_core/blob/develop/src/hashtree.erl).

PR #400, opened 2013-09-27
Fixes #189 -- include socket peer info in lager messages

### LINKED ISSUE #189: Log handoff sender on node receiving handoff
Right now, we currently do not log which node sent the handoff on the receiving node:

```
2012-06-12 19:32:01.295 [info] <0.28019.343>@riak_core_handoff_receiver:handle_info:71 Handoff receiver for partition 35681192317648997026457149236237378409568665600 exited after processing 299 objects
```

It would be useful to know which node was sending the partition on the receiving node.

slfritchie:
Subject says it all ... the handoff protocol doesn't include the node name of the source node, so the best we can do without a protocol change (and capability negotiation and ...) is to log the socket's peername() info.

### DIFF
--- a/src/riak_core_handoff_receiver.erl
+++ b/src/riak_core_handoff_receiver.erl
@@ -30,6 +30,7 @@
          terminate/2, code_change/3]).
 
 -record(state, {sock :: port(),
+                peer :: term(),
                 ssl_opts :: [] | list(),
                 tcp_mod :: atom(),
                 timeout_len :: non_neg_integer(),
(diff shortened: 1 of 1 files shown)


engelsanchez:
Given all the talk about less erlangy log messages lately, I was wondering if it would be OK to display the peername as ip:port instead of directly printing the {IP, Port} tuple.  Or maybe [ip]:port to be more IPV6 friendly. Although I'm not sure how IPV6 addresses are represented here.

seancribbs:
@engelsanchez Agree, `inet_parse:ntoa/1` works nicely for that purpose.

engelsanchez:
Nice, that's perfect. Thanks @seancribbs

engelsanchez:
Looks good. A bit of manual shows it working and showing the remote host and port on the receiving part. Caveat: I did not try it with SSL.

:+1: and merging now so it makes it into pre3

PR #401, opened 2013-09-27
Fix #263 (riak_core should not directly reference `riak-admin` commands in output)

### LINKED ISSUE #263: riak_core should not directly reference `riak-admin` commands in output
https://github.com/basho/riak_core/blob/master/src/riak_core_console.erl#L179 and 9 other console output messages reference `riak-admin`. This is pretty confusing if you make a riak_core app that has nothing to do with riak, and try and use one of the console commands.

slfritchie:
Stolen (poorly) from GNU gettext.

### DIFF
--- a/src/riak_core_console.erl
+++ b/src/riak_core_console.erl
@@ -25,6 +25,7 @@
          clear_staged/1, transfer_limit/1, pending_claim_percentage/2,
          transfers/1, add_user/1, add_source/1, grant/1, revoke/1,
          print_users/1, print_user/1, print_sources/1]).
+-export([extra_text/1]).                        % For extra info strings
 
 %% @doc Return for a given ring and node, percentage currently owned and
 %% anticipated after the transitions have been completed.
@@ -185,9 +186,11 @@ unreachable_status(Down) ->
     io:format("WARNING: The cluster state will not converge until all nodes~n"
               "are up. Once the above nodes come back online, convergence~n"
               "will continue. If the outages are long-term or permanent, you~n"
-              "can either mark the nodes as down (riak-admin down NODE) or~n"
-              "forcibly remove the nodes from the cluster (riak-admin~n"
-              "force-remove NODE) to allow the remaining nodes to settle.~n"),
+              "can either mark the nodes as down ~sor~n"
+              "forcibly remove the nodes from the cluster~n"
+              "~sto allow the remaining nodes to settle.~n",
+              [extra_text_cb(unreachable_status1),
+               extra_text_cb(unreachable_status2)]),
     ok.
 
 %% Provide a list of nodes with pending partition transfers (i.e. any secondary vnodes)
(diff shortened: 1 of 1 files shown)


slfritchie:
Looking for review: @engelsanchez @engelsanchez @jtuple @jrwest or other?

jrwest:
confirmed this is already on the kill bill marked for pending review. no one has claimed yet

src/riak_core_console.erl
```diff
@@ -956,3 +957,57 @@ parse_cidr(CIDR) ->
...
+
+%% For riak_core users outside of Basho's Riak KV application: If you
+%% define an OTP environment variable ExtraKey (see below) for the
+%% riak_core application, for example my_module, then any extra/custom
+%% text formatting text can be returned by calling the callback
+%% function my_module:extra_text/1 ... stolen (poorly) from GNU
+%% gettext.
+
+extra_text_cb(Label) ->
+    case app_helper:get_env(riak_kv) of
```
jrwest, line 969: we shouldn't be referencing `riak_kv` in `riak_core`. Besides tests the one place we do this is `hashtree.erl` and that is to support rolling upgrades from previous versions of riak where `hashtree.erl` was in `riak_kv`. It seems like this function should just look for a "registered" `extra_text` callback mod, which `riak_kv_app` or something should register.

jrwest:
I'm ok w/ the general idea here but I think the callback can be simpler. The name of the `-admin` script is  what we care about parameterizing over so why not just have applications indicate a script name, e.g. "riak-admin" or "my-admin" or whatever?

I suppose some applications could change the names of the `-admin` commands. For the sake of simplicity, however, I think it would be sufficient in a first pass to ignore this case. If we don't want to, then indicating the name of the `-admin` script and the name of the command being paired w/ the call in `riak_core_console` seems to cover all cases and is a bit more structured than the current approach in this PR.

Overall, I think we can probably get away w/ leaving this PR for a future cycle if there is not time to make changes. There are many other obstacles to writing `riak_core` apps outside of Riak that are larger than this one. Perhaps we can revisit this in a larger push to address those obstacles?

jrwest:
@slfritchie I didn't see any response to my comment above so I'm going to assume this isn't making it for 2.0. Did you want to leave it open to continue the discussion?

src/riak_core_console.erl
slfritchie, line 969: I don't understand "shouldn't be referencing riak_kv" ... there gotta be some way of figuring out if this riak_core app is bundled with riak_kv.  I'm peeking at OTP env vars to see if we're bundled -- oughtta work inside or outside Riak, am I forgetting something?

slfritchie:
This PR goes back to the refrigerator for pondering later....

jrwest:
The refrigerator contains things in the `2.1` milestone. marked as such.

jrwest:
NOTE: original issue can be found at https://github.com/basho/riak_core/issues/263

Issue #376, opened 2013-09-09
Support Creation Semantics Needed by Bucket Types


jrwest:
The bucket types implementation added in #357 does not properly enforce the expected creation semantics  -- that it cannot result in siblings. This is because #363 only includes eventually consistent writes. Creation of bucket types must be conflict-free to ensure there are no ambiguities when features like strong-consistency and CRDTs use them to specify the types of data stored under buckets in a type.  

Some (hand-wavey) proposals that have been thrown out:
1. serialize through claimaint
2. some sort of "seen" set overlaid on cluster metadata
3. use strongly consistent store that will be added to riak core as part of strong consistency in riak work

1 & 3 seem to be the most likely and easiest to implement candidates.

In addition to the work in cluster metadata, bucket types should be converted to using the changes before closing this issue.

jrwest:
a first pass has been implemented here https://github.com/basho/riak_core/commit/b4d9525b97c499f8b2397cdfeb3550318d82dac1

jrwest:
a second pass has been implemented here: https://github.com/basho/riak_core/tree/jrw-bucket-type-invariant

jrwest:
also related https://github.com/basho/riak_kv/pull/673 and https://github.com/basho/riak_test/pull/394

jrwest:
Also related and can be reviewed together w/ the above PRs (although they don't necessarily need to be): https://github.com/basho/riak_kv/pull/677 https://github.com/basho/riak/pull/395

jrwest:
The web of PRs above can be a bit hard to follow, so here is some more detail:

#402 is the basis for these changes and must be reviewed along-side https://github.com/basho/riak_kv/pull/673. The bucket types test has been updated as well in https://github.com/basho/riak_test/pull/394. This should help w/ review of the previous two PRs. In addition riak-admin support has been added. This is done in https://github.com/basho/riak_kv/pull/677 and https://github.com/basho/riak/pull/395 which also need to be reviewed together but may be separate from the previous group of PRs.

Vagabond:
Can we close this now?

jrwest:
ah yes, sorry, got lost in the shuffle.

PR #402, opened 2013-09-30
Proper Bucket Type Creation

jrwest:
Features that plan to use bucket types require that certain properties be "create-only" -- they cannot change after data has been written to buckets under the type. To ensure these semantics it is important that nodes see either no "create-only" properties for a type, or only one version of the "create-only" properties. Since types are stored in Cluster Metadata, the current implementation does not guarantee this. To ensure that it does, we take the following two steps to handle concurrent writers:
1. all writes are serialized through the current claimant
2. to handle the case where the claimant changes over before properties have had a chance to fully propogate, creation is a two step process. Only when all nodes agree can the second step be reached, and nodes do not see the properties (while handling requests, etc) until the second step is complete.

This, also, allows for the "create-only" properties to be changed between the first and second steps, referred to as "creation" and "activation". This is useful in the case where a user accidentally specifies the wrong value for a "create-only" property. After activating the type, all properties that are not "create-only" may continue to be updated. More details can be found in the module edoc for `riak_core_bucket_type`.

riak_core does not enforce the notion of "create-only" types. This PR includes an extension to riak_core's bucket validators (changing the interface to validator modules). Validation functions are now passed an atom indicating if the properties are being validated are for bucket type creation, or are type or bucket properties for updating. In addition, the function is passed the type and bucket, as well as the old properties if they existed. Validators can use this information to enforce "create-only" types.

Lastly, this PR makes a minor change to `riak_core_bucket_props` to use the `max` function in `resolve` -- as suggested somewhere during review of an earlier PR.

NOTE: currently this PR does not include very comprehensive tests. I've begun an eqc test but that may have to wait until pre4. The bucket types riak_test has been updated (https://github.com/basho/riak_test/pull/394) and should pass

### DIFF
--- a/src/riak_core_bucket_type.erl
+++ b/src/riak_core_bucket_type.erl
@@ -17,22 +17,96 @@
 %% under the License.
 %%
 %% -------------------------------------------------------------------
+
+%% @doc Bucket Types allow groups of buckets to share configuration
+%% details.  Each bucket belongs to a type and inherits its
+%% properties. Buckets can override the properties they inherit using
+%% {@link riak_core_bucket}.  The "Default Bucket Type" always
+%% exists. The Default Type's properties come from the riak_core
+%% `default_bucket_props' application config, so the Default Type and
+%% its buckets continue to act as they had prior to the existence of
+%% Bucket Types.
+%%
+%% Unlike Buckets, Bucket Types must be explicitly created. In
+%% addition, types support setting some properties only on creation
+%% (via {@link riak_core_bucket_props:validate/4}). Since, types are
+%% stored using {@link riak_core_metadata}, in order to provide safe
+%% creation semantics the following invariant must be satisfied: all
+%% nodes in the cluster either see no type or a single version of the
+%% type for the lifetime of the cluster (nodes need not see the single
+%% version at the same time). As part of ensuring this invariant, creation
+%% is a two-step process:
+%%
+%%   1. The type is created and is inactive. To the node an inactive type
+%%      does not exist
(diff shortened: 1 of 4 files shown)


jrwest:
this addresses issue #376

src/riak_core_bucket_type.erl
```diff
@@ -45,63 +119,59 @@ defaults() ->
...
 get(BucketType) when is_binary(BucketType) ->
-    get(BucketType, undefined).
-
-%% @private
-get(BucketType, Default) ->
-    riak_core_metadata:get(?KEY_PREFIX,
-                           BucketType,
-                           [{default, Default}, {resolver, fun riak_core_bucket_props:resolve/2}]).
-
+    riak_core_claimant:get_bucket_type(BucketType, undefined).
```
Vagabond, line 161: The comment above says that get() is the only operation NOT serialized through the claimant, yet here we are?

jrwest, line 161: not serialized through the claimant _process_ (maybe I should add that word?). All the code is in the claimant module so the defines don't need to be copied or moved to a .hrl

Vagabond, line 161: Oh never mind, it done in the claimant module, but in the current process (why?).

jrwest, line 161: so the defines don't need to be moved around/copied and all the riak_core_metadata calls are in one module

Vagabond:
The code for this looks fine. I would like to see more tests around the progression from creation to activated, but hopefully we can address that in pre4. +1

PR #406, opened 2013-10-01
export index type from chash

beerriot:
riak_pipe wants to use the `chash:index()` type: https://github.com/basho/riak_pipe/blob/develop/src/riak_pipe_vnode.erl#L69

Can it be exported, please?

### DIFF
--- a/src/chash.erl
+++ b/src/chash.erl
@@ -50,7 +50,7 @@
          successors/3,
          update/3]).
 
--export_type([chash/0]).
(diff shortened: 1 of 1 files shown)


jrwest:
There are a few places where I think this type is recreated in the core source that it would be nice to fix but no need to do that here or spread it to pipe. Actually, this fixes an unknown type error, woohoo! +1 to merge.

PR #407, opened 2013-10-02
Remove riak_core_gen_server (gen_server2)

rzezeski:
This version of gen_server2 is over 3 years old.  There have been at
least 15 comments to the canonical OTP gen_server in the last 3 years.
The longer we stay with this outdated copy of gen_server the higher
the probability of hitting a bug that is fixed in OTP.  We also miss
out on an performance benefits that might be made in OTP.  Speaking of
perf, there is no evidence that gen_server2 helps us at all.  It's
original purpose was to avoid lots of redundant work when doing
selective receive on large mailboxes.  Only the handoff receiver is
still using this behavior and it has an ack-based model to prevent
overflow.  We have already moved all other processes off of this
behavior and it has been agreed, twice over, by the majority of the
engineering team to remove gen_server2 completely.


slfritchie:
+1, DIAF.

bookshelfdave:
in use for v2 replication (pre-BNW)

Vagabond:
We never used any of the special queing features in Repl that gen_server2 provided, so it was literally a drop-in replacement for gen_server with some slightly different internal behaviour around mailbox handling.

If the repl team really want to keep it out of paranoia, I guess that is fine, but +1 to removing it from everything else.

rzezeski:
Setting to milestone 2.1 so this can be hashed out later. Perhaps, to play nice with people using riak core outside of Riak, it might be nicer to deprecate this module and then remove it in future 2.x or 3.0 release. But, none of our code should use it.

jrwest:
leaving at milestone `2.1` for same reasons @rzezeski outlined 2 months ago

PR #410, opened 2013-10-04
export vsn of riak_core_ring:future_index that doesn't take ring

jrwest:


engelsanchez:
Looks good, tests still pass and they have coverage for the code because it's a simple refactor :+1:

PR #411, opened 2013-10-04
Sort AAE differences before acting upon them via read-repair

slfritchie:
Apply patch to hashtree.erl that was developed for
https://github.com/basho/riak_kv/pull/646.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -90,14 +90,14 @@
 %% to LevelDB using a single batch write. Writes are flushed whenever the
 %% buffer becomes full, as well as before updating the hashtree.
 %%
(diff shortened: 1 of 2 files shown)


engelsanchez:
The patch removed hashtree:compare/3 in favor of hashtree:compare/4 (yay!), but did not modify the code using hashtree:compare/3. I've pushed that in  1240324.

The change looks good, eunit tests pass as before (I saw an anomaly in core_vnode_eqc that I had seen this morning in develop).  I've tested this with basho/riak_kv#646 and aae repairs are indeed happening in sort order

:+1: :dancer: :boat: 

I'm going to merge this myself so it makes it into pre3

PR #413, opened 2013-10-04
Fix eunit test failures

engelsanchez:
- Scott's accidental cast to infinity (and beyond!!)
- core_vnode_eqc was failing due to meck unload module on process death
  behavior, removed with no_link flag


jrwest:
+1. the case obv. needs to be fixed. the other change doesn't seem to have an effect on my machine but it doesn't cause problems either.

PR #414, opened 2013-10-04
Increase disterl buffer sizes.

evanmcc:
At this point in riak's history disterl is our primary bottleneck.
Something about the way that it is framing small messages is causing its
tcp connection to size its send and recv buffers much smaller than
is ideal.

This change adds code to automatically set dist port sizes on nodeup,
as well as code to manually set all disterl connections to specific
sizes, meant to be used from the console for testing and tuning.


evanmcc:
distbufs-base-pre3base-uniform-fixed-write_heavy-binsize1024-2013-10-4-20-9-51vsdistbufs-pre3bufs-uniform-fixed-write_heavy-binsize1024-2013-10-4-18-55-38-summary
This is the perf run.  I am told that the baseline run (on the left) for develop may be especially slow because of a known regression.

evanmcc:
distbufs-base142-1 4 2-uniform-fixed-write_heavy-binsize1024-2013-10-5-14-48-28vsdistbufs-pre3bufs-uniform-fixed-write_heavy-binsize1024-2013-10-4-18-55-38-summary

1.4.2 doesn't fare a ton better.

ebin/riak_core.app
```diff
@@ -32,6 +32,8 @@
              riak_core_console,
              riak_core_coverage_fsm,
              riak_core_coverage_plan,
+             riak_core_dist_mon,
+             riak_core_dist_mon_sup,
```
slfritchie: I'm curious about why add a supervisor.  Do you have plans to add additional worker processes?

evanmcc: I didn't know whether or not I needed to. I know Erlang the language OK at this point, but the proper OTP idioms are still somewhat opaque to me.  Suggestions as to how to properly structure a piece of code like this are welcomed.

slfritchie:
I pushed a couple of small refactoring patches, if you like.  They compile without errors.  :-)

evanmcc:
updated

ebin/riak_core.app
```diff
@@ -146,6 +147,10 @@
 
          %% Handoff IP/port
          {handoff_port, 8099},
-         {handoff_ip, "0.0.0.0"}
+         {handoff_ip, "0.0.0.0"},
+         
+         %% Disterl buffer sizes
+         {dist_send_buf_size, 393216},
```
Vagabond: What unit are these in and why were these particular numbers chosen? Can we record that for posterity here in this file?

evanmcc: bytes, scattershot empirical testing/what worked on my machines.  More science could likely be applied.  I'll add some comments.

src/riak_core_sup.erl
```diff
@@ -46,10 +48,14 @@ start_link() ->
 %% ===================================================================
 
 init([]) ->
+    DistMonEnabled = app_helper:get_env(riak_core, disable_dist_mon,
+                                        false) == false,
     Children = lists:flatten(
                  [?CHILD(riak_core_sysmon_minder, worker),
                   ?CHILD(riak_core_vnode_sup, supervisor, 305000),
                   ?CHILD(riak_core_eventhandler_sup, supervisor),
+                  ?IF(DistMonEnabled,
```
Vagabond: You could do [?CHILD(riak_core_dist_mon, worker) || DistMonEnabled] instead of making a macro....

Vagabond:
Forgive my ignorance, but how does this interact with the zddbl flag, if at all?

src/riak_core_sup.erl
evanmcc: I pulled the macro from riak_kv_sup.  I didn't realize that list comprehensions could be bent that way.

evanmcc:
My understanding is that they're entirely separate.  I'll look into it.

Vagabond:
I does seem to be independant, yes.

src/riak_core_sup.erl
```diff
@@ -46,10 +48,14 @@ start_link() ->
 %% ===================================================================
 
 init([]) ->
+    DistMonEnabled = app_helper:get_env(riak_core, disable_dist_mon,
```
Vagabond: Backwards flags like this are hard to understand, why not use enable_dist_mon?

Vagabond:
I'm seeing some odd results in theshell:

```
(dev1@127.0.0.1)13> net_kernel:monitor_nodes(true, [{node_type, visible}, nodedown_reason]).
ok
(dev1@127.0.0.1)14> nodes().
[]
(dev1@127.0.0.1)15> flush()
Shell got {nodedown,'dev2@127.0.0.1',
                    [{nodedown_reason,connection_closed},{node_type,visible}]}
Shell got {nodeup,'dev2@127.0.0.1',[{node_type,visible}]}
ok
(dev1@127.0.0.1)16> nodes().
['dev2@127.0.0.1']
(dev1@127.0.0.1)17> [{Name, inet:getopts(Port, [sndbuf, recbuf])} || {Name, Port} <- erlang:system_info(dist_ctrl)].
[{'dev2@127.0.0.1',{ok,[{sndbuf,425984},{recbuf,425984}]}},
 {'c_29911_dev1@127.0.0.1',{ok,[{sndbuf,663750},
                                {recbuf,262030}]}}]
(dev1@127.0.0.1)18> riak_core_dist_mon:get_riak_env_vars().
{393216,786432}
```

Vagabond:
Oh, nevermind. net.core.[rw]mem_max was screwing me. Once I bumped that to 6mb the values reported were esactly double what was set in the config file.

Other than the previous comments, this seems to work fine.

+1

src/riak_core_sup.erl
evanmcc: My thinking was that since it's enabled by default, one would need to disable it, but it does look really ugly.

PR #415, opened 2013-10-09
Add infrastructure to change the OTP net_kernel's net_ticktime

slfritchie:
Changing the net_ticktime on a single node is easy: it's a single
function call.  However, in a cluster environment, that function
must be run on all nodes simultaneously, including any new nodes
that connect to the cluster during the time change ... and that
includes any hidden nodes.

In Riak's case, those hidden nodes are usually `nodetool`, which
is an escript.  Those scripts don't have access to `riak_core_util.erl`
or to any other module module that a fun sent over-the-network
might refer to.  So there's this icky dance of checking the remote
node's code server path and appending to the code path if the
riak_core dir is missing.

There is additional icky spawns and timeout checking and exit()
use to avoid: 1. being blocked by the local net_kernel when trying
to use `rpc:call/4` or `spawn/2` to a remote node that's down/slow,
and 2. leaking zillions of processes during such net_kernel instability.

Yes, there's an accompanying riak_test test.

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -58,7 +58,9 @@
          make_fold_req/1,
          make_fold_req/2,
          make_fold_req/4,
-         make_newest_fold_req/1
+         make_newest_fold_req/1,
+         start_set_net_ticktime_daemon/2,
+         stop_set_net_ticktime_daemon/1
         ]).
 
 -include("riak_core_vnode.hrl").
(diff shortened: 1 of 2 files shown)


slfritchie:
It hits me now that perhaps these new functions should be moved to a new module?  Then that beam file could be added to the `basho-patches` dir of older Riak packages and migrate tick_time on older Riak clusters.  Thoughts?

slfritchie:
Why do you want to change net_ticktime?

For all those times when the Erlang/OTP distributed Erlang default timeout of 60 seconds is wrong, but you don't want to shut down your entire cluster to change the tick time (by editing vm.args (or whatever Cuttlefish config thingie)) and restarting ... but you also want to do it in such a way that won't interfere with nodes being added & removed while the tick time is being changed, and you don't want to interfere with "riak-admin" commands and other "hidden" nodes that might wink in and out of the cluster without warning.

src/riak_core_util.erl
```diff
@@ -581,6 +583,100 @@ make_fold_reqv(v2, FoldFun, Acc0, Forwardable, Opts)
...
+
+start_set_net_ticktime_daemon(Node, Time) ->
+    EbinDir = filename:dirname(code:which(?MODULE)),
+    try
+        Dirs = rpc:call(Node, code, get_path, []),
+        case lists:member(EbinDir, Dirs) of
+            false ->
+                lager:info("start_set_net_ticktime_daemon: adding to code path "
+                           "for node ~p\n", [Node]),
+                rpc:call(Node, code, add_pathz, [EbinDir]);
```
jrwest: my tired brain wonders if there is ever a case where the node we are calling to is not on the same physical machine as the node doing the call, in which case the path may not be valid -- i can't think of a case where `nodetool` connects to a remote node, however. Could we perform some additional check? Would that additional check be worth it?

jrwest: my less tired brain wonders about mixed clusters where (even if we move this to a module) the path may exist but the code may not be there. 

Before we make available the beams on nodes that don't have them we could check if `riak_core_ring_manager` is defined and running, but if we've already started migrating net_ticktime we're out of luck (unless we actually load the module on the node remotely?). If the first daemon process is always run on a real riak node perhaps we could use a capability to pre-empt this (there would still be a small window for error as caps converge on downgrade?).

jrwest: This sort of illustrates a benign version of what I'm talking about. Here the `riak_test` node doesn't have the module so the node which has the first daemon run passes its ebin on it `riak_test` which then passes it on to every other node in the cluster despite them already having the module. Not sure how big of a deal this is since its just extra paths to search when code is loaded. {shrug}

```
12:42:12.014 [info] start_set_net_ticktime_daemon: adding to code path for node 'dev1@127.0.0.1': "/home/jordan/sandbox/basho/rt/riak/current/dev/dev1/lib/riak_core-1.4.2-253-gda2cb42/ebin"
12:42:12.014 [info] start_set_net_ticktime_daemon: adding to code path for node 'dev2@127.0.0.1': "/home/jordan/sandbox/basho/rt/riak/current/dev/dev1/lib/riak_core-1.4.2-253-gda2cb42/ebin"
12:42:12.014 [info] start_set_net_ticktime_daemon: adding to code path for node 'dev3@127.0.0.1': "/home/jordan/sandbox/basho/rt/riak/current/dev/dev1/lib/riak_core-1.4.2-253-gda2cb42/ebin"
12:42:12.014 [info] start_set_net_ticktime_daemon: adding to code path for node 'dev4@127.0.0.1': "/home/jordan/sandbox/basho/rt/riak/current/dev/dev1/lib/riak_core-1.4.2-253-gda2cb42/ebin"
```

jrwest:
To make sure I understand what you  envision as the process, if I were a user planning to change the `net_ticktime` I would:
1. use whatever method i use to change the vm.args/riak.conf/whatever to set net_ticktime in future restarts (no restart is actually required here) across all nodes
2. attach to a node, and run `start_set_net_ticktime_daemon` w/ the same time as the one specified in the new configs
3. PROFIT

yes?

jrwest:
Also, yes it would be nice to move this to a separate module so we can load it on older clusters.

src/riak_core_util.erl
slfritchie: Yes, I've noticed that too, but it doesn't happen in the common case when the Riak nodes are on separate machines.

AFAIK, nodetool isn't talking other other nodes in this case.

slfritchie:
> To make sure I understand what you envision as the process, [...]

Yup.

src/riak_core_util.erl
jrwest: Ok. For the mixed cluster case where the module may not be available/in basho-patches what do you think? 

It seems like we can guard it by a capability since it will always first be executed on a Riak node  

Not sure about the requirements of having that done for merge though given the state of mixed clusters right now...

slfritchie: I am perfectly ok with saying that you can't do it in a mixed cluster: you have enough craziness in your life.  Settle down, then change the damn tick.
- Riak 2.0: use it as-is
- Riak 1.x: put the module in basho-patches, but compile the silly thing with R15B0x first.

jrwest: I agree but I think our default position is to protect the user who doesn't fully read the docs, which can be done, to a degree, w/ a capability

jrwest:
I'm going to go ahead and give this a +1 since we are running out of time today. We should either address the capability comment I made or open an issue for it. I'm fine with the latter since we need to add capabilities for many new things in 2.0.

PR #416, opened 2013-10-09
move ebin/riak_core.app to src/riak_core.app.src

vinoski:
Move the ebin/riak_core.app file to src/riak_core.app.src for easier maintenance. Also adjust .gitignore to now ignore the ebin dir and contents.


seancribbs:
I'm :+1: but I think we should have someone from core/kv confirm.

vinoski:
FWIW I manually diffed the generated `riak_core.app` against the original, and found no substantive differences, and I also verified that the tests work the same as they do on the `develop` branch.

jtuple:
Already made this change on the consistency branch and was going to have it merge in during that PR. No reason not to just go ahead and do change this now. +1 from me

PR #417, opened 2013-10-09
refactored riak_core bits of the schema in here, with tests

joedevivo:
Although the rebar.config specifies `cuttlefish:develop`, it will need jd-test-tools to run until https://github.com/basho/cuttlefish/pull/42 is merged. If you do review this, be a dear and review that too? :)

I've refactored the riak_core bits of the schema into the riak_core project for easier testing and less duplication of schema files between the riak and riak-ee repos..

### DIFF
--- a/test/riak_core_schema_tests.erl
+++ b/test/riak_core_schema_tests.erl
@@ -1,0 +1,68 @@
+-module(riak_core_schema_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+-compile(export_all).
+
+basic_schema_test() ->
+    Config = cuttlefish_unit:generate_templated_config("../priv/riak_core.schema", [], context()),
+
+    cuttlefish_unit:assert_config(Config, "riak_core.ring_creation_size", 64),
+    cuttlefish_unit:assert_config(Config, "riak_core.ring_state_dir", "./ring"),
+    cuttlefish_unit:assert_config(Config, "riak_core.handoff_port", 8099 ),
+    cuttlefish_unit:assert_config(Config,"riak_core.dtrace_support", false),
+    cuttlefish_unit:assert_config(Config, "riak_core.platform_bin_dir",  "./bin"),
+    cuttlefish_unit:assert_config(Config, "riak_core.platform_data_dir", "./data"),
+    cuttlefish_unit:assert_config(Config, "riak_core.platform_etc_dir",  "./etc"),
+    cuttlefish_unit:assert_config(Config, "riak_core.platform_lib_dir",  "./lib"),
+    cuttlefish_unit:assert_config(Config, "riak_core.platform_log_dir",  "./log"),
+    ok.
+
+override_schema_test() ->
+    Conf = [
+        {["ring_size"], 8},
+        {["ring", "state_dir"], "/absolute/ring"},
+        {["handoff", "port"], 8888},
+        {["dtrace"], on},
+        %% Platform-specific installation paths (substituted by rebar)
+        {["platform_bin_dir"], "/absolute/bin"},
+        {["platform_data_dir"],"/absolute/data" },
+        {["platform_etc_dir"], "/absolute/etc"},
+        {["platform_lib_dir"], "/absolute/lib"},
+        {["platform_log_dir"], "/absolute/log"},
+
+        %% Optional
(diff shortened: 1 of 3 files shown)


test/riak_core_schema_tests.erl
```diff
@@ -0,0 +1,68 @@
+-module(riak_core_schema_tests).
+
+-include_lib("eunit/include/eunit.hrl").
```
jrwest, line 3: missing an `-ifdef(TEST)`

test/riak_core_schema_tests.erl
```diff
@@ -0,0 +1,68 @@
...
+-compile(export_all).
+
+basic_schema_test() ->
+    Config = cuttlefish_unit:generate_templated_config("../priv/riak_core.schema", [], context()),
```
jrwest, line 10: maybe put a comment here saying something like "look, this is where all the defaults we are testing are coming from!". Always, nice to know what other files to look at in a test.

test/riak_core_schema_tests.erl
joedevivo, line 3: I thought we didn't bother with `-ifdef(TEST)` for modules in the test dir

test/riak_core_schema_tests.erl
```diff
@@ -0,0 +1,68 @@
...
+    cuttlefish_unit:assert_config(Config, "riak_core.platform_lib_dir", "/absolute/lib"),
+    cuttlefish_unit:assert_config(Config, "riak_core.platform_log_dir", "/absolute/log"),
+    cuttlefish_unit:assert_config(Config, "riak_core.ssl.certfile", "/absolute/etc/cert.pem"),
+    cuttlefish_unit:assert_config(Config, "riak_core.ssl.keyfile", "/absolute/etc/key.pem"),
+    cuttlefish_unit:assert_config(Config, "riak_core.ssl.cacertfile", "/absolute/etc/cacertfile.pem"),
+    cuttlefish_unit:assert_config(Config, "riak_core.handoff_ssl_options.certfile", "/tmp/erlserver.pem"),
+    cuttlefish_unit:assert_config(Config, "riak_core.handoff_ssl_options.keyfile", "/tmp/erlkey/pem"),
+    ok.
+
+context() ->
```
jrwest, line 68: maybe a comment about what this is for the test? maybe its just my lack of familiarity w/ cuttlefish source idk. My assumption is these are the "template vars" that are passed in during generation (of riak.conf?).

test/riak_core_schema_tests.erl
```diff
@@ -0,0 +1,68 @@
...
+    ok.
+
+override_schema_test() ->
+    Conf = [
+        {["ring_size"], 8},
+        {["ring", "state_dir"], "/absolute/ring"},
+        {["handoff", "port"], 8888},
+        {["dtrace"], on},
+        %% Platform-specific installation paths (substituted by rebar)
+        {["platform_bin_dir"], "/absolute/bin"},
```
jrwest, line 32: we don't do it in a lot of our tests, but personal opinion, you've used these strings twice, they should be in variables.

test/riak_core_schema_tests.erl
jrwest, line 3: hrm i guess its mixed, some do some dont [1] [2]

[1] https://github.com/basho/riak_core/blob/develop/test/hashtree_eqc.erl#L4
[2] https://github.com/basho/riak_core/blob/develop/test/riak_core_service_mgr_tests.erl#L23

I'd vote for having it but don't really care that much

test/riak_core_schema_tests.erl
joedevivo, line 68: Good point. It's not that, so I definitely should.

It's actually the variables that rebar will use during the templating phase of the riak build.

test/riak_core_schema_tests.erl
joedevivo, line 32: for production code, I'm with you.

with tests, I really like only having to look at one line to figure out what's being checked. also personal opinion

jrwest:
Going to go ahead w/ the +1 now since any changes are just comments and maybe an ifdef. Tests run as part of `make test` and pass for me. Schema looks good. I mentioned to @joedevivo elsewhere that there are some `riak_core` application config vars we probably want to add the to schema, but those can be added separately (we should open an issue though).

+1 merge.

PR #422, opened 2013-10-14
Dialyzer fixes for the connection manager.

cmeiklejohn:
Some minor fixes for the connection manager found via dialyzer.  

Also upgrades rebar to support debug_info via protobuffs.


slfritchie:
+1, thanks!

cmeiklejohn:
Will reapply these in another branch, since the connection manager moved out of riak core.

PR #423, opened 2013-10-14
fill in missing props in riak_core_bucket_type:defaults/0

jrwest:
defaults taken from https://github.com/basho/riak_kv/blob/develop/src/riak_kv_app.erl#L83-L97

cc/ @Vagabond @seancribbs @broach @jonmeredith @metadave

### DIFF
--- a/src/riak_core_bucket_type.erl
+++ b/src/riak_core_bucket_type.erl
@@ -112,12 +112,25 @@
 %% @doc The hardcoded defaults for all bucket types.
 -spec defaults() -> bucket_type_props().
 defaults() ->
-      [{n_val,3},
-       {allow_mult,false},
-       {last_write_wins,false},
-       {precommit, []},
-       {postcommit, []},
-       {chash_keyfun, {riak_core_util, chash_std_keyfun}}].
+    [{linkfun, {modfun, riak_kv_wm_link_walker, mapreduce_linkfun}},
+     {old_vclock, 86400},
+     {young_vclock, 20},
+     {big_vclock, 50},
+     {small_vclock, 50},
+     {pr, 0},
+     {r, quorum},
+     {w, quorum},
+     {pw, 0},
+     {dw, quorum},
+     {rw, quorum},
+     {basic_quorum, false},
+     {notfound_ok, true},
+     {n_val,3},
+     {allow_mult,false},
+     {last_write_wins,false},
+     {precommit, []},
+     {postcommit, []},
+     {chash_keyfun, {riak_core_util, chash_std_keyfun}}].
 
 %% @doc Create the type. The type is not activated (available to nodes) at this time. This
 %% function may be called arbitratily many times if the claimant does not change between
--- a/src/riak_core_bucket_type.erl
+++ b/src/riak_core_bucket_type.erl
@@ -126,7 +126,7 @@ defaults() ->
      {basic_quorum, false},
      {notfound_ok, true},
      {n_val,3},
-     {allow_mult,false},
+     {allow_mult, true},
      {last_write_wins,false},
      {precommit, []},
      {postcommit, []},


jrwest:
also pinging @rzezeski @coderoshi for yokozuna. Want to make sure there are no other defaults that need be in this list. 

---

if they are not in this list, they must always be specified when a user creates a bucket type -- or they could set it on update depending on validation. in addition, properties that are missing from this list cannot be reset w/ `riak_core_bucket_type:reset/1`.

For completeness, more details on create/update of bucket types here: https://github.com/basho/riak_core/blob/develop/src/riak_core_bucket_type.erl

rzezeski:
AFAIK this should work for both Riak Search and Yokozuna.

RS uses `search` bucket property which will trigger bucket fixups to add/remove from pre-commit list.

YZ currently uses `yz_index` property to determine if to index and which index to use.  This could change but if it does then the issue can be revisited.

coderoshi:
yz_index isn't required to start, but you may want it to reset. It's a tough question, based on the usecase of reset. If it should convert a bucket completely back to its initial state, I say add it. If a user should call reset and reasonably assume that their bucket values are still indexed, leave it off.

rzezeski:
Well Yokozuna also respects the sentinel value `_dont_index_` as an indication not to index.  So that can be used as a "reset".  Ideally we have an API that hides this as it is an implementation detail.  The API of yokozuna is assoc/disassocan index with a bucket.

f we add `yz_index` to the default list then it returns for every bucket whether or not Yokozuna is enabled.  Are we okay with that?

jrwest:
following https://github.com/basho/riak_core/commit/3de2ca207ea2fe0878292c499187670fe50cee6c I've changed `allow_mult` to `true` as part of this PR

jrwest:
@rzezeski so bucket types currently knows nothing about bucket fixups....I've been wondering about this.

jrwest:
also, it sounds like the `_dont_index_` sentinel should maybe be in this list. let me know and I can add it (or feel free to push to this PR).

Vagabond:
I'm a little concerned that this sort of removes the ability for riak_core apps to add things to the 'defaults'. In our discussions the other dai I imagined 2 inheritance chains:

Buckets under default type:
Registered default properties -> App.config properties ->Bucket properties

And for a regular bucket type:
Registered default properties -> Bucket type properties -> Bucket properties

If we literally just hardcode everything into core, then we preclude other core applications inserting their own defaults without also patching core, which is not reasonable.

rzezeski:
`_dont_index_` is a value of a bucket property `yz_index` to indicate not to index.  Are you saying you want this added to default list?  You want every user, even when Yokozuna is off, to see this?

Maybe it should be.  Or maybe Yokozuna shouldn't use bucket props.  Perhaps it should just offer an API to assoc/disassoc an index with a bucket and store that somewhere else.  It's a semantic only relevant to Yokozuna, so perhaps it should be solely contained in its API.

jrwest:
@rzezeski i don't want anything ;). I think the problem is what @Vagabond pointed out  and I agree w/ his points (I don't like this PR) -- even tho yoko is not a core app it still applies.

Vagabond:
Is yokozuna opt-in or opt-out for indexing? If it is opt-in surely the lack of a yz_index should mean "don't index this"?

jrwest:
@Vagabond one question/clarification: are the "Registered default properties" the same list in your example? iirc the problem people had w/ this is that changing the list in one hierarchy used to affect not only new but old buckets whereas changing it in the bucket-type hierarchy wont. Would you be opposed to them being registered separately (i know its kind of ugly but avoids the ambiguity). e.g. `append_bucket_defaults/1` and `append_bucket_type_defaults/1`.

rzezeski:
Yokozuna is opt-in.  If there is no `yz_index` property it takes that as "don't index this bucket".  If you opt-in then now you have this bucket property.  If you want to stop indexing then you set the value to `_dont_index_`.  All of these are implementation details.  It just so happens to use bucket properties right now.

Vagabond:
@jrwest I think they should be the same list. That list should be _only_ the defaults for what the core apps _need_ to operate. Id you want to change them for default buckets, edit the app.config, if you want to change them for a particular type, configure the type, and of course you can still set them per-bucket.

I don't understand the ambiguity you refer to,

jrwest:
@Vagabond the ambiguity I believe people were referring to was if I append a **new** property to that list all previously created buckets in the default type will see that change [1] whereas no other bucket-types/buckets will. Idk if I feel this is ambiguous but it was discussed and was the source for the suggestion of having the lists be separate iirc.

[1 ]https://github.com/basho/riak_core/blob/develop/src/riak_core_ring_manager.erl#L558-L564

Vagabond:
@jrwest I think if a core app registers a new prop, it should be see by buckets under both default and custom types. As I said above the "Registered default properties" should be the root of the inheritance tree in both cases and should be tracked seperately from crap added in app.config, which only applies to the default bucket type.

jrwest:
@Vagabond the problem w/ that, which I think @jtuple brought up during our mumble, is if we apply something like bucket fixups to existing types we may invalidate the creation invariant. (am i wrong in thinking this is basically what you are suggesting since thats how what i was describing works now?). We don't really know until we add new properties but I thought when we discussed this we decided to be cautious and not modify existing types or buckets underneath them.

Vagabond:
OK, I conceed the point. Types should be created with a snapshot of the default properties at their creation time and new properties should not be inherited from the registered defaults if later versions of riak register them. This is a stronger invariant than just not changing existing properties via inheritance and makes future proofing easier.

As an interim fix for pre4, I think we should merge this to make the Java client happy.

+1 to merge and fix correctly later.

jrwest:
Opened an issue to track the improvements mentioned https://github.com/basho/riak_core/issues/434

Merging.

PR #425, opened 2013-10-14
GET ON THIS LEVEL

joedevivo:
@jrwest @seancribbs 

moved default bucket props to {level, advanced}


jrwest:
verified no default bucket props config lines end up in `riak.conf` (they do in pre4 tag) and that adding some to `riak.conf` actually changes things. +1.

joedevivo:
I love that this didn't need a separate riak and riak_ee review! :)

PR #430, opened 2013-10-17
Fix bad type specifications.

cmeiklejohn:
max_nb_const should be widened to include 'undefined', considering how
all callers into this function only ever pass undefined at this point.

Also, define a new 'target' type which is used, because currently, the
type is declared as string which is incorrect and triggering dialyzer
type warnings anywhere this type is referenced.


slfritchie:
+1, thanks!

Issue #409, opened 2013-10-03
Avoid gen server call in metadata gets


engelsanchez:
The performance hit has not been quantified, but it should be non-trivial in the critical paths where metadata will be used.  Jordan says that there is no current reason to do lookups through the gen server anyway.

PR #431, opened 2013-10-17
do not serialize through gen_server on metadata get

jrwest:
addresses #409 

The reason we serialized is historical and no longer applies (and was going to be removed).

https://github.com/basho/riak_test/pull/388 should still pass w/ this change.

cc/ @engelsanchez

### DIFF
--- a/src/riak_core_metadata_manager.erl
+++ b/src/riak_core_metadata_manager.erl
@@ -122,10 +122,12 @@ start_link(Opts) ->
 -spec get(metadata_pkey()) -> metadata_object() | undefined.
 get({{Prefix, SubPrefix}, _Key}=PKey) when (is_binary(Prefix) orelse is_atom(Prefix)) andalso
                                            (is_binary(SubPrefix) orelse is_atom(SubPrefix)) ->
-    gen_server:call(?SERVER, {get, PKey}, infinity).
+    read(PKey).
 
 %% @doc Same as get/1 but reads the value from `Node'
 -spec get(node(), metadata_pkey()) -> metadata_object() | undefined.
+get(Node, PKey) when node() =:= Node ->
+    ?MODULE:get(PKey);
 get(Node, {{Prefix, SubPrefix}, _Key}=PKey)
   when (is_binary(Prefix) orelse is_atom(Prefix)) andalso
        (is_binary(SubPrefix) orelse is_atom(SubPrefix)) ->


evanmcc:
So, measuring this, I wasn't actually able to produce any slowdown.  In fact, the change actually seems to slow down the low-contention case.  My current guess is that the ets lookup penalty being assesed to the FSM process and causing a schedule-out sooner is more expensive than talking to some other process and having it take the reduction bump. The total amount of work seems to be about the same.  Once you get up to as many processes as will run in parallel, the performance is the same.

Performance without this PR:

```
(perfdev@127.0.0.1)1> client:preload().
[ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,
 ok,ok,ok,ok,ok,ok,ok,ok,ok,ok|...]
(perfdev@127.0.0.1)2> client:start(1,20).
did 4675 gets/s over 20 seconds 
did 49652 cs/s over 20 seconds 
ok
(perfdev@127.0.0.1)3> client:start(4,20).
did 16367 gets/s over 20 seconds 
did 170225 cs/s over 20 seconds 
ok
(perfdev@127.0.0.1)4> client:start(12,20).
did 34158 gets/s over 20 seconds 
did 347494 cs/s over 20 seconds 
ok
(perfdev@127.0.0.1)5> client:start(128,20).
did 55904 gets/s over 20 seconds 
did 429208 cs/s over 20 seconds 
ok
```

Performance with this PR:

```
(perfdev@127.0.0.1)2> client:preload().
[ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,ok,
 ok,ok,ok,ok,ok,ok,ok,ok,ok,ok|...]
(perfdev@127.0.0.1)5> client:start(1,20).
did 3321 gets/s over 20 seconds 
did 36633 cs/s over 20 seconds 
ok
(perfdev@127.0.0.1)6> client:start(4,20).
did 15362 gets/s over 20 seconds 
did 158483 cs/s over 20 seconds 
ok
(perfdev@127.0.0.1)7> client:start(12,20).
did 28805 gets/s over 20 seconds 
did 283240 cs/s over 20 seconds 
ok
(perfdev@127.0.0.1)3> client:start(128,20).
did 56467 gets/s over 20 seconds 
did 432389 cs/s over 20 seconds 
ok
```

evanmcc:
Even weirder:

I merged an incomplete patch I have that lessens the work that the get FSM does, and all of the sudden this patch has approximately no performance effect.  Guessing that the current FSM is very near to another schedule-out on an average runthrough, in terms of reductions.

Since I plan to merge the above code later, I don't think that this matters either a lot way.  :+1: to merge if you'd like, otherwise leaving it out is also fine.

jrwest:
went ahead and merged based on the last comment. If it turns out to be an issue its easy to change back.

PR #432, opened 2013-10-17
Make load_certs function exported and more useful (works on a single file, too)

Vagabond:
Related to basho/riak_api#39


cmeiklejohn:
:+1:

Issue #428, opened 2013-10-15
Share LevelDB Instance Between Cluster Metadata Hashtrees


jrwest:
Currently, cluster metadata uses a different segment path per prefix when creating hashtrees resulting in the opening of many leveldb instances. Long term we don't want these hashtrees to be backed by leveldb at all (they don't need to be persistent) but for now we should at least share the database between them to not open dbs unbounded.

jrwest:
fixed by #433 closing

PR #433, opened 2013-10-17
share segment store accross all nodes in hashtree_tree

jrwest:
this ensures that we don't create an unbounded number of segment
stores (leveldb databases), instead only creating one.

addresses #428 

exports `hashtree:new/3` so that each level can still pass different size options

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -103,6 +103,7 @@
 
 -export([new/0,
          new/2,
+         new/3,
          insert/3,
          insert/4,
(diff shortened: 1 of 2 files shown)


cmeiklejohn:
`cluster_meta_basic` is failing on this branch for me.

jrwest:
rebased and force-pushed on most recent `develop` which should address the test failures

cmeiklejohn:
:+1: looks good, merge away!

PR #435, opened 2013-10-21
Rework table formatting to be better and output better info

Vagabond:
Basicially tables now auto-size much better and handle multiline fields
much better, without breaking the table. The original table.erl was
something Andrew Stone had lying around, so it wasn't particularly
polished.

cc @jaredmorrow hoping he can review it before Wednesday ;)


Vagabond:
Oh, ccing @andrewjstone is probably a good idea, too.

Vagabond:
It can now print tables like:

```
+----------+----------+----------------------------------------+
|   type   |  bucket  |                 grants                 |
+----------+----------+----------------------------------------+
|    *     |    *     |           riak_kv.mapreduce            |
| default  | mybucket |      riak_kv.get, riak_kv.delete,      |
|          |          |              riak_kv.put               |
+----------+----------+----------------------------------------+
```

where even though the permission list stretches to the next line (word wrapped before being fed into the table), the table formatter pads the other fields to include blank cells to compensate.

jaredmorrow:
I'll take a look.

jaredmorrow:
Table formatting looks better, +1

PR #436, opened 2013-10-22
2.0 version of handoff backwards compatability

evanmcc:
add a simple, point to point capability check for batch message support. if not, fall back to per-object messages.


engelsanchez:
I gave the code a cursory review, as it has been reviewed before and this is just a port.  The handoff mixed riak test now passes for legacy and previous configurations without the unknown msg making handoffs fail anymore.

:+1: :dancer: :boat:

PR #437, opened 2013-10-23
Implement riak security enable/disable/status

Vagabond:
Implemented in terms of a cluster_metadata modulated capability.

I'm not actually sure this yields the right semantics. On the one hand it makes sure that the status of security across the cluster is consistent, but on the other it means that if you join one old (or malicious) node to the cluster, your security goes bye-bye.

### DIFF
--- a/src/riak_core_security.erl
+++ b/src/riak_core_security.erl
@@ -26,7 +26,7 @@
 -export([authenticate/3, add_user/2, alter_user/2, del_user/1,
          add_source/4, del_source/2,
          add_grant/3, add_revoke/3, check_permission/2, check_permissions/2,
-         get_username/1, is_enabled/0,
+         get_username/1, is_enabled/0, enable/0, disable/0, status/0,
          get_ciphers/0, set_ciphers/1, print_ciphers/0]).
 %% TODO add rm_source, API to deactivate/remove users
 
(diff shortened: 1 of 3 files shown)


src/riak_core_security.erl
```diff
@@ -366,9 +366,46 @@ add_source(User, CIDR, Source, Options) ->
     add_source([User], CIDR, Source, Options).
 
 is_enabled() ->
-    %% TODO this should be some kind of capability or cluster-wide config
-    app_helper:get_env(riak_core, security, false).
+    case riak_core_capability:get({riak_core, security}) of
+        true ->
+           case  riak_core_metadata:get({<<"security">>, <<"status">>},
```
engelsanchez, line 510: Why not atoms instead of binaries?

engelsanchez:
Ok, so we have a switch but nobody is using it to avoid the security checks, etc?

src/riak_core_security.erl
Vagabond, line 510: Because binaries are used everywhere else in security because cluster_metadata didn't use to support atoms.

No better reason.

Vagabond:
On Wed, Oct 23, 2013 at 07:07:24AM -0700, Engel A. Sanchez wrote:

> Ok, so we have a switch but nobody is using it to avoid the security checks, etc?

I don't know what this means. I updated the is_available function to
read the value out of the cluster metadata...

engelsanchez:
Oops. You mean is_enabled(). I missed that, nevermind.

engelsanchez:
The code looks good. There is a pending discussion to follow up regarding cluster metadata and its convergence properties here https://github.com/basho/riak_core/issues/440.

We should open an issue to track what we can do to improve the situation with PB requests, which only check if security is enabled/disabled on connection creation. Connections tend to stick around for a long time, so the switch may not have a visible effect for a long time, or not on all requests.

With those caveats, I've tested the basic functionally and is as advertised.

:+1:

slfritchie:
Ping?  I see a +1 from @engelsanchez, is there any other reason to wait, aside from ETOOBUSY from @Vagabond?

Vagabond:
Rebased.

Vagabond:
I'm going to merge this, including the test that nobody has been able to review because of breakage on develop.

The reasons I'm doing so are the following:
- Both code PR have beem +1ed
- The failing test can be addressed in the bugfixing phase

There is going to be one regression after we merge, riak_search may service requests over a secured connection if the node started up with security disabled but it was enabled at runtime (or if the cluster was in mixed mode with an older release without the security capability). The riak_test tests for this, so it will break once we fix the other issues with develop.

cc @jaredmorrow

PR #441, opened 2013-10-24
Support for strongly consistent Riak

jtuple:
This is the `riak_core` related pull-request for on-going work to add strong consistency to Riak.

First, this pull-request integrates Riak with [riak_ensemble](http://github.com/basho/riak_ensemble), pulling in `riak_ensemble` as an included-app dependency, hooking up `riak_ensemble_sup` into the `riak_core_sup`, and bootstrapping the root ensemble (root consensus group).

Second, this pull-request fixes some issues/extends the handoff/forwarding API exposed by `riak_core_vnode` to implementing vnode modules. This is needed in [riak_kv](http://github.com/basho/riak_kv) to properly extend `riak_kv_vnode` for strong consistency.

See commit messages for more details.

**Note**: The `riak_ensemble` subsystem is disabled by default. Changing `enable_consensus` to true in top-level `riak.conf` or by defining the `riak_core/enable_consensus` environment variable is necessary for this system to be activated on start-up. When disabled, most of the code in this pull-request is intentionally not executed.

Sibling pull-requests: basho/riak_kv#710, basho/riak_ensemble#1


src/riak_core_claimant.erl
```diff
@@ -625,13 +626,20 @@ schedule_tick() ->
...
-    State.
+tick(State=#state{last_ring_id=LastID}) ->
+    case riak_core_ring_manager:get_ring_id() of
+        LastID ->
+            schedule_tick(),
+            State;
+        RingID ->
+            {ok, Ring} = riak_core_ring_manager:get_raw_ring(),
+            maybe_bootstrap_root_ensemble(Ring),
+            maybe_force_ring_update(Ring),
```
jrwest, line 637: @jtuple this caught my attention because we no longer run `maybe_force_ring_update` in some cases. From my reading, these cases would always result in `maybe_force_ring_update` doing nothing, so its safe and one case I know hits this code path, ring resizing on a single node, acts the same w/ or w/o this change. Just wanted to double check.

jrwest:
Focused on the 3 areas that have changed when `consensus_enabled =false` (claimant tick, core_vnode, and core_sup). Made one comment but its a question for @jtuple that I think will result in no changes. `make test` passes and I've run several different riak tests, including `verify_dynamic_ring` & `verify_build_cluster` which cover the changed code paths. To run the tests I was using `jdb-strong-2.0` in kv and `jdb-wip-2.0` in ensemble but I have not looked at the code in those two branches.

+1 re: `consensus_enabled=false`

jtuple:
As discussed internally, going ahead and merging with the understanding that strong consistency and the consensus subsystem are disabled by default. Likewise, updated `rebar.config` to point directly to the work-in-progress feature branch of `riak_ensemble` that has not yet been fully/reviewed and merged. This is sufficient to build our upcoming tech preview.

A full review of basho/riak_ensemble#1 is forth coming, and will involve a more detailed review of the code paths in this PR that are hit when the consensus system is enabled.

PR #445, opened 2013-11-04
initial add of riak_core_bucket_types.hrl

bookshelfdave:


src/riak_core_bucket_type.erl
```diff
@@ -89,6 +89,8 @@
 %%      by most riak_core applications and is considered acceptable (so dont do it!).
 -module(riak_core_bucket_type).
 
+-include_lib("riak_core_bucket_type.hrl").
```
jrwest: we seem to usually use `include` in these cases (e.g. https://github.com/basho/riak_core/blob/develop/src/riak_core_metadata_hashtree.erl#L42)

bookshelfdave: fixed

jrwest:
+1. code changes look fine and I quickly built and ran a rel node w/ this branch

PR #446, opened 2013-11-05
cleanup conn_mgr/service_mgr eunit output

bookshelfdave:
See the docs on the NODEBUG macro [here](http://www.erlang.org/doc/apps/eunit/chapter.html#Debugging_macros)


evanmcc:
Test output is much cleaner, :+1:

PR #447, opened 2013-11-06
Removed webmachine from applications that should be started.

andreashasse:
This line becomes a problem when building a release with the newest version of riak_core without having webmachine in the project.


seancribbs:
:+1: Thanks!

PR #451, opened 2013-11-11
Implement alter-user, del-user and del-source

Vagabond:
Previously to this, you could add users and sources, but they were effectively immutable because I had not implemented API for changing them. alter-source does not exist because add-source will effectively do what you want. We could add a synonym at the riak-admin level if we want to.

Also, fixes a bug in print-users I found before RICON that didn't make it into the tech preview.

### DIFF
--- a/src/riak_core_console.erl
+++ b/src/riak_core_console.erl
@@ -824,21 +824,31 @@ check_limit(Str) ->
     end.
 
 add_user([Username|Options]) ->
-    case riak_core_security:add_user(list_to_binary(Username),
+    try riak_core_security:add_user(list_to_binary(Username),
                                      parse_options(Options)) of
         ok -> ok;
         Error ->
             io:format("~p~n", [Error]),
             Error
+    catch
+        throw:{error, {invalid_option, Option}} ->
+            io:format("Invalid option ~p, options are of the form key=value~n",
+                      [Option]),
+            error
     end.
 
 alter_user([Username|Options]) ->
-    case riak_core_security:alter_user(list_to_binary(Username),
-                                     parse_options(Options)) of
+    try riak_core_security:alter_user(list_to_binary(Username),
+                                       parse_options(Options)) of
         ok -> ok;
         Error ->
             io:format("~p~n", [Error]),
             Error
+    catch
+        throw:{error, {invalid_option, Option}} ->
+            io:format("Invalid option ~p, options are of the form key=value~n",
+                      [Option]),
+            error
     end.
 
 del_user([Username]) ->
@@ -856,7 +866,7 @@ add_source([Users, CIDR, Source | Options]) ->
         Other ->
             [list_to_binary(O) || O <- Other]
     end,
-    case riak_core_security:add_source(Unames, parse_cidr(CIDR),
+    try riak_core_security:add_source(Unames, parse_cidr(CIDR),
                                   list_to_atom(Source),
                                   parse_options(Options)) of
         ok ->
@@ -864,6 +874,11 @@ add_source([Users, CIDR, Source | Options]) ->
         Error ->
             io:format("~p~n", [Error]),
             Error
+    catch
+        throw:{error, {invalid_option, Option}} ->
+            io:format("Invalid option ~p, options are of the form key=value~n",
+                      [Option]),
+            error
     end.
 
 del_source([Users, CIDR]) ->
--- a/src/riak_core_console.erl
+++ b/src/riak_core_console.erl
@@ -23,7 +23,8 @@
          stage_leave/1, stage_remove/1, stage_replace/1, stage_resize_ring/1,
          stage_force_replace/1, print_staged/1, commit_staged/1,
          clear_staged/1, transfer_limit/1, pending_claim_percentage/2,
-         transfers/1, add_user/1, add_source/1, grant/1, revoke/1,
+         transfers/1, add_user/1, alter_user/1, add_source/1,
+         del_source/1, grant/1, revoke/1,
          print_users/1, print_user/1, print_sources/1, ciphers/1]).
 
 %% @doc Return for a given ring and node, percentage currently owned and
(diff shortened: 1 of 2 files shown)


src/riak_core_security.erl
```diff
@@ -500,10 +594,15 @@ accumulate_grants([], Seen, Acc) ->
     {Acc, Seen};
 accumulate_grants([Role|Roles], Seen, Acc) ->
     Options = riak_core_metadata:get({<<"security">>, <<"roles">>}, Role),
-    NestedRoles = [R || R <- lookup("roles", Options), not lists:member(R, Seen)],
+    NestedRoles = [R || R <- lookup("roles", Options),
+                        not lists:member(R,Seen),
+                        user_exists(R)],
+    io:format("Nested roles ~p~n", [NestedRoles]),
```
slfritchie: Debugging leftover?

slfritchie:
I've given this a quick read.  It appears sane.  Didn't run it.  +1 for what that's worth, sorry, I've other PRs to drive-by now on this iteration.........

src/riak_core_console.erl
```diff
@@ -831,6 +832,23 @@ add_user([Username|Options]) ->
             Error
     end.
 
+alter_user([Username|Options]) ->
+    case riak_core_security:alter_user(list_to_binary(Username),
+                                     parse_options(Options)) of
```
evanmcc: whitespace?

src/riak_core_security.erl
```diff
@@ -282,6 +291,72 @@ add_user(Username, Options) ->
...
+                                            riak_core_metadata:delete({<<"security">>,
+                                                                       <<"grants">>},
+                                                                       Key),
+                                            Acc
+                                    end, undefined,
+                                    {<<"security">> ,<<"grants">>},
+                                    [{match, {Username, '_'}}]),
+            %% delete the user out of any other user's 'roles' option
+            %% this is kind of a pain, as we have to iterate ALL roles
+            riak_core_metadata:fold(fun({_, [?TOMBSTONE]}, Acc) ->
```
evanmcc: might be nice to factor this deeply-nested case out into a few functions.

evanmcc:
the whole system would more discoverable, I guess, if there were also a print-roles command.

evanmcc:
this PR fails to updated the help text for riak-admin secuity

evanmcc:
nevermind, found the accompanying riak PR

evanmcc:
exceptions, like invalid_option, thrown when processing console commands, aren't caught, so you get no notification (other than a 1 shell return code) that you're having a problem or doing something wrong. e.g.:

```
12:07:27 <<0.1113.1>> {riak_core_console,parse_options,
                          [["password","asdkljasdk"],[]]}

12:07:27 <<0.1113.1>> {riak_core_console,parse_options,2} -> {throw,
                                                              {error,
                                                               {invalid_option,
                                                                "password"}}}
```

evanmcc:
It looks like user-specific sources should be found and deleted when the user is deleted

``` bash
evan@chain:~/src/vagabond/dev/dev1$ ./bin/riak-admin security del-user user 
evan@chain:~/src/vagabond/dev/dev1$ riak-admin security add-source all 0.0.0.0/0 trust
evan@chain:~/src/vagabond/dev/dev1$ ./bin/riak-admin security print-sources
+--------------------+------------+----------+----------+
|       users        |    cidr    |  source  | options  |
+--------------------+------------+----------+----------+
|        user        |127.0.0.1/32| password |    []    |
|        all         | 0.0.0.0/0  |  trust   |    []    |
+--------------------+------------+----------+----------+
```

Additionally, setting an all source again seems to break printing:  Looks like `all` is becoming an atom when it's updated from the command line, and only the list case is handled.

Kind of a complicated setup, not sure how reproducible it'll be:
1. build a cluster with these changes
2. run the http_security riak-test
3. after it fails: ./bin/riak-admin security del-user user 
4. ./bin/riak-admin security add-source all 0.0.0.0/0 trust 
5. ./bin/riak-admin security add-source user 0.0.0.0/0 trust  <-- suspect this isn't actually needed
6. ./bin/riak-admin security print-sources

I see this:

```
evan@chain:~/src/vagabond/dev/dev1$ ./bin/riak-admin security print-sources
RPC to 'dev1@127.0.0.1' failed: {'EXIT',
                                 {badarg,
                                  [{erlang,binary_to_list,[all],[]},
                                   {riak_core_security,
                                    '-prettyprint_users/2-lc$^0/1-0-',1,
                                    [{file,"src/riak_core_security.erl"},
                                     {line,73}]},
                                   {riak_core_security,prettyprint_users,2,
                                    [{file,"src/riak_core_security.erl"},
                                     {line,73}]},
                                   {riak_core_security,
                                    '-print_sources/1-lc$^0/1-0-',1,
                                    [{file,"src/riak_core_security.erl"},
                                     {line,88}]},
                                   {riak_core_security,
                                    '-print_sources/1-lc$^0/1-0-',1,
                                    [{file,"src/riak_core_security.erl"},
                                     {line,90}]},
                                   {riak_core_security,print_sources,1,
                                    [{file,"src/riak_core_security.erl"},
                                     {line,88}]},
```

Lastly, I am likely doing something wrong, but changing the `all` role to `trust` doesn't seem to allow unauthenticated users to access the cluster after it has been changed.

evanmcc:
With those things said, deletion and altering of users seem to work as advertised, sources are now deletable and re-addable, so I think that this is almost ready to go.

Vagabond:
What would print-roles do that print-user(s) does not?

evanmcc:
I don't know, I couldn't really figure out roles, what they are, how
they're created, etc.  Orthogonal to this PR in any case, just a comment as
I was first reviewing, mostly expressing that confusion.

Vagabond:
I've addressed all the concerns raised, except there is a bug with the interaction of a source for 'all' and a source for particular users. If they overlap you'll get the error you mentioned above.

As discussed with Evan, as this bug was not added by this PR, we will fike a seperate issue and resolve it independantly.

evanmcc:
+1, changes look good.

PR #452, opened 2013-11-13
Supports riak_kv changes for riak_kv#679

russelldb:
Some additions to vclock API to support creating fresh vclocks with a given timestamp.

A change to riak_core_util so that opening a non-existent file returns an error rather than crashing on a badmatch.

### DIFF
--- a/src/vclock.erl
+++ b/src/vclock.erl
@@ -1,6 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%% riak_core: Core Riak Application
+%% Riak_Core: Core Riak Application
 %%
 %% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
 %%
@@ -34,6 +34,7 @@
 -export([fresh/0,descends/2,merge/1,get_counter/2,get_timestamp/2,
 	increment/2,increment/3,all_nodes/1,equal/2,prune/3,timestamp/0]).
 -export([fresh/2]).
+-export([fresh/3]).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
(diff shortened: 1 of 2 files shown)


src/vclock.erl
```diff
@@ -59,6 +60,11 @@ fresh() ->
 fresh(Node, Count) ->
     [{Node, {Count, timestamp()}}].
 
+-spec fresh(vclock_node(), counter(), timestamp()) ->
+                   vclock().
+fresh(Node, Count, Ts) ->
+    [{Node, {Count, Ts}}].
```
evanmcc, line 67: do we want to do some structural validation here?  or will a bogus timestamp value immediately cause something else to crash?  what I am thinking of here is bad timestamps getting stored and then coming back later to explode in a manner hard to link to this change.  I am likely overthinking it.

russelldb, line 67: No, not overthinking, I guess we should make sure it is a pos_integer().

seancribbs, line 67: Guards would seem appropriate in this case.

russelldb, line 67: Sure, I guess because it is not enforced at `increment/3` I didn't think to do it here. But I will

slfritchie:
Ping?

russelldb:
@slfritchie parked while we figure out what to do for riak_kv#679

jrwest:
marked for milestone `2.1` since this all seems to be baking. someone more familiar @russelldb @slfritchie, can correct me, if need be.

russelldb:
yeah, work restarted on this today. 2.1 seems right. Thanks @jrwest

slfritchie:
+1

russelldb:
Out of date. No longer needed, pertinent changes to riak_core_util merged already. Sorry all. Killing.

PR #453, opened 2013-11-18
add missing apps to src/riak_kv.app.src

vinoski:
The basho_stats, eleveldb, pbkdf2 and poolboy apps are used in riak_core
but were not listed as application dependencies in the
src/riak_core.app.src file.


cmeiklejohn:
:+1: to merge.

Issue #450, opened 2013-11-11
Metadata Folds Should Close Iterator


jrwest:
As of https://github.com/basho/riak_core/commit/0b59e4cd30ffedd29cea592f2edb347f41095019 metadata iterators must always be closed.  `riak_core_metadata:fold/3,4` were written prior to this they and were not updated to account for the change.

jrwest:
Addressed by #457

Issue #443, opened 2013-10-28
metadata ETS tables should set read_concurrency and write_concurrency to true


evanmcc:
Otherwise lock contention will make riak sad.

jrwest:
Addressed by #457

Issue #440, opened 2013-10-24
Provide Option to Resolve but Not Update on Metadata Read


jrwest:
There is a case that exists in EC stores when writing with `w =1` that could persist indefinitely under the right conditions. Although for most, uses hitting this case is unlikely, providing the option to not write back the resolved value on read would be a good addition because cluster metadata reads are often performed on individual requests in Riak.

cc/ @engelsanchez

jrwest:
Addressed by #457

Issue #438, opened 2013-10-23
Change Initial Eager Set Size for Larger Clusters


jrwest:
Testing has shown that a larger initial size for the eager sets improves certain metrics in larger clusters. The static choice of the initial size for clusters greater than five nodes should be improved to choose the initial size based on the cluster size (perhaps ln(n) + 1?).

jrwest:
Addressed by #457

PR #457, opened 2013-11-21
Yet Another Round of Cluster Metadata Improvements (YAROCMI #1)

jrwest:
This PR includes several minor changes to Cluster Metadata to address outstanding issues:
- #450 
- #440 
- #438 
- #443 

In addition, this PR adds two additional functions to `riak_core_metadata`. `to_list/1,2` are primarily useful for testing and debugging.

The commits contain details about the motivations for each change but I've put some extra notes below.
### Increasing Eager Set Size for Larger Clusters

Prior to these changes the initial size of the eager peer set for all nodes had a maximum of two. This increased the chances of relying on Cluster Metadata's anti-entropy mechanism. Ideally, we want to reduce this chance as much as possible. These changes increase that maximum depending on the size of the cluster. In addition, as the number of nodes in the cluster increases or decreases, because the tree is re-initialized at each membership change, it will be re-optimized for the new size. 

The following table shows the average percentage of failed nodes before an exchange is needed for 10, 20 and 40 nodes. These numbers were obtained by running a controlled test against the actual implementation.

<table>
<tr>
<th># nodes</th>
<th>before</th>
<th>after</th>
</tr>
<tr>
<td>10</td>
<td>64%</td>
<td>85%</td>
</tr>
<tr>
<td>20</td>
<td>44%</td>
<td>78%</td>
</tr>
<tr>
<td>40</td>
<td>40%</td>
<td>76%</td>
</tr>
</table>

### allow_put

What this option does is described in the commit and edocs added as part of it. However, the motivation for it is roughly explored [here](https://gist.github.com/jrwest/f8c0d49174f4db1c4c88)
## Other Notes for Review

https://github.com/basho/riak_test/pull/450 may be useful for testing these changes and verifying the properties of the larger eager set sizes (although you will need to be able to support running a large number of nodes on one machine to test the same sizes)


src/riak_core_metadata.erl
```diff
@@ -313,19 +344,24 @@ current_context(PKey) ->
...
     RValue = riak_core_metadata_object:value(Reconciled),
-    Stored = riak_core_metadata_manager:put(PKey, RContext, RValue),
-    broadcast(PKey, Stored),
+    case AllowPut of
+        false ->
+            ok;
+        _ ->
+            Stored = riak_core_metadata_manager:put(PKey, RContext, RValue),
+            broadcast(PKey, Stored)
+    end,
```
jtuple, line 364: Have any opinion on going with `true` for the second clause rather than `_`? Yes, `_` let's us not crash on unexpected values. On the other hand, it makes it easy to not notice an accidental typo such as `fallse`. Thoughts?

src/riak_core_broadcast.erl
```diff
@@ -609,11 +609,18 @@ init_peers(Members) ->
...
             InitEagers = orddict:fetch(node(), Tree),
-            InitLazys  = [lists:nth(random:uniform(N - 3), Members -- [node() | InitEagers])]
+            InitLazys  = [lists:nth(random:uniform(N - 3), Members -- [node() | InitEagers])];
+        N ->
+            %% 10 or more members, use a tree similar to riak_core_gossip
+            %% but with higher fanout (larger initial eager set size)
+            NEagers = round(math:log(N)+1),
+            Tree = riak_core_util:build_tree(NEagers, Members, [cycles]),
+            InitEagers = orddict:fetch(node(), Tree),
+            InitLazys  = [lists:nth(random:uniform(N - (NEagers +1)), Members -- [node() | InitEagers])]
```
jtuple: Yeah, yeah, complaining about whitespace. But, `(NEagers +1)` should probably be `(NEagers+1)` or `(NEagers + 1)`.

jtuple:
All code changes / commits look good. Ran both `cluster_meta_basic` and `cluster_meta_rmr` from basho/riak_test#450 and both tests passed (after the minor change mentioned in the riak_test PR for the rmr test).

Still think #440 is better solved long-term by smarter merging that carries forward the semantics of merging rather than looking identical to an independent write (eg. using CRDT-like logic). But, the `allow_put` approach is good enough for now.

Likely still need more cluster metadata testing, both more automated tests + more manual testing, but that's going to have to come during the 2.0 QA cycle.

+1 merge after addressing (or deciding to punt on) the two minor comments above.

src/riak_core_metadata.erl
jrwest, line 364: agreed. that is usually my default opinion and I could not remember a reason why I did differently here so changed in https://github.com/basho/riak_core/commit/d166a409173582c89301749bdcb5663a3ef4035b

src/riak_core_broadcast.erl
jrwest: yep, silly typing b/c i made this change quickly and at night so I could run the test as I slept. fixed along w/ one other whitespace thing that bugged me in https://github.com/basho/riak_core/commit/d166a409173582c89301749bdcb5663a3ef4035b

jrwest:
Addressed both review comments since I agreed and it was easy to fix and test quickly. I agree on your point re #440 @jtuple, which is why I was initially against adding the option but I agree its sufficient for 2.0. 

More testing will be forthcoming, primarily in the form of an EQC based `riak_test`. Work on that is being tracked in https://github.com/basho/riak_core/issues/373

Merging.

Issue #442, opened 2013-10-26
active and claimant bucket (type) properties should be immutable


jrwest:
otherwise you can circumvent some of the bucket type invariants.

jrwest:
closing since this was fixed in #460

PR #460, opened 2013-11-26
Validate Bucket Type Properties Required by Core

jrwest:
The `claimant` and `active` properties added in #402 need to immutable. addresses #442 

The eqc test included in this PR adds coverage around the creation semantics of Bucket Types that was missing. The test reproduces the previously mentioned issue if https://github.com/basho/riak_core/commit/bc0d3989fb0de21cf2f9c6dac9c175c193602c4c is not applied. The test should pass with the change.

The test does not have all the coverage I would like. Since it tests on a single node, Bucket Types are never in the `created` state -- they move immediately to `ready` -- and we don't test the time between creation and propagation. For this, it will be better to make the test run inside a `riak_test`. 

With that in mind, and because `eunit` was giving me a hard time, the test cannot be run via `eunit` (or `make test`). Instead it can be run like this:

Compile via rebar (or whatever method you use, I do this in edts normally):

```
./rebar skip_deps=true eunit suites=dne
```

Start Erlang and Run:

```
$ erl -pa deps/*/ebin -pa .eunit 
Erlang R16B02 (erts-5.10.3) [source] [64-bit] [smp:8:8] [async-threads:10] [kernel-poll:false]

Eshell V5.10.3  (abort with ^G)
1> btypes_eqc:run_eqc(1000).
```

### DIFF
--- a/src/riak_core_bucket_props.erl
+++ b/src/riak_core_bucket_props.erl
@@ -40,7 +40,8 @@ merge(Overriding, Other) ->
                undefined | [{atom(), any()}],
                [{atom(), any()}]) -> {ok, [{atom(), any()}]} | {error, [{atom(), atom()}]}.
 validate(CreateOrUpdate, Bucket, ExistingProps, BucketProps) ->
-    validate(CreateOrUpdate, Bucket, ExistingProps, BucketProps, riak_core:bucket_validators(), []).
+    CoreErrors = validate_core_props(CreateOrUpdate, Bucket, ExistingProps, BucketProps),
+    validate(CreateOrUpdate, Bucket, ExistingProps, BucketProps, riak_core:bucket_validators(), CoreErrors).
 
 validate(_CreateOrUpdate, _Bucket, _ExistingProps, Props, [], ErrorLists) ->
     case lists:flatten(ErrorLists) of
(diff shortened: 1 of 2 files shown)


bookshelfdave:
ran eqc w/ 100000 tests: 

```
OK, passed 100000 tests

30.0576% {btypes_eqc,activate_type,1}
29.9617% {btypes_eqc,create_type,2}
10.0307% {btypes_eqc,get_type,1}
10.0028% {btypes_eqc,type_status,1}
9.9826% {btypes_eqc,set_bucket,2}
9.9647% {btypes_eqc,update_type,2}
```

I'll review the PR more tomorrow.

src/riak_core_bucket_props.erl
```diff
@@ -51,6 +52,46 @@ validate(CreateOrUpdate, Bucket, ExistingProps, BucketProps0, [{_App, Validator}
...
+                            Error -> [Error | Errors]
+                        end
+                end, [], BucketProps).
+
+validate_core_prop(create, {_Bucket, undefined}, undefined, {claimant, Claimant}) when Claimant =:= node()->
+    %% claimant valid on first call to create if claimant is this node
+    true;
+validate_core_prop(create, {_Bucket, undefined}, undefined, {claimant, _BadClaimant}) ->
+    %% claimant not valid on first call to create if claimant is not this node
+    {claimant, "Invalid claimant"};
```
bookshelfdave, line 68: does "Invalid claimant" get propagated to the user, and if so, can you provide a more descriptive message?

src/riak_core_bucket_props.erl
```diff
@@ -51,6 +52,46 @@ validate(CreateOrUpdate, Bucket, ExistingProps, BucketProps0, [{_App, Validator}
...
+validate_core_prop(create, {_Bucket, undefined}, Existing, {claimant, Claimant}) ->
+    %% subsequent creation calls cannot modify claimant and it should exist
+    case lists:keyfind(claimant, 1, Existing) of
+        false -> {claimant, "No claimant details found in existing properties"};
+        {claimant, Claimant} -> true;
```
bookshelfdave, line 73: just a nit, but it seems like `claimant` should be a constant to prevent typo errors

src/riak_core_bucket_props.erl
jrwest, line 68: the message may be propogated to the user but its such an exceptional case that only a very long message would describe it. I thought a bit about how it might be more descriptive but didn't come up w/ something I liked. In short, there is a bug if it fails here (we called across nodes to validate properties?)

bookshelfdave, line 68: works for me

test/btypes_eqc.erl
```diff
@@ -0,0 +1,361 @@
...
+    riak_core_bucket_type:get(TypeName).
+
+%% @doc get_type postcondition
+get_type_post(S, [TypeName], Res) ->
+    case type_active(TypeName, S) of
+        false-> eq(undefined, Res);
+        true ->
+            ExpectedDefaults = riak_core_bucket_type:defaults(),
+            {TypeName, true, ExpectedProps} = type(TypeName, S),
+            %% TODO: just adding the list here is kind of wrong, use riak_core_bucket_props:merge? or similar?
```
bookshelfdave: what's the risk in using a fake list of props instead of a result from riak_core_bucket_props?

jrwest: heh, I actually realized when I scrolled through the other day that this comment no longer applies. I tried to stay away from using `riak_core_bucket_props:merge` since its indirectly under test. Maybe a little too paranoid but i think `ukeysort` is sufficient. fwiw, the reason the comment lingers is when I wrote this there was no call to `ukeysort` and the test inevitably failed. I fixed the test but forgot about the comment.

EDIT: i'll remove the comment as part of one final sweep to address any other comments you have.

bookshelfdave:
+1, each grouped operated looks correct, test passes 100000k times

Noting several TODO's related to faults.

I like your use of eqc_group_commands, it def improves test readability.

jrwest:
Removed the comment discussed above, and rebased. Thanks again @metadave. Merging.

PR #462, opened 2013-12-03
Add EQC statem property for vclock. [rebased]

seancribbs:
This is a QuickCheck property for vclock. The property was created so that the `dominates/2` could be added, with confidence that other parts of the API are solid as well. `dominates/2` essentially implements the `>` relation (strict-order) where `descends/2` implements the `>=` relation (partial-order). This could be used in various parts of `riak_kv` where vclocks are compared.

(h/t @joedevivo and @Quviq for their help on this)

/cc @russelldb

### DIFF
--- a/test/vclock_qc.erl
+++ b/test/vclock_qc.erl
@@ -26,9 +26,10 @@ initial_state() ->
 %% Command generator, S is the state
 command(#state{vclocks=Vs}) ->
     oneof([{call, ?MODULE, fresh, []}] ++
-          [{call, ?MODULE, timestamp, []}] ++ 
+          [{call, ?MODULE, timestamp, []}] ++
           [{call, ?MODULE, increment, [gen_actor_id(), elements(Vs)]} || length(Vs) > 0] ++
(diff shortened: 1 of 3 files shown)


test/vclock_qc.erl
```diff
@@ -0,0 +1,162 @@
...
+-define(QC_OUT(P),
+        eqc:on_output(fun(Str, Args) -> io:format(user, Str, Args) end, P)).
+
+-record(state, {vclocks = []}).
+
+eqc_test_() ->
+    {timeout,
+     60,
+     ?_assert(quickcheck(eqc:testing_time(20, more_commands(10,?QC_OUT(prop_vclock())))))}.
+
```
russelldb, line 28: Would be nice to have convenience `test/0` and `test/1` functions, like a lot of our eqc tests have. Makes it far easier to run from the console when one is making potentially breaking changes.

russelldb:
I added the convenience funs myself, 'cos time. If you're happy with them, I'm happy with the rest :+1: merge away!

seancribbs:
Nice, thanks @russelldb.

PR #464, opened 2013-12-05
Fixes

bbrietzke:
This pull request has fixes for a few problems that I have been experiencing when trying to get up and running with riak_core.

Steps to reproduce:
1) Install the templates from rebar_riak_core
2) Create the application from the provided templates
3) Update the version of riak_core to the latest and greatest
4) make rel
5) rel/<<appid>>/bin/<<appid>> console

This will generate a series of exceptions.
a) riak_core_sup attempts calls to application:get_env to find platform_data_dir
    1) application:get_env returns a given value, not {ok, Value} which is leading to a bad_match exception
b) When (a) is fixed, riak_core_metadata_manager doesn't start, returning a no_data_dir reason
    1) riak_core_metadata_manager is looking for platform_data_dir, which is not provided in the app.src

While I was digging around, I noticed that riak_core_ring_manager had it's own, independent setting.  That didn't make much sense to me, so I changed it to use platform_data_dir and build the path from there.  You can consider this optional and I don't have a problem resubmitting the pull request without that changeset.

### DIFF
--- a/src/riak_core.app.src
+++ b/src/riak_core.app.src
@@ -27,6 +27,7 @@
 
          %% Default location of ringstate
          {ring_state_dir, "data/ring"},
+         {platform_data_dir, "data"},
 
          %% Default ring creation size.  Make sure it is a power of 2,
          %% e.g. 16, 32, 64, 128, 256, 512 etc
--- a/src/riak_core.app.src
+++ b/src/riak_core.app.src
@@ -25,8 +25,7 @@
          %% Cluster name
          {cluster_name, "default"},
 
-         %% Default location of ringstate
-         {ring_state_dir, "data/ring"},
+         %% Default location for ring, cluster and other data files
          {platform_data_dir, "data"},
 
          %% Default ring creation size.  Make sure it is a power of 2,
--- a/src/riak_core_ring_manager.erl
+++ b/src/riak_core_ring_manager.erl
@@ -466,7 +466,7 @@ code_change(_OldVsn, State, _Extra) ->
 ring_dir() ->
     case app_helper:get_env(riak_core, ring_state_dir) of
         undefined ->
-            app_helper:get_env(riak_core, platform_data_dir, "data");
+            filename:join(app_helper:get_env(riak_core, platform_data_dir, "data"), "ring");
         Dir ->
             Dir
     end.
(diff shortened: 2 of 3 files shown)


jrwest:
@bbrietzke have you tried defining `platform_data_dir` in https://github.com/basho/rebar_riak_core/blob/master/riak_core.app.config before generating your application?

russelldb:
This is what I did here https://github.com/basho/rebar_riak_core/commit/d00a08b4769dbd0fd1521368df2e8c4bdb45cd77 wfm, will issue a PR when I'm happy it all works.

bbrietzke:
@jrwest @russelldb 

During my initial experimentations, that is exactly what I did.  I was originally going to submit a PR to the rebar_riak_core with that fix until I noticed your branch.

The more I thought about it, the more that solution doesn't sit well with me.

My first concern is that we are attempting to solve a problem that is occurring in this code base, by making changes in another code base.  My understanding of rebar_riak_core is to provide working templates to generate a basic riak_core code base, not to provides solutions to missing settings in riak_core.

As a matter of good practice, shouldn't an application define logical defaults for any infrastructure related items it might need, such as where files are stored, networked services and the like?

Shouldn't the vm.args file provide overrides for the release/node, not be the single source of truth for a particular setting?  

There are also other files/data that needs to be stored as well such as the ring and metadata.  By having a single setting such as platform_data_dir provided with a logical default value, it makes it easier to move/change the location all of the files/data at one time without having to mess with individual settings.

jrwest:
@bbrietzke the addition of the `platform_data_dir` default to the .app.src is fine. Although I agree it would be nicer to have one config variable for the ring and platform data directories, this change may break users upgrading `riak_core` applications from previous versions (e.g. Riak 1.x -> Riak 2.0). By sharing the same directories it also limits where that data can live meaning some users would have to manually move some files if they have things configured a certain way. I suppose you could check for the `ring_state_dir` before using `platform_data_dir` instead of out-right removing it...

Thanks!

bbrietzke:
Would it be worth me closing this PR and submitting a new one without all the intermediate commits?

jrwest:
Either that or rebase+force push would be fine since I don't think anyone has pulled it down yet. Whatever you decide Ill pull it down and make sure everything still builds/runs ok on our end. We are approaching a code freeze so this will likely be merged before then assuming it goes as i expect. However, if I fail to update you after a few days please don't hesitate to remind me. Its a bit hectic.

bbrietzke:
These changes have been rebased/squashed and submitted under PR468.  

I'll go ahead and close this PR, and we can continue the conversation with that new request.

PR #470, opened 2013-12-09
Provide a synchronous registration and unregistration of services.

cmeiklejohn:
Some parts of repl require being able to disable a service and block
until it's disabled.  Provide this API.

@metadave

### DIFF
--- a/src/riak_core_service_mgr.erl
+++ b/src/riak_core_service_mgr.erl
@@ -59,9 +59,12 @@
                 refs = []
                }).
 
--export([start_link/0, start_link/1,
+-export([start_link/0,
+         start_link/1,
          register_service/2,
          unregister_service/1,
+         sync_register_service/2,
+         sync_unregister_service/1,
          is_registered/1,
          register_stats_fun/1,
          get_stats/0,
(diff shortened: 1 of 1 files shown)


src/riak_core_service_mgr.erl
```diff
@@ -179,6 +195,15 @@ handle_call(get_stats, _From, State) ->
     PStats = [{Protocol, Count} || {Protocol,{_Stats,Count}} <- Stats],
     {reply, PStats, State};
 
+handle_call({register_service, Protocol, Strategy},  _From, State) ->
+    {{ProtocolId,_Revs},_Rest} = Protocol,
```
bookshelfdave: how would you feel about register_service + unregister_service calling the same shared functions?

cmeiklejohn: I feel pretty good about it.  Do you mind looking at the amended commit?

bookshelfdave:
can register_service/unregister_service be removed in favor of all sync msgs?

cmeiklejohn:
Not sure, but I would rather do that deprecation in 2.0 instead of 1.4.4.

bookshelfdave:
+1

cmeiklejohn:
I'm going to merge this and open a second PR to forward port this to 2.0, then I'll open another issue for tracking the deprecation of the async registration, sound good?

PR #476, opened 2013-12-11
Remove connection manager and service manager.

cmeiklejohn:
Remove connection manager and service manager.  They are not needed here, and not supervised here.


kellymclaughlin:
:+1:

PR #479, opened 2013-12-11
Extract riak_core_connection and tests.

cmeiklejohn:


bookshelfdave:
+1, repl builds fine and is able to make connections across clusters

PR #480, opened 2013-12-12
Port 1.4 changes to the 2.0 branch.

cmeiklejohn:
Port 1.4 changes to the 2.0 branch.

@metadave


bookshelfdave:
We'll need to do something w/ the ranch dep: https://github.com/basho/riak_core/blob/develop/rebar.config#L15

cmeiklejohn:
In what way?

bookshelfdave:
it only exists in riak_core because of the conn_mgr stuff. It should probably go back to repl.

cmeiklejohn:
Added commit to remove ranch dependency.

bookshelfdave:
+1

PR #481, opened 2013-12-13
Integration of background manager with Handoff and AAE tree rebuild

buddhisthead:
Replaced by https://github.com/basho/riak_core/pull/484

NOTE: DO NOT MERGE YET (requires #475 + a rebase to remove the bg_manager itself from this PR)

The riak_core background manager PR is here: basho/riak_core#475

The riak_repl PR is here: https://github.com/basho/riak_repl/pull/488

The background manager allows subsystems to coordinate with locks (and tokens) in a non-blocking API when desired. In our case, we've seen several customer support issues related to a production cluster being overloaded by the combined weight of MDC fullsync replication in combination with either handoff, or AAE tree rebuilds (or all three!).

Each of these subsystems is being taught, for 2.0, to "take" a lock on the kv vnode/partition before it does a fold over that partition. Locks are returned when the process dies or exists normally. Each subsystem (at this point in time) is coded to keep working on other partitions if one partition is denied a lock via the return value of max_concurrency, which means progress is made even when one partition is busy.

Use of this feature can be bypassed in production. Besides a global skill switch on the background manager itself, the riak_core configuration supports two new parameters: aae_skip_bg_manager and handoff_skip_bg_manager, which defaults to false. Setting it to true will not call the background manager and proceed like it used to.

### DIFF
--- a/src/riak_core_handoff_sender.erl
+++ b/src/riak_core_handoff_sender.erl
@@ -22,6 +22,7 @@
 
 -module(riak_core_handoff_sender).
 -export([start_link/4, get_handoff_ssl_options/0]).
+-include("riak_core_locks.hrl").
 -include("riak_core_vnode.hrl").
 -include("riak_core_handoff.hrl").
 -define(ACK_COUNT, 1000).
@@ -92,7 +93,20 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
     SrcNode = node(),
     SrcPartition = get_src_partition(Opts),
     TargetPartition = get_target_partition(Opts),
-     try
+    %% Take per-vnode concurrency limit "lock". It will be released when this process exists/dies.
+    case maybe_get_vnode_lock(Module, SrcPartition) of
+        max_concurrency ->
+            %% shared lock not registered yet or limit reached.  Failing with
+            %% max_concurrency will cause this partition to be retried again later.
+            lager:info("Failed to get vnode lock for partition ~p", [SrcPartition]),
+            exit({shutdown, max_concurrency});
+        ok ->
+            %% Got the lock or didn't try to. If we got it, our process is now monitored;
+            %% the lock will be released when we exit.
+            ok
+    end,
+
+    try
          Filter = get_filter(Opts),
          [_Name,Host] = string:tokens(atom_to_list(TargetNode), "@"),
          {ok, Port} = get_handoff_port(TargetNode),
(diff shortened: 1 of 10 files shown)


buddhisthead:
Notes to self:
- squash commits into a single entry.
- add cuttlefish schema for bg_mgr configuration parameters

src/riak_core_vnode.erl
```diff
@@ -145,6 +146,7 @@ init([Mod, Index, InitialInactivityTimeout, Forward]) ->
     process_flag(trap_exit, true),
     State = #state{index=Index, mod=Mod, forward=Forward,
                    inactivity_timeout=InitialInactivityTimeout},
+    try_set_vnode_lock_limit(Mod, Index),
```
jrwest, line 149: As we discussed previously I'm not sure `riak_core_vnode` is the correct place to be doing this. As the PR states, the goal of this work is to limit folds over `riak_*kv*_vnode`'s. This registers a lock per-partition, per module so there is also a `{vnode_lock, riak_pipe_vnode, Partition}` for each partition for example. This may have the desired effect of also providing the locks for `riak_kv_vnode` but the scope is too wide. I'll leave some comments elsewhere to try and illustrate. 

It still seems to me that if we want to limit folds over `riak_kv_vnode` backends than `riak_kv` is the correct place to be registering these locks and acquiring them. The challenge is integrating that with handoff because handoff is a general mechanism in core and its hooks at the `riak_kv` level are insufficient for grabbing locks. The correct way is probably to extend these hooks but alas it is probably to late for that. As I previously expressed, I think we should skip handoff this go around because of these issues. Handoff can still be limited via the node-wide concurrency limit so provided exclusive access to the partition can be done via a mixture of two knobs instead of just one. I understand that handoff has been a large pain point and motivation for this work and that a more ideal approach is desired but we have to start being realistic about what we can get in with the time we have.

Without handoff this becomes a much simpler problem. Register the locks in `riak_kv_vnode` as `{riak_kv_vnode_lock, Partition}` and the Repl and AAE changes should basically remain the same. However, I really think the solution we should be discussing is doing locking across all [backend folds that use ?FOLD_REQ](https://github.com/basho/riak_kv/blob/develop/src/riak_kv_vnode.erl#L1564)

src/riak_core_handoff_sender.erl
```diff
@@ -92,7 +93,20 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
     SrcNode = node(),
     SrcPartition = get_src_partition(Opts),
     TargetPartition = get_target_partition(Opts),
-     try
+    %% Take per-vnode concurrency limit "lock". It will be released when this process exists/dies.
+    case maybe_get_vnode_lock(Module, SrcPartition) of
```
jrwest, line 97: this is where we would take the `riak_pipe_vnode` locks. I don't really see anywhere it harms things to do so right now but I also don't see any reason to do it. I know this gets us handoff integration in a way but I'm still struggling with the implementation here and I fear we are trying to force it in.

src/riak_core_vnode.erl
buddhisthead, line 149: I agree that we should not register the pipe locks. I missed that. In the other integration points, I match on riak_kv_vnode and I think we should do the same here. There are plenty of other places in core that do the same thing - it's not like kv is completely isolated from core. I have run riak tests with this configuration and it works fine with handoff.

But maybe you have a good point about where to do the registration. We can do it from kv and still integrate with handoff.

jrwest, line 149: We should definitely not match on the name of a specific vnode here. That is akin to a super-type being knowledgeable of its child types. In addition, `riak_core` is not only being used by Riak. This is registering locks on behalf of every vnode in every `riak_core` application. Core is certainly not the cleanest abstraction but the only references to a specific vnode are examples in tests.

buddhisthead, line 149: Ok, I understand. Moving the registration to kv makes total sense to me. Sorry about the confusion.

buddhisthead:
replaced by https://github.com/basho/riak_core/pull/484

PR #482, opened 2013-12-14
Ensure stats progress and tag individual stale stats

evanmcc:
2.0 version of https://github.com/basho/riak_core/pull/467, with additional tagging of individual stale stats, even though the rest of the stat cache and the overall cache timestamps may continue to progress.

cc @rsltrifork @russelldb

### DIFF
--- a/src/riak_core_stat_calc_proc.erl
+++ b/src/riak_core_stat_calc_proc.erl
@@ -84,7 +84,8 @@ handle_call(value, From, State0=#state{active=Active0, awaiting=Awaiting0,
     Reply = case cache_get(TS, TTL) of
                 No when No == miss; No == stale ->
                     {Active, Awaiting} = maybe_get_stat(Stat, From, Active0, Awaiting0 ),
-                    {noreply, State0#state{active=Active, awaiting=Awaiting}};
+                    {noreply, State0#state{active=Active, awaiting=Awaiting}, 
+                     timer:seconds(10)};
                 hit ->
                     {reply, Value, State0}
             end,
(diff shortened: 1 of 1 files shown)


evanmcc:
should be reviewed with https://github.com/basho/riak_kv/pull/763, see that PR for testing notes and strategies.

src/riak_core_stat_calc_proc.erl
```diff
@@ -93,9 +94,30 @@ handle_call(_Request, _From, State) ->
...
-handle_cast({value, Value, TS}, State=#state{awaiting=Awaiting}) ->
-    [gen_server:reply(From, Value) || From <- Awaiting],
-    {noreply, State#state{value=Value, timestamp=TS, active=undefined, awaiting=[]}};
+handle_cast({value, Value, TS}, State=#state{awaiting=Awaiting, 
+                                             value=OldValue}) ->
+    case Value of
+        {error, Reason} ->
+            lager:debug("stat calc failed: ~p ~p", [Reason]),
+            {Reply, State1} = 
+                case OldValue of
```
russelldb: I prefer `tag_stale/1` had a clause that matched `<<"stale:", _/binary>>` and you could lose this case statement, and keep the knowledge about tagging encapsulated in the `tag_stale/1` function, or is the state mutation too expensive in that case since you'd be setting `State#state{value=V}` even if it is unchanged.

russelldb:
I like it.

I'm curious about your aggressive short line length policy, though. Sometimes I think it makes the code less readable.

src/riak_core_stat_calc_proc.erl
```diff
@@ -140,6 +174,12 @@ do_calc_stat(Stat) ->
     spawn_link(
       fun() ->
               StatVal = riak_core_stat_q:calc_stat(Stat),
-              gen_server:cast(ServerPid, {value, StatVal, folsom_utils:now_epoch()}) end
+              gen_server:cast(ServerPid, {value, StatVal,
```
russelldb: Why the line break change?

russelldb:
Oh, I meant to say +1

src/riak_core_stat_calc_proc.erl
evanmcc: this one because I thought that it looked better with end easier to match with fun, and the rest because it was long line, which are the most horrible thing in the entire universe.

evanmcc:
I suppose that sometimes the long-line thing might make things harder to read.  I think that in erlang it's particularly bad, because of how the default modes push things to the right quite quickly, especially in anonymous functions, which are handled more like right-hand-rule stuff than just adding another level of scope indention.

Anyway, I can revert, I don't really care, it's just that 80 columns is a hard habit to break.

evanmcc:
updated, if you want to take another look.

russelldb:
All the ones are plussed. Merge at will. Many thanks.

PR #483, opened 2013-12-16
Numerous AAE improvements

jtuple:
This pull request enables bloom filters for the LevelDB instances used by AAE. The use of bloom filters was assumed in the design of AAE, but the proper setting was mistakenly overlooked. This fixes that.

Also, this change exports the `hashtree:flush_buffer/1` function that is used by the sibling K/V pull request: basho/riak_kv#765

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -111,6 +111,7 @@
          update_snapshot/1,
          update_perform/1,
          rehash_tree/1,
+         flush_buffer/1,
          close/1,
          destroy/1,
          read_meta/2,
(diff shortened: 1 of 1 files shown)


engelsanchez:
The changes are simple. I manually confirmed the option is correct and the Leveldb logs clearly show the bloom filter is now enabled. I wonder what the impact will be in terms of memory. We should keep an eye on that during testing.

+1

engelsanchez:
Sorry, I forgot these:  :+1: :dancer: :boat:

PR #484, opened 2013-12-18
Background Manager Integration with Handoff

buddhisthead:
NOTE: DO NOT MERGE YET (this PR contains code from #475)

The riak_core background manager PR is here: basho/riak_core#475
The riak_repl PR is here: basho/riak_repl#488

The background manager allows subsystems to coordinate with locks (and tokens) in a non-blocking API when desired. In our case, we've seen several customer support issues related to a production cluster being overloaded by the combined weight of MDC fullsync replication in combination with either handoff, or AAE tree rebuilds (or all three!).

Each of these subsystems is being taught, for 2.0, to "take" a lock on the kv vnode/partition before it does a fold over that partition. Locks are returned when the process dies or exists normally. Each subsystem (at this point in time) is coded to keep working on other partitions if one partition is denied a lock via the return value of max_concurrency, which means progress is made even when one partition is busy.

Use of this feature can be bypassed in production. Besides a global skill switch on the background manager itself, the riak_core configuration supports two new parameters: aae_skip_bg_manager and handoff_skip_bg_manager, which defaults to false. Setting it to true will not call the background manager and proceed like it used to.

This PR is a rebase and re-implementation. It replaces https://github.com/basho/riak_core/pull/481

### DIFF
--- a/include/riak_core_bg_manager.hrl
+++ b/include/riak_core_bg_manager.hrl
@@ -1,0 +1,88 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2013 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% NOTES:
+%% The background manager allows tokens and locks to be "acquired" by
+%% competing processes in a way that limits the total load on the cluster.
+%%
+%% The model is different than your typical semaphore. Here, we are
+%% interested in coordinating background jobs that start, run, and die.
+%% 
+%%
+%% The term "given" is a general version of "held", "acquired", or
+%% "allocated" for both locks and tokens. Held doesn't make sense for
+%% tokens since they aren't held. So, "given" applies to both locks
+%% and tokens, but you can think "held" for locks if that's more fun.
+%%
+%% Resources are defined by their "names", which is the same as "type"
+%% or "kind". A lock name might be the atom 'aae_hashtree_lock' or the
+%% tuple '{my_ultimate_lock, 42}'.
+%%
+%% Usage:
+%% 1. register your lock/token and set it's max concurrency/rate.
+%% 2. "get" a lock/token by it's resource type/name
+%% 3. do stuff
+%% 4. let your process die, which gives back a lock.
+%% -------------------------------------------------------------------
+-type bg_lock()  :: any().
+-type bg_token() :: any().
+-type bg_resource()      :: bg_token() | bg_lock().
+-type bg_resource_type() :: lock | token.
+
+-type bg_meta()  :: {atom(), any()}.                %% meta data to associate with a lock/token
+-type bg_period() :: pos_integer().                 %% token refill period in milliseconds
+-type bg_count() :: pos_integer().                  %% token refill tokens to count at each refill period
+-type bg_rate() :: undefined | {bg_period(), bg_count()}.       %% token refill rate
+-type bg_concurrency_limit() :: non_neg_integer() | infinity.   %% max lock concurrency allowed
+-type bg_state() :: given | blocked | failed.       %% state of an instance of a resource.
+
+%% Results of a "ps" of live given or blocked locks/tokens
+-record(bg_stat_live,
+        {
+          resource   :: bg_resource(),            %% resource name, e.g. 'aae_hashtree_lock'
+          type       :: bg_resource_type(),       %% resource type, e.g. 'lock'
+          consumer   :: pid(),                    %% process asking for token
+          meta       :: [bg_meta()],              %% associated meta data
+          state      :: bg_state()                %% result of last request, e.g. 'given'
+        }).
+-type bg_stat_live() :: #bg_stat_live{}.
+
+%% Results of a "head" or "tail", per resource. Historical query result.
+-record(bg_stat_hist,
+        {
+          type    :: undefined | bg_resource_type(),  %% undefined only on default
+          limit   :: non_neg_integer(),  %% maximum available, defined by token rate during interval
+          refills :: non_neg_integer(),  %% number of times a token was refilled during interval. 0 if lock
+          given   :: non_neg_integer(),  %% number of times this resource was handed out within interval
+          blocked :: non_neg_integer()   %% number of blocked processes waiting for a token
+        }).
+-type bg_stat_hist() :: #bg_stat_hist{}.
+-define(BG_DEFAULT_STAT_HIST,
+        #bg_stat_hist{type=undefined, limit=undefined, refills=0, given=0, blocked=0}).
+
+-define(BG_DEFAULT_WINDOW_INTERVAL, 60*1000).    %% in milliseconds
+-define(BG_DEFAULT_OUTPUT_SAMPLES, 20).     %% default number of sample windows displayed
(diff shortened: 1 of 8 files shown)


include/riak_core_locks.hrl
```diff
@@ -0,0 +1,8 @@
+%% Copyright(C), Basho Technologies, 2013
+%%
+%% @doc Definitions of shared locks/tokens for use with background manager.
+%% See @link riak_core_background_mgr:get_lock/1
+%%
+%% @doc vnode_lock(Module, PartitionIndex) is a kv per-vnode lock, used possibly,
+%% by AAE tree rebuilds, fullsync, and handoff.
+-define(VNODE_LOCK(Mod, Idx), {vnode_lock, Mod, Idx}).
```
jrwest: there isn't any reason for this to be in `riak_core` any more. It should be moved to `riak_kv` and it no longer needs to be parameterized on the `Mod`.

src/riak_core_handoff_sender.erl
```diff
@@ -92,7 +92,28 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
     SrcNode = node(),
     SrcPartition = get_src_partition(Opts),
     TargetPartition = get_target_partition(Opts),
-     try
+
+    %% Consult handoff module for optional implemenation of handoff_started/2
+    case lists:member({handoff_started, 2}, Module:module_info(exports)) of
+        true ->
+            Source = {SrcPartition, SrcNode},
```
jrwest: This information isn't useful to the called vnode because it already has it (its information about itself). All the information to construct the same argument passed to `handoff_starting/2` is available here but I wonder if we should just pass the worker pid. 

Unfortunately, also, I realized we don't have the vnode state which makes the callback a bit unorthodox. I think for now its fine since this is a "special callback" that most people won't need but I wondered briefly about calling into the `riak_core_vnode` and allowing it to make the callback to `handoff_started` (this last part is just a future note for myself).

src/riak_core_handoff_sender.erl
```diff
@@ -92,7 +92,28 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
     SrcNode = node(),
     SrcPartition = get_src_partition(Opts),
     TargetPartition = get_target_partition(Opts),
-     try
+
+    %% Consult handoff module for optional implemenation of handoff_started/2
```
jrwest: extraneous comment? think it can be removed or should be clarified.

jrwest:
The actual diff is very small so i've done a quick review of the code already. Couple comments I think should be addressed and then I will run this along w/ https://github.com/basho/riak_kv/pull/769

src/riak_core_handoff_sender.erl
```diff
@@ -92,7 +92,28 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
     SrcNode = node(),
     SrcPartition = get_src_partition(Opts),
     TargetPartition = get_target_partition(Opts),
-     try
+
```
jrwest, line 95: it would also be nice to pull this all out into a function (or at least the part about performing the callback, maybe leave the case statement). `start_fold` is already a beast.

src/riak_core_handoff_sender.erl
jrwest: ah I see why you pass that info now that I look at kv more closely. Its precisely because you don't have the state. For the sake of time this is probably sufficient since, again, its not a "real" callback for the vnode. But in the future it would be great to try to make it operate more like the "real" callbacks do.

buddhisthead: Agreed, but I didn't know how to do that since we aren't in the same context.

include/riak_core_locks.hrl
buddhisthead: ok. Can we just add it to an existing kv include file? Or just rename and move the file? I liked it in core because it would be one place to see all the lock types we define. But I see your point and it's been bugging me a little too.

src/riak_core_handoff_sender.erl
buddhisthead: I don't understand what's not clear with this comment. It says that we're checking for an optional function in the module. How is that not useful to comment? If you have wording that you prefer, please suggest it.

src/riak_core_handoff_sender.erl
buddhisthead, line 95: Addressed by b03f95b

test/bg_manager_eqc.erl
```diff
@@ -0,0 +1,875 @@
+%%% @author Jordan West <>
```
jrwest: this still needs a proper header before merging

test/bg_manager_eqc.erl
```diff
@@ -0,0 +1,875 @@
...
+%%% @copyright (C) 2013, Jordan West
+%%% @doc
+%%%
+%%% @end
+%%% Created : 13 Nov 2013 by Jordan West <>
+
+-module(bg_manager_eqc).
+
+%%-ifdef(TEST).
```
jrwest: these should be un-commented before merge.

test/bg_manager_tests.erl
```diff
@@ -0,0 +1,176 @@
+-module(bg_manager_tests).
+-compile(export_all).
+
+-include_lib("riak_core_bg_manager.hrl").
```
jrwest, line 4: I was under the impression we were removing most of these tests because it is now covered by eqc and this suite would cover what eqc does not, specifically the `Kill=true` scenario for `set_concurrency_limit/3`?

test/bg_manager_eqc.erl
```diff
@@ -0,0 +1,875 @@
...
+          %% max counts per "period" for token types
+          counts :: [{bg_eqc_type(), bg_eqc_limit()}],
+          %% locks held (or once held then released) for each lock type
+          %% and their state
+          locks  :: [{bg_eqc_type(), [{reference(), pid(), [], held | released}]}],
+          %% number of tokens taken by type
+          tokens :: [{bg_eqc_type(), non_neg_integer()}]
+         }).
+
+run_eqc() ->
```
jrwest, line 57: the eqc test isn't plugged into `eunit` right now. not necessary before merge but maybe open an issue to remind us?

include/riak_core_bg_manager.hrl
```diff
@@ -0,0 +1,73 @@
...
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% NOTES:
+%% The background manager allows tokens and locks to be "acquired" by
```
jrwest: usually these sorts of docs would go in the module edoc where you already have the info about the ets tables.

test/bg_manager_eqc.erl
buddhisthead, line 57: Addressed by https://github.com/basho/riak_core/issues/485

jrwest:
+1 merge after addressing the license header and commented out ifdefs in eqc. Other comments being addressed would be great but I don't see them as necessary to merge. Opened issues where you agree and it makes sense would be good. 

This includes the background manager work from #475 whose review jumped here because i lost track of PRs.

Final Comments:
- In the end, I believe the use of the table manager added more complexity than was necessary over the `public` ets table owned by a parent supervisor approach. The former, used in this PR, adds a third state callers must handle, besides process alive/process dead. Process alive but doesn't own table, admittedly a small window but besides the point, means we need functions like `try_set_concurrency_limit` in each subsystem that wants to register a lock. I'd like to see us move away from the table manager and use the approach taken by the entropy manager and the new `riak_kv_hooks`, before we start integrating into other subsystems more.
- We should document the many ways the background manager can be disabled somewhere
- cuttlefish mappings are needed for any new config stuff added

include/riak_core_locks.hrl
buddhisthead: addressed by b645b94

test/bg_manager_eqc.erl
buddhisthead: addressed by b645b94

test/bg_manager_eqc.erl
buddhisthead: addressed by b645b94

buddhisthead:
Rebased off develop and force pushed before final merge.

PR #487, opened 2013-12-21
Feuerlabs stat combo

uwiger:
Same functionality as in #465, but with the possibility to choose the old ('legacy') stats package at compile-time.

The function identifying which system is used is `riak_core_stat:stat_system()`. This function is hard-coded to 'exometer', but can be changed to 'legacy' by setting the OS environment variable RIAK_CORE_STAT_SYSTEM to "legacy", and ensuring that the module riak_core_stat.erl is recompiled.


jrwest:
seems like `2.1` thing, someone with more knowledge can perhaps correct me

jburwell:
Superseded by https://github.com/basho/riak_core/pull/650

PR #490, opened 2013-12-25
Fix mailing list link in README

sinasamavati:


rzezeski:
Thank you for the fix.

PR #491, opened 2013-12-25
remove legacy forwarding code that snuck in as part of 632af2b3ce9

jrwest:
Handling of forwarding code was changed in #301. Changes made as part
of the commit mentioned above were most likely made prior to that PRs merge.

Remove the forwarding code as it is redundant (the decision to forward
is handled further up the call stack). It caused vnode when resizing to
crash, since the check was not as comprehensive as its
caller.  The vnode would crash when attempting to forward to a
partition identified by `{Index, [{Index, Owner}]}` which is invalid.

See http://giddyup.basho.com/#/projects/riak_ee/scorecards/66/66-712-verify_dynamic_ring-centos-6-64-bitcask/30889/artifacts/435114 for example failure. `verify_dynamic_ring` in `riak_test` should now pass with this change and `sync_command_test` in `riak_core` should continue to pass.

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -297,28 +297,14 @@ forward_or_vnode_command(Sender, Request, State=#state{forward=Forward,
 
 vnode_command(_Sender, _Request, State=#state{modstate={deleted,_}}) ->
     continue(State);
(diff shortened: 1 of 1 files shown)


slfritchie:
+1, sorry I didn't know what r_t test to use to find the regression.

Issue #494, opened 2013-12-29
riak_core_bucket:get_buckets/1 improperly handles reset typed buckets


jrwest:
The tombstones are not filtered:

``` erlang
1> riak_core_bucket_type:create(<<"a">>, []).
ok
2> riak_core_bucket_type:activate(<<"a">>).
ok
3> riak_core_bucket:set_bucket({<<"a">>, <<"b">>}, [{n_val, 5}]).
ok
4> riak_core_bucket:get_buckets(element(2, riak_core_ring_manager:get_my_ring())).
[[{active,true},
  {allow_mult,true},
  {basic_quorum,false},
  {big_vclock,50},
  {chash_keyfun,{riak_core_util,chash_std_keyfun}},
  {claimant,'dev1@127.0.0.1'},
  {dw,quorum},
  {last_write_wins,false},
  {linkfun,{modfun,riak_kv_wm_link_walker,mapreduce_linkfun}},
  {n_val,5},
  {name,{<<"a">>,<<"b">>}},
  {notfound_ok,true},
  {old_vclock,86400},
  {postcommit,[]},
  {pr,0},
  {precommit,[]},
  {pw,0},
  {r,quorum},
  {rw,quorum},
  {small_vclock,50},
  {w,quorum},
  {young_vclock,20}]]
5> riak_core_bucket:reset_bucket({<<"a">>, <<"b">>}).
ok
6> riak_core_bucket:get_buckets(element(2, riak_core_ring_manager:get_my_ring())).
['$deleted']
```

PR #495, opened 2013-12-29
Filter Typed Bucket Tombstones in riak_core_ring:get_buckets/1

jrwest:
addresses #494

This should be merged before final release or resetting a typed bucket could have bad affects on at least ring re-sizing and primary partition repair (e.g. `riak_kv_vnode:repair`)


russelldb:
Tested by hand, works as advertised. +1

Issue #493, opened 2013-12-28
Claimant Bucket Type Operations Must Account for Mixed Cluster


jrwest:
Riak 2.0 claimant's must account for other nodes running pre-2.0 versions of Riak. Otherwise the operations are likely to crash it needlessly.

Although extra protection will be provided by #492 this case should be handled directly in the claimant as well.

Example:

``` erlang
$ dev/dev2/bin/riak-admin bucket-type status type
RPC to 'dev2@127.0.0.1' failed: {'EXIT',
                                 {{{case_clause,
                                    {badrpc,
                                     {'EXIT',
                                      {undef,
                                       [{riak_core_metadata,get,
                                         [{core,bucket_types},
                                          <<"type">>,
                                          [{default,[]}]],
                                         []},
                                        {rpc,'-handle_call_call/6-fun-0-',5,
                                         [{file,"rpc.erl"},{line,203}]}]}}}},
                                   [{lists,ukeysort,2,
                                     [{file,"lists.erl"},{line,833}]},
                                    {riak_core_claimant,
                                     '-get_remote_type_status/2-lc$^0/1-0-',2,
                                     [{file,"src/riak_core_claimant.erl"},
                                      {line,713}]},
                                    {riak_core_claimant,
                                     get_remote_type_status,2,
                                     [{file,"src/riak_core_claimant.erl"},
                                      {line,713}]},
                                    {riak_core_claimant,handle_call,3,
                                     [{file,"src/riak_core_claimant.erl"},
                                      {line,304}]},
                                    {gen_server,handle_msg,5,
                                     [{file,"gen_server.erl"},{line,585}]},
                                    {proc_lib,init_p_do_apply,3,
                                     [{file,"proc_lib.erl"},{line,239}]}]},
                                  {gen_server,call,
                                   [{riak_core_claimant,'dev2@127.0.0.1'},
                                    {bucket_type_status,<<"type">>},
                                    infinity]}}}
$ dev/dev2/bin/riak-admin bucket-type activate type
RPC to 'dev2@127.0.0.1' failed: {'EXIT',
                                 {{{case_clause,
                                    {badrpc,
                                     {'EXIT',
                                      {undef,
                                       [{riak_core_metadata,get,
                                         [{core,bucket_types},
                                          <<"type">>,
                                          [{default,[]}]],
                                         []},
                                        {rpc,'-handle_call_call/6-fun-0-',5,
                                         [{file,"rpc.erl"},{line,203}]}]}}}},
[...]

PR #498, opened 2013-12-31
Fix a Couple Mixed Cluster Issues w/ Bucket Types

### LINKED ISSUE #492: Bucket Type Capability for Mixed Clusters Required
In mixed clusters where the claimant is a running a pre-2.0 version of Riak, some bucket type functions, when run on another node, will fail. In some cases, it will do so silently because of the [catch-all](https://github.com/basho/riak_core/blob/1.4.2/src/riak_core_claimant.erl#L208-L210) in `handle_call/2`.

To prevent this, any function in `riak_core_bucket_type` that contacts the claimant should be guarded by a capability. The following functions are affected:
- `create/2`
- `status/1`
- `activate/1`
- `update/2`
- `reset/1` -- although not exposed it should be guarded just in case

`get/1` does not need to be guarded because it only uses the local cluster metadata process on the node.

Example where `dev1` is running Riak 1.4.x and `dev2` is running Riak 2.0preX:

```
$ dev/dev2/bin/riak-admin bucket-type create type '{"props": {}}'
type created # nope it lies
$ dev/dev2/bin/riak-admin bucket-type status type
RPC to 'dev2@127.0.0.1' failed: {'EXIT',
                                 {function_clause,
                                  [{riak_kv_console,bucket_type_print_status,
                                    [<<"type">>,ok],
                                    [{file,"src/riak_kv_console.erl"},
                                     {line,451}]},
                                   {riak_kv_console,bucket_type_status,1,
                                    [{file,"src/riak_kv_console.erl"},
                                     {line,448}]},
```

jrwest:
Fixes #492 & #493. 

For review, for now, testing by hand + running any existing tests (e.g. `btypes_eqc`) is the easiest way. I am working on extending the bucket types eqc test to cover these cases but didn't want to hold up this PR.

### DIFF
--- a/src/riak_core_bucket_type.erl
+++ b/src/riak_core_bucket_type.erl
@@ -108,6 +108,11 @@
 -type bucket_type()       :: binary().
 -type bucket_type_props() :: [{term(), term()}].
 
+-define(IF_CAPABLE(X, E), case riak_core_capability:get({riak_core, bucket_types}) of
+                              true -> X;
+                              false -> E
+                          end).
+
 
 %% @doc The hardcoded defaults for all bucket types.
 -spec defaults() -> bucket_type_props().
@@ -141,15 +146,16 @@ defaults() ->
 create(?DEFAULT_TYPE, _Props) ->
     {error, default_type};
 create(BucketType, Props) when is_binary(BucketType) ->
-    riak_core_claimant:create_bucket_type(BucketType,
-                                          riak_core_bucket_props:merge(Props, defaults())).
+    ?IF_CAPABLE(riak_core_claimant:create_bucket_type(BucketType,
+                                                      riak_core_bucket_props:merge(Props, defaults())),
+                {error, not_capable}).
 
 %% @doc Returns the state the type is in.
 -spec status(bucket_type()) -> undefined | created | ready | active.
 status(?DEFAULT_TYPE) ->
     active;
 status(BucketType) when is_binary(BucketType) ->
-    riak_core_claimant:bucket_type_status(BucketType).
+    ?IF_CAPABLE(riak_core_claimant:bucket_type_status(BucketType), undefined).
 
 %% @doc Activate the type. This will succeed only if the type is in the `ready' state. Otherwise,
 %% an error is returned.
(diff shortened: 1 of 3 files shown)


src/riak_core_bucket_type.erl
```diff
@@ -141,15 +146,16 @@ defaults() ->
 create(?DEFAULT_TYPE, _Props) ->
     {error, default_type};
 create(BucketType, Props) when is_binary(BucketType) ->
-    riak_core_claimant:create_bucket_type(BucketType,
-                                          riak_core_bucket_props:merge(Props, defaults())).
+    ?IF_CAPABLE(riak_core_claimant:create_bucket_type(BucketType,
```
russelldb, line 149: I don't understand how this works…doesn't the first argument get evaluated anyway?

russelldb:
Tested by hand. Works as advertised. I have a question as a comment above.

Also, when I start a node of 2.0 up, and a node of 1.4.6 up. And make a cluster of them, I get a flurry of errors reports:

```
10:53:32.548 [error] CRASH REPORT Process <0.895.0> with 0 neighbours exited with reason: no such process or port in call to gen_server:call({riak_core_metadata_hashtree,'dev1@127.0.0.1'}, {prefix_hash,{core,bucket_types}}, 1000) in gen_server:terminate/6 line 744
10:53:32.548 [error] gen_fsm <0.1523.0> in state ready terminated with reason: no such process or port in call to gen_server:call({riak_core_metadata_hashtree,'dev1@127.0.0.1'}, {prefix_hash,{core,bucket_types}}, 1000)
10:53:32.548 [error] gen_fsm <0.836.0> in state ready terminated with reason: no such process or port in call to gen_server:call({riak_core_metadata_hashtree,'dev1@127.0.0.1'}, {prefix_hash,{core,bucket_types}}, 1000)
```

src/riak_core_bucket_type.erl
jrwest, line 149: no because the macros arguments are not evaluated just replaced. they are only evaluated when the case statement is at runtime. A quick example, using `io:format` shows this:

``` erlang
-module(mod).

-compile(export_all).

-define(M(X, Y), case random:uniform(5) < 2 of
                     true -> X;
                     false -> Y
                 end).

a() ->
    ?M(io:format("a~n"), io:format("b~n")).
```

```
11> mod:a().
b
ok
12> mod:a().
a
ok
```

jrwest:
@russelldb those errors are cause by https://github.com/basho/yokozuna/issues/258. You can comment out the offending code in `riak_kv_vnode` to verify.

src/riak_core_bucket_type.erl
russelldb, line 149: I am amaze.

russelldb:
Thanks for the explanation. +1.

PR #499, opened 2013-12-31
properly handle default bucket type in get/1

jrwest:
addresses #496


russelldb:
Tested by hand, works as advertised, +1

PR #500, opened 2014-01-03
Allow PLT destination to be specified

reiddraper:


rzezeski:
Works like a charm. +1 to merge.

PR #501, opened 2014-01-06
Add logging and pretty printing of log file to background manager.

buddhisthead:
The background manager already contains functions to view the current set of locks and tokens, but lacks any historical view of how subsystems have interacted. It would be highly useful to examine past get_lock/get_token patterns in the event of performance problems around the participating subsystems that use the background manager, especially on nodes with extreme overload or crashed nodes.

This PR adds logging to disk, which will persist after a crashed node. All writes are done using Erlang's disk_log module and are asynchronous so that normal operations are not blocked by slow disk writes. Losing a little data from the disk_log cash is preferable to blocking a caller of the background manager.

Three logging levels are provided, with `on` being the default of the config variable `background_manager_loglevel`:
- `off` - no logging is done
- `on` - logs only the get_lock, get_token, enable/disable, and bypass calls as events.
- `debug` - logs all operations

Logfile entries are in disk_log's "internal" format under `riak_core_bg_manager/events.xxx` in the data directory of each node. Writing is fast and the log files are repairable when read. The logs are `wrap` logs and thus will occupy a fixed maximum space on disk (and memory for the disk_log write buffer). The total size of the logs are controlled by `background_manager_log_size` application configuration variable, and default to 5 files of 1MB each, so the maximum disk space used is 5 MB, which should give days of logged events for postmortem analysis.

Testing was done using both the `bg_manager_eqc.erl` EQC test as a driver (find log files in /tmp/riak_core_bg_manager/events.xxx) , and with `riak_test -t rtee replication2` (find log files in devN/data/riak_core_bg_manager/events.xxx). After running the replication2 test, we see entries like the following:

```
6/Jan/2014:18:29:0 get_lock {vnode_lock, 685078892498860742907977265335757665463718379520} [{task, handoff}] ok
6/Jan/2014:18:29:0 get_lock {vnode_lock, 685078892498860742907977265335757665463718379520} [{task,  repl_fullsync}] max_concurrency
6/Jan/2014:18:29:0 get_lock {vnode_lock, 68507889249886074290797726533575766546371837952} [{task,  repl_fullsync}] ok
```

### DIFF
--- a/src/riak_core_bg_manager.erl
+++ b/src/riak_core_bg_manager.erl
@@ -1032,14 +1032,14 @@ maybe_log(_Op, _Args, _Status, _State) ->
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 %% @private
-%% @doc Format the timestamp into human readable date string.
+%% @doc Format the timestamp into a human readable date string. Timezone depends
+%% on the value of the app env var 'sasl/utc_log' which is true or false. If true,
+%% then universal time is returned, otherwise local time prevails.
+%% @link https://github.com/basho/lager/blob/master/src/lager_stdlib.erl#L80
 format_utc_timestamp(TS) ->
-    {{Year,Month,Day},{Hour,Minute,Second}} = 
-	calendar:now_to_universal_time(TS),
-    Mstr = element(Month,{"Jan","Feb","Mar","Apr","May","Jun","Jul",
-			  "Aug","Sep","Oct","Nov","Dec"}),
-    io_lib:format("~w/~s/~w:~w:~w:~w",
-		  [Day,Mstr,Year,Hour,Minute,Second]).
+    DateTimeMsecs = lager_util:localtime_ms(TS),
+    {D,T} = lager_util:format_time(lager_util:maybe_utc(DateTimeMsecs)),
+    lists:flatten([D,$ ,T]).
 
 %% @private
 %% @doc
--- a/src/riak_core_bg_manager.erl
+++ b/src/riak_core_bg_manager.erl
@@ -133,7 +133,7 @@
 
 -define(NOT_TRANSFERED(S), S#state.info_table == undefined orelse S#state.entry_table == undefined).
 
--define(MAX_LOG_BYTES, 1000000).    %% maximum size of a single logfile
+-define(MAX_LOG_BYTES, 100 * 1024). %% maximum size of a single logfile
 -define(MAX_LOG_FILES, 5).          %% maximum number of logfiles in the wrap rotation
 
 %%%===================================================================
--- a/src/riak_core_bg_manager.erl
+++ b/src/riak_core_bg_manager.erl
@@ -435,7 +435,7 @@ init([]) ->
                    entry_table=undefined, %% resolved in the ETS-TRANSFER handler
                    enabled=true,
                    bypassed=false,
-                   loglevel=app_helper:get_env(riak_core, background_manager_loglevel, debug),
+                   loglevel=app_helper:get_env(riak_core, background_manager_loglevel, on),
                    log=Log
                   },
     {ok, State}.
--- a/src/riak_core_bg_manager.erl
+++ b/src/riak_core_bg_manager.erl
@@ -115,7 +115,9 @@
          all_locks/0,
          all_locks/1,
          all_tokens/0,
-         all_tokens/1
+         all_tokens/1,
+         print_log/0,
+         iter_log/1
         ]).
 
 %% Convenience
@@ -130,6 +132,9 @@
 -define(SERVER, ?MODULE).
 
 -define(NOT_TRANSFERED(S), S#state.info_table == undefined orelse S#state.entry_table == undefined).
+
+-define(MAX_LOG_BYTES, 1000000).    %% maximum size of a single logfile
+-define(MAX_LOG_FILES, 5).          %% maximum number of logfiles in the wrap rotation
 
 %%%===================================================================
 %%% API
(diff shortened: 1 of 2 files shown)


src/riak_core_bg_manager.erl
```diff
@@ -943,3 +972,113 @@ use_bg_mgr() ->
...
+maybe_log(_Op, _Args, _Status, _State) ->
+    ok.
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+%%% Called from outside of riak
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+%% @private
+%% @doc Format the timestamp into human readable date string.
+format_utc_timestamp(TS) ->
```
Vagabond: You may want to consider using lager_util:format_time and friends so that you don't have a completely different timestamp format to the rest of riak (and the Erlang ecosystem in general).

src/riak_core_bg_manager.erl
```diff
@@ -943,3 +972,113 @@ use_bg_mgr() ->
...
+print_entry({TS, Op, {Resource, Value}, Status}) ->
+    io:format("~s ~p ~p ~p ~p~n", [format_utc_timestamp(TS), Op, Resource, Value, Status]);
+print_entry({TS, Op, Args, Status}) ->
+    io:format("~s ~p ~p ~p~n", [format_utc_timestamp(TS), Op, Args, Status]);
+print_entry(Term) ->
+    io:format("Unexpected: ~p~n", [Term]).
+
+%% @doc
+%% Pretty print background manager's logfile.
+print_log() ->
```
Vagabond, line 1058: What is the intended use of this, and will it io:format all 5mb worth of logs, or just the latest one?

buddhisthead, line 1058: As of now, it's the only way to dump the logs in a human readable way. I'd like to add more in the future, but I don't think we can jam it into 2.0 at this point. Yes, it will dump all of the logs. I could easily add an entry point that will only print N entries, but I'm trying to keep this dirt simple. Since the iterator walks over chunks, it won't io:format anything bigger than a single entry from the list at a time.

src/riak_core_bg_manager.erl
buddhisthead: Addressed by aefe367. Good call. Besides the benefit of using consistent formats, it also adds support for both local and UTC time zone pretty printing based on the app env `sasl/utc_log`

src/riak_core_bg_manager.erl
buddhisthead, line 1058: I guess it's not so easy to add the print(N) thing. It would be easy to print the first N, but not the last N without doing a massive list creation and reversal in memory. Boo. While it's possible to index to the end of the log minus some amount, that seems complicated for a PR that's last minute. We can always print_log() | tail -N. It's only 5 MB.

Vagabond, line 1058: I'm still unsure about the wisdom of pushing 5mb through erlang's IO subsystem.

Vagabond:
I guess I'm sort of conceptually stuggling with this feature. If we want Riak to provide historical information, shouldn't it be more structured than just a sequential event log? For example, I would think it would be more interesting to inspect via vnode ID, as well as show the current status of all of the vnodes.

If each vnode had a bounded buffer for its own events, rather then all of them be in a global one, it would also help debugging stuck ones. If, for example, a vnode gets stuck and that is not noticed for a day or two and the cluster is under a workload that uses the bgmgr, the chances are high that the logs for the stuck vnode will have been rotated out by all the successful messages. One way to do this would be one disk_log per vnode, or even just an in-memory dict with a sized list of events (eg. 20 events per vnode).

This logging approach also doesn't seem to help much when looking for an current status of all the locks, without a bit of mental bookkeeping.

I think if we want to punt on something more complicated we should at least remove the print() call (or not add it to any console commands). The more I think about that, the worse it seems it could be on a busy cluster.

src/riak_core_bg_manager.erl
buddhisthead, line 1058: Fair. I'm running a test now to generate the full 5 MB of logs and then I'll pretty print it. Since it's an iterator, it should do fine, but we'll see.

buddhisthead:
A separate log per vnode is a good idea and I should have done that straight away. But I don't like the in-memory approach because it makes the logging more complicated, doing calculations to trim the size. I had originally stored history in a queue which is really easy to write to, but again wanted to avoid in-memory solutions that vanish on node crash. I think that we could reduce the file size way down from 1 MB or even reduce the number of rotation files. That would also reduce the impact of the print call a lot. It's probably ok to just leave this accessible from the shell console and not make a riak-admin command for it, maybe not until after we have more experience with it.

The current status of all locks is available via existing api. See riak_core_bg_manager:all_locks() and friends.

src/riak_core_bg_manager.erl
buddhisthead, line 1058: Well, it was hard generating that much data, so I turned down the file size to 100K and it prints the log with no effort at all. That gives 500K worth of log data.

So, hopefully 1a759c6 addresses your concerns.

buddhisthead:
Well, writing a separate log for each vnode is a huge pain and I don't think it's worth the code complexity. For one thing, we just happen to care about vnodes right now, but there is nothing vnode-centric about the background manager. If a particular vnode is stuck, we'll just have to grep the logs for that partition number. The rate of lock and token acquisition is not high for the vnode lock and if there is a super busy lock, then at least it will survive in the log the longest. I think some operational experience during the testing phase will help determine the best log file size needed. I am still due to write a riak_test for the background-manager, but we have the RC candidate looming as well.

jonmeredith:
@Vagabond does anything need to happen with this?

jrwest:
moved to milestone `2.0.1`. This would have been a nice addition (in some form) but its not going to make it I don't think

PR #503, opened 2014-01-06
Don't discard config options that are not validated

Vagabond:
This was a change made as part of the security work that breaks SSL replication.


cmeiklejohn:
:+1: to merge

PR #504, opened 2014-01-10
fix make test

andrewjstone:
Fix 2 failing tests
- `refresh_my_ring_test` was failing due to an ETS table not being setup properly.
- `core_vnode_eqc` was failing in multiple places
  - shutdown had to be changed to kill so that processes didn't hang during testing
  - an asyncwork response wasn't being counted for asynccrash calls so that needed to be added back in (it was commented out)
  - With worker pools of 1, if an asyncrash call was followed by another async call that calls work wouldn't be counted because the message was dropped. I changed the possible pool sizes from [0,1,10] to [0,4,10] to make this issue extremely unlikely to happen. You'd need 4 worker crashes then another async call before any of them were restarted. Since they all have 100ms delays before crashing on purpose in mock_vnode that seems highly unlikely.

### DIFF
--- a/src/riak_core_ring_manager.erl
+++ b/src/riak_core_ring_manager.erl
@@ -244,7 +244,7 @@ do_write_ringfile(Ring, FN) ->
             lager:error("Unable to write ring to \"~s\" - ~p\n", [FN, Err]),
             {error,Err}
     end.
-    
+
 
 %% @spec find_latest_ringfile() -> string()
 find_latest_ringfile() ->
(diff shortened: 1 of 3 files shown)


rzezeski:
Mind amending your commit message and copy/pasting the PR body into it? Has lots of good context in case someone has to do code archeology in the future.

andrewjstone:
@rzezeski fixed and pooshed. Thanks for catching this. I need to get into a this habbit.

kellymclaughlin:
Great work! +1 to merge.

PR #489, opened 2014-01-10
riak_core_net_ticktime should be guarded by capability

kellymclaughlin:
To prevent case where a user has not yet loaded the module on a pre-2.0 node.

Quick sketch:
- Capability has two modes, supported & unsupported.
- On startup, if module is available register the capability w/ supported as the preferred mode, otherwise unsupported
- Since `start_set_net_ticktime_daemon` runs first on a riak node, guard it by a check for the currently preferred mode across the cluster and only proceed if it is supported. If it fails, this indicates to the user that one or more nodes does not have the module
- User loads module on node (add its to basho-patches) and calls new function `riak_core_net_ticktime:enable/0`. This re-registers the capability on the node w/ supported as the preferred mode
- Once done on all nodes that did not have the module and ring has converged user can safely call `start_set_net_ticktime_daemon` and since the capability prefers supported it will proceed

### DIFF
--- a/src/riak_core_app.erl
+++ b/src/riak_core_app.erl
@@ -98,8 +98,8 @@ start(_StartType, _StartArgs) ->
                                           [true, false],
                                           false),
             riak_core_capability:register({riak_core, net_ticktime},
-                                          [supported, unsupported],
-                                          unsupported),
+                                          [true, false],
+                                          false),
 
             {ok, Pid};
         {error, Reason} ->
--- a/src/riak_core_app.erl
+++ b/src/riak_core_app.erl
@@ -97,6 +97,9 @@ start(_StartType, _StartArgs) ->
             riak_core_capability:register({riak_core, bucket_types},
                                           [true, false],
                                           false),
+            riak_core_capability:register({riak_core, net_ticktime},
+                                          [supported, unsupported],
+                                          unsupported),
 
             {ok, Pid};
         {error, Reason} ->
--- a/src/riak_core_net_ticktime.erl
+++ b/src/riak_core_net_ticktime.erl
@@ -36,14 +36,12 @@ enable() ->
                                   false).
 
 start_set_net_ticktime_daemon(Node, Time) ->
-    Capability = riak_core_capability:get({riak_core, net_ticktime}),
-    start_set_net_ticktime_daemon(Node, Time, Capability).
+    start_set_net_ticktime_daemon(Node, Time, net_ticktime_active()).
 
 start_set_net_ticktime_daemon(Node, Time, true) ->
     EbinDir = filename:dirname(code:which(?MODULE)),
     try
         Dirs = rpc:call(Node, code, get_path, []),
-        lager:info("Node: ~p EbinDir: ~p Member? ~p", [Node, EbinDir, lists:member(EbinDir, Dirs)]),
         case lists:member(EbinDir, Dirs) of
             false ->
                 lager:info("start_set_net_ticktime_daemon: adding to code path "
@@ -142,4 +140,13 @@ set_net_ticktime(Time) ->
             Status;
         A when is_atom(A) ->
             A
+    end.
+
+-spec net_ticktime_active() -> boolean().
+net_ticktime_active() ->
+    case catch riak_core_capability:get({riak_core, net_ticktime}) of
+        true ->
+            true;
+        _ ->
+            false
     end.
--- a/src/riak_core_net_ticktime.erl
+++ b/src/riak_core_net_ticktime.erl
@@ -23,12 +23,23 @@
 %% @doc Change net_kernel's ticktime on-the-fly.
 
 -module(riak_core_net_ticktime).
--export([start_set_net_ticktime_daemon/2,
+-export([enable/0,
+         start_set_net_ticktime_daemon/2,
          stop_set_net_ticktime_daemon/1]).
 
 -define(REGNAME, net_kernel_net_ticktime_change_daemon).
 
+-spec enable() -> ok.
+enable() ->
+    riak_core_capability:register({riak_core, net_ticktime},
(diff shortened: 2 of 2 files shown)


kellymclaughlin:
Regarding the use of `nodes(connected)` [here](https://github.com/basho/riak_core/blob/develop/src/riak_core_net_ticktime.erl#L105), it seems to have the potential to cause issues. While testing this capability I see errors such as this once I start the net_ticktime daemon from a `riak attach` console:

```
=ERROR REPORT==== 10-Jan-2014::14:45:39 ===
Error in process <0.91.0> on node 'c_31409_dev1@127.0.0.1' with exit value: {{badfun,#Fun<riak_core_net_ticktime.1.104612057>},[{erlang,apply,2,[]}]}
```

The console is a hidden node, but it does not have the `riak_core_net_ticktime` module loaded. It seems like it would be safer to just call `net_kernel:set_net_ticktime` on hidden nodes, but I could certainly be missing some context about this. 

Beyond this issue the capability seems to be working fine.

kellymclaughlin:
Related to the previous comment, I see errors similar to the following when running the `verify_tick_change` riak_test:

```
16:21:54.534 [error] Error in process <0.533.0> on node 'riak_test@127.0.0.1' with exit value: {{nocatch,{unknown_capability,{riak_core,net_ticktime}}},[{riak_core_capability,get,1,[{file,"src/riak_core_capability.erl"},{line,140}]},{riak_core_net_ticktime,start_set_net_ticktime_daemon,2,[{file,"src/riak_cor...
```

kellymclaughlin:
Just an update on this:

I have been working on refactoring my changes and also some general refactoring to make this module easier to follow, but I keep getting stuck. 

Currently I am trying to understand how this code really helps at all beyond the short life span that processes started by calling `start_set_net_ticktime` daemon run in. If this is a daemon shouldn't there be a long-running process for it? Why don't we have a `gen_server` that continuously runs and periodically checks for new nodes? It would need need to check at least once per N seconds where N is the currently set tick time, which is where a server keeping that state would be useful. New hidden nodes from `riak attach` sessions can happen any time and they all start with the default tick time of 60 seconds and currently it seems like we rely on manual execution of `start_net_ticktime_daemon` to handle this. 

If the tick time is 5 or 10 seconds that is potentially a lot of noise generated from just checking nodes that have already been set. A good answer to that is probably to run a daemon process on each Riak node and part of the state tracks nodes that have been set so we avoid as much repeat work as possible and each node can be responsible for detecting hidden nodes attached to it and only contacting them. 

Another problem is that the spawned processes that run the `start_set_net_ticktime_daemon` function on each node are killed after 5 seconds, but the default transition time for changes to the tick time to take effect is 60 seconds. This results in a lot of unnecessary process churn during that time and is probably better avoided.

I am happy to work on some changes to address these concerns, but I would love to hear other input on it first. 

cc: @slfritchie

kellymclaughlin:
The `verify_tick_change` now passes without the errors previously reported. Also wrapped the capability check in a function to hide the details of handling the exception is the capability doesn't exist. The `verify_tick_change` test is currently failing for me using HEAD of `develop` for unrelated reasons. It passes if I use the `pre9` tag and augment it with these changes.

russelldb:
I can't get the test to pass on develop either.

Unrelated to your changes, after reading the module I have an idea of what it is doing, still have no idea _why_ we'd want to do this ever: I really think a little documentation on the purpose of this service would go long way.

russelldb:
Tried with pre9 tag and still see same cluster join errors. Code looks fine. Would rather get a test pass before merging _but_ +1 from reading.

kellymclaughlin:
I reran the `verify_tick_change` test with this branch rebased from latest `develop` and it passes some of the time for me. Other times the tick time change is still ongoing for some nodes after the calls to `write_read_poll_check` complete and causes the test to fail.

The test as written is insufficient to show the problems I have seen so I have modified the test on [this](https://github.com/basho/riak_test/compare/klm-demo-tick-change-issue?expand=1) branch to show how it is insufficient. Even after the 3 minute sleep, the riak attach call should find the ticktime to be 11, but instead it finds it to be 60 every time I run it.

slfritchie:
I'm not understanding what `rt:attach/2` is supposed to do, but `_Res` here: https://github.com/basho/riak_test/blob/c24288518bfbb4f7491b24ecb84d5b1727c0476a/tests/verify_tick_change.erl#L64

... is simply `ok` when, regardless of whether the `expect` criteria succeeds/fails.

The test is also missing an important step: adding a ticktime config item to `riak.conf` ... which, it seems, didn't exist until pr 133 above ... and isn't honored until pr 113 above.

slfritchie:
I pushed a commit to https://github.com/basho/riak_test/compare/klm-demo-tick-change-issue?expand=1 which fixes that test, I believe, if the two PRs in this ticket, cuttlefish 133 and node_package 113, are also applied

kellymclaughlin:
Thanks for making those changes. I verified that they do resolve the issue. Assuming we can get those merged in I'll kick this back to @russelldb for another look.

Vagabond:
Correct me if I am wrong, but the riak_test does not include a hidden node?

Also, some comments in the code would make it a lot easier to understand, Kelly and Jordan bootstrapped me, but it should not be necessary.

slfritchie:
@Vagabond The rt:attach() is -remsh-style pseudo-attachment, and it uses -hidden.

andrewjstone:
- [x] code inspection
- [x] eunit + eqc
- [x] dialyzer
- [x] xref
- [x] verify_tick_change

andrewjstone:
:+1:

kellymclaughlin:
Merging this, but I will follow-up with a riak_test PR to ensure we take advantage of the test improvements that resulted from this review.

PR #506, opened 2014-01-15
Handle tombstones when accumulating sources

Vagabond:
This prevents a function clause like the one seen in basho/riak#469.


jrwest:
this change is pretty much the same idea as https://github.com/basho/riak_core/pull/495. +1 merge. Similar to #495 it would be nice for security to move to using the `default` option to `fold/4` instead of the tombstone defined in `riak_core_metadata.erl`.

Vagabond:
I'm waiting for an option to skip tombstones, did not want to change all the folds twice.

PR #509, opened 2014-01-18
User iterator prefetching in hashtree module

kellymclaughlin:
Use prefetch and prefetch_stop actions for iterator_move calls in the hashtree
module. This replaces the use of the next action with the goal of making
segment traversal more efficient. This change adds a prefetch field to the
itr_state record and some additional case clauses in the iterate/2 function.
This ensures that an iterator can be safely used with other actions after
being used for prefetch iteration.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -189,7 +189,8 @@
                     remaining_segments :: ['*' | integer()],
                     acc_fun            :: fun(([{binary(),binary()}]) -> any()),
                     segment_acc        :: [{binary(), binary()}],
-                    final_acc          :: [{integer(), any()}]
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
Testing note: this work relies on the work merged in [this](https://github.com/basho/eleveldb/commit/a7ff06a0c0779be88e82769e7978870d5eb18d0f) eleveldb commit so make sure to pull those changes.

lordnull:
Reviewing...
- [x] eunit passes
- [x] eqc passes (no new failures)
- [x] r_t passes (aae seems to be the primary one this would effect, but general functionality should work too)
- [x] code review

lordnull:
Looks good, :+1:

PR #510, opened 2014-01-22
Dominates should take timestamp into account.

cmeiklejohn:
Ensure that dominates takes into account the timestamp; two concurrent
puts happening one second apart should not be considered related.

### DIFF
--- a/src/vclock.erl
+++ b/src/vclock.erl
@@ -100,8 +100,22 @@ descends(Va, Vb) ->
 descends_dot(Vclock, Dot) ->
     descends(Vclock, [Dot]).
 
+%% @doc true if `A' strictly dominates `B'. Note: uses timestamp
+%% information as well as causal information. In Riak it is possible
+%% to have vclocks that are identical except for timestamps. See
+%% source comment for more details.
+%%
+
 -spec dominates(vclock(), vclock()) -> boolean().
 dominates(A, B) ->
+    %% In a sane world if two vclocks descend each other they MUST be
+    %% equal. In riak they can descend each other and have different
+    %% timestamps(!) How? Deleted keys, re-written, then restored is
+    %% one example. See riak_kv#679 for others. This is why we must
+    %% check descends both ways rather than checking descends(A, B)
+    %% and not equal(A, B). Do not "optimise" this to dodge the second
+    %% descends call! I know that the laws of causality say that each
+    %% actor must act serially, but Riak breaks that, so timestamps.
     descends(A, B) andalso not descends(B, A).
 
 % @doc Combine all VClocks in the input list into their least possible
--- a/src/vclock.erl
+++ b/src/vclock.erl
@@ -102,7 +102,7 @@ descends_dot(Vclock, Dot) ->
 
 -spec dominates(vclock(), vclock()) -> boolean().
 dominates(A, B) ->
-    descends(A, B) andalso not equal(A, B).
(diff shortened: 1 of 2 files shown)


src/vclock.erl
```diff
@@ -102,7 +102,7 @@ descends_dot(Vclock, Dot) ->
 
 -spec dominates(vclock(), vclock()) -> boolean().
 dominates(A, B) ->
-    descends(A, B) andalso not equal(A, B).
+    descends(A, B) andalso not descends(B, A).
```
russelldb, line 122: In a sane world if two vclocks descend each other they _have_ to be equal. In riak they can descend each other and have different timestamps? How? Deleted keys, re-written, then restored is one example. See riak_kv#679 for others. This is why we must check descends both ways rather than checking descends(A, B) and not equal(A, B). I think we need a comment that says this. 'Cos someone who knows version vectors may try and optimise this to use one descends and equals. (I know that, 'cos I did :( )

russelldb:
Pushed fix to the model, and added a lengthy comment to `dominates`

russelldb:
I plus one your changes. Please review mine (cc @seancribbs). Still not happy about this generally, tho.

seancribbs:
@russelldb :+1: on the code, but I'm also not happy with this change in general. Not sure what else we can do.

cmeiklejohn:
Wait, so do we want to merge this?  Seems like we have no choice?

russelldb:
Not without a corresponding fix on the kv side for DVV

src/vclock.erl
```diff
@@ -378,4 +392,10 @@ valid_entry_test() ->
     ?assertNot(valid_entry("huffle-puff")),
     ?assertNot(valid_entry([])).
 
+dominates_test() ->
+    A1 = vclock:increment(a, vclock:fresh()),
+    timer:sleep(1000), %% Sleep 1, to alter timestamp for same actor.
```
reiddraper: For testing, might it be useful to offer another arity of `vclock:increment` that allows you to explicitly pass in a timestamp?

russelldb: There is one, it should be used here, yes.

cmeiklejohn: There is one already, `increment/3`, and we could switch to use it.

reiddraper: I can do this, if one of you hasn't started already.

russelldb: I am too busy weeping over the reason / meaning off this fix, please go ahead.

reiddraper: Is it expected that [this](https://gist.github.com/reiddraper/c23b836b513c14c860b8) should pass?

cmeiklejohn: If you print the objects, you'll notice that they both have the same timestamp because the creation happens within the same second, so no, it should not pass.

reiddraper:
I've verified that the `vclock_qc` test fails without @cmeiklejohn's change.

russelldb:
Got to say, the PR comment is misleading "timestamps _shouldn't_ be taken into account"

russelldb:
And that the test fails without mine but with Chris's, right?

cmeiklejohn:
True, I think earlier yesterday we were thinking about how they should be.  Can someone just amend the commit title and force-push?

reiddraper:
I've amended the commit and force-pushed.

cmeiklejohn:
:+1:, depends on basho/riak_kv#814.

PR #513, opened 2014-01-28
plumb through coverage queries during overload

andrewjstone:
Send back overload responses during coverage queries. 

Part of a fix for https://github.com/basho/riak_kv/issues/753


kellymclaughlin:
:+1:

PR #514, opened 2014-01-30
Re-order Resize Transfers so Partitions Not Changing Ownership Go First

jrwest:
This commit provides a stop-gap measure to prevent a degenerate case
when resizing a ring with larger datasets. When a ring is resized, it
is possible that some partitions do not change ownership. If these
partitions have data transferred to them before being allowed to
perform their transfers they are regularly iterating over and ignoring
a bunch of data from the new ring. In some cases, this can cause
handoff to timeout because the sender fold is spending too much time
skipping items not bound for the current reciever. This can lead to
the stalling of the resize operation itself. A temporary workaround
has been to raise the handoff timeout values to let these partitions
complete their transfers, however, with this commit, these transfers
are scheduled first -- taking advantage of the next list being treated
as ordered by the vnode manager. Although this is by no means a
solution in 100% of cases it should be a general improvement and
should prevent the need for timeout adjustments, assuming when folding
that the data is well-distributed accross the various preflists the vnode
is responsible for a portion of.

For the reviewer:
- [x] `verify_dynamic_ring` should still pass (there is known intermittent failure on giddyup being investigated)
- [x] verify next list sort order is correctly modified

NOTE: this also needs some large dataset testing, which I plan to do, before merging.

### DIFF
--- a/src/riak_core_ring.erl
+++ b/src/riak_core_ring.erl
@@ -874,12 +874,31 @@ set_pending_resize(Resizing, Orig) ->
     %% all existing indexes must transfer data when the ring is being resized
     Next = [{Idx, Owner, '$resize', [], awaiting} ||
                {Idx, Owner} <- riak_core_ring:all_owners(Orig)],
+
+    %% Whether or not the ring is shrinking or expanding, some
+    %% ownership may be shared between the old and new ring. To prevent
+    %% degenerate cases where partitions whose ownership does not
+    %% change are transferred a bunch of data which they in turn must
+    %% ignore on each subsequent transfer, we move them to the front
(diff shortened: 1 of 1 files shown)


bookshelfdave:
Sort order looks correct. `lists:member({Idx, Owner}, FutureOwners)` returning `true` will put these entries at the front of the list.

Nicely commented. 

parts of verify_dynamic_ring pass w/ multiple runs, however the test fails repeatedly with the aforementioned intermittent failure ("testing force-replace during resize").

bookshelfdave:
riak_test output: https://gist.github.com/metadave/9009819

bookshelfdave:
+1

PR #516, opened 2014-01-30
Add property_hash/1 function to riak_core_bucket_type module

kellymclaughlin:
Add riak_core_bucket_type:property_hash/1 function to provide a convenient way
to calculate a hash of certain bucket type properties whose values could
impact the treatment of buckets created with a particular bucket type.

### DIFF
--- a/src/riak_core_bucket_type.erl
+++ b/src/riak_core_bucket_type.erl
@@ -102,7 +102,8 @@
          itr_next/1,
          itr_done/1,
          itr_value/1,
-         itr_close/1]).
+         itr_close/1,
+         property_hash/1]).
 
 -export_type([bucket_type/0]).
 -type bucket_type()       :: binary().
@@ -112,7 +113,7 @@
                               true -> X;
                               false -> E
                           end).
-
+-define(HASHED_BUCKET_TYPES_PROPS, [consistent, datatype, n_val, allow_mult, last_write_wins]).
 
 %% @doc The hardcoded defaults for all bucket types.
 -spec defaults() -> bucket_type_props().
@@ -218,3 +219,13 @@ itr_value(It) ->
 -spec itr_close(riak_core_metadata:iterator()) -> ok.
 itr_close(It) ->
     riak_core_metadata:itr_close(It).
+
+%% @doc Returns a hash of the bucket type properties whose values may
+%% have implications on the treatment or handling of buckets created
+%% using the bucket type.
+-spec property_hash(undefined | bucket_type_props()) -> undefined | integer().
+property_hash(undefined) ->
+    undefined;
+property_hash(Props) when is_list(Props) ->
+    erlang:phash2([lists:keyfind(Prop, 1, Props) ||
+                      Prop <- ?HASHED_BUCKET_TYPES_PROPS]).
--- a/src/riak_core_bucket_type.erl
+++ b/src/riak_core_bucket_type.erl
@@ -103,7 +103,6 @@
          itr_done/1,
          itr_value/1,
          itr_close/1,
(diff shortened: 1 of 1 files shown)


src/riak_core_bucket_type.erl
```diff
@@ -218,3 +219,13 @@ itr_value(It) ->
...
 itr_close(It) ->
     riak_core_metadata:itr_close(It).
+
+%% @doc Returns a hash of the bucket type properties whose values may
+%% have implications on the treatment or handling of buckets created
+%% using the bucket type.
+-spec property_hash(undefined | bucket_type_props()) -> undefined | integer().
+property_hash(undefined) ->
+    undefined;
+property_hash(Props) when is_list(Props) ->
```
jrwest: as this function is written, i noticed two things:
- its actually just a hashing fucntion right now, its not really specifically tied to bucket types. Why pass in the properties vs. passing in the bucket name? Or having both options in cases where doing two reads is insufficient.
- the list of props to hash over should probably be a second argument. There seems to be some duplication in repl and these properties are specific to it.

kellymclaughlin: Yep very true. Jon and I also had some discussions along similar lines the other day. I think it does make sense to at least have the option to pass in the type and have it handle the propery lookup for you. I also like the idea of passing in the list of keys that are of interest. I'll rework this a bit and push a new commit.

The duplicated code in repl is going to be taken out and it will use this instead. [Here](https://github.com/basho/riak_repl/issues/510) is the pr for that.

src/riak_core_bucket_type.erl
```diff
@@ -112,7 +115,7 @@
                               true -> X;
                               false -> E
                           end).
-
+-define(DEFAULT_HASH_PROPS, [consistent, datatype, n_val, allow_mult, last_write_wins]).
```
kellymclaughlin: I was hoping to keep this self-contained to replication, but since we are storing these in riak_kv objects I think it makes sense to have a default that works for replication here. This allows us to avoid a call into a repl module or having to epose kv to this detail at all. It's not perfect, but seems like a reasonable compromise.

jrwest: so should this be in `riak_kv` then where `riak_object` lives?

kellymclaughlin: Good question. I don't know if it _should_, but it certainly _could_. The downside of that is that then it would have to be duplicated in `riak_repl` as well so if the list ever needed to be altered it would have to happen in twice as many places.

jrwest: why would it need to be duplicated? doesn't repl depend on kv? i could be totally out of my depth here...just trying to keep things clean (says the one who just added a function exactly where he didn't want to)

kellymclaughlin:
I pushed a new commit to remove the default hash properties macro and the `property_hash/1` function. More discussion is [here](https://github.com/basho/riak_repl/issues/510).

kellymclaughlin:
@jrwest Any more thoughts on this or can I get a green light to merge?

jrwest:
@kellymclaughlin looks like my comment here was adressed. i had just passer-by reviewed, though, so +1 from me, but I don't know if someone else should run it through its paces w/ the rest of the changes?

bowrocker:
@jrwest @kellymclaughlin I have been running riak_tests for replication with this branch, as well as the one on this PR:

https://github.com/basho/riak_kv/pull/823

 for a week now -- replication2, repl_bucket_types -- and they are consistently passing.

andrewjstone:
:+1:  Code is good and I tested with riak_kv#823

PR #519, opened 2014-02-05
If the security capability is unknown, consider security off

Vagabond:
This fixes some tests in riak_kv. Along with basho/riak_kv#829, this makes the KV tests all pass for me.

cc @andrewjstone


andrewjstone:
:+1:

PR #520, opened 2014-02-06
add {refresh_iterator, true} for backend folds during handoff send

andrewjstone:
Should be merged with these other PRs:
- https://github.com/basho/riak_kv/pull/852  
- https://github.com/basho/riak_repl/pull/521


jrwest:
as discussed on hipchat, the commit I pushed addresses the issue that riak_core shouldn't be passing riak_kv backend specific options.

jrwest:
I'm +1 on this but I also added code to this so I don't know if we want to get one more pair of eyeballs on it....

Issue #522, opened 2014-02-06
User roles are lost when setting passwords


macintux:
Looking at a fix for this now.

If `riak-admin security alter-user <user> password=<password>` is invoked for a user with assigned roles, `riak_core_security:alter_user/2` loses the existing roles.

macintux:
Closing after merge

PR #524, opened 2014-02-06
Stop doing implicit role manipulation

macintux:
Per #522, when invoking `riak-admin security alter-user` the code loses all existing roles if only a password is provided.

The option parsing code creates `{"roles", []}` as a tuple when `riak-admin security alter-user <user> role=` is invoked (which is the only way to remove all roles), but `riak_core_security:validate_role_option/1` also added `{"roles", []}` as a tuple when no roles information was passed at all, hence the bug.

This patch instructs `validate_role_option/1` to simply return the options without adding the unnecessary tuple. It also handles the one case in the code where (I believe) an absence of such a tuple might cause problems by adding a default value of `[]` to a roles `lookup`.

/cc @Vagabond


Vagabond:
+1.

PR #526, opened 2014-02-06
Cuttlefish RHS Substitutions

joedevivo:
This is the first "in the wild" use of Cuttlefish's new Right Hand Side Substitutions.

Check out the change to `ring.state_dir` for an idea of how it's changed, but basically, instead of using `{{platform_data_dir}}` at build time to determine the default, you can now reference `#(platform_data_dir)` which will be evaluated at runtime.
- [x] CSE love
- [x] ProdMgmt love
- [x] Eng love

### DIFF
--- a/priv/riak_core.schema
+++ b/priv/riak_core.schema
@@ -46,7 +46,8 @@
 %% @doc Default location of ringstate
 {mapping, "ring.state_dir", "riak_core.ring_state_dir", [
   {datatype, directory},
-  {default, "#(platform_data_dir)/ring"}
+  {default, "#(platform_data_dir)/ring"},
+  {level, advanced}
 ]}.
 
 %% @doc Default cert location for https can be overridden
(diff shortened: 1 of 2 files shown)


angrycub:
CSE +1

gburd:
Why did "ring" move into the data dir?

joedevivo:
For who knows how long `{{ring_state_dir}}` was `{{platform_data_dir}}/ring` this just moves that abstraction into cuttlefish

angrycub:
Isn't that only on a devrel?  On a rel or package, the data dir is like /var/lib/riak, eh?

joedevivo:
It's the same for both devrel and rel. I'm not sure about the package.

gburd:
Honestly, I forget where it traditionally lived.

gburd:
PM +1

seancribbs:
Eng +1

PR #527, opened 2014-02-06
Support role= as alternative to roles= for alter-user

macintux:
Previously, if an admin created or updated a user to have new roles but mistyped `role=` instead of `roles=` the command would succeed but the list of roles would be stashed into extra options, not actually useful for security.

This creates an effective alias for `role=<whatever>`.

Stole @Vagabond's much smarter approach than my original attempt.


macintux:
Recommend rejecting this; should instead make the security commands verbose about their outcomes

PR #528, opened 2014-02-11
Fix Dialyzer warnings

seancribbs:
Getting a clean slate. Only remaining warnings may be:

```
Unknown functions:
  cluster_info:format/3
  cluster_info:register_app/1
```

Those will disappear if cluster_info is on the code path.


src/riak_core_claimant.erl
```diff
@@ -400,8 +398,8 @@ maybe_commit_staged(State) ->
 maybe_commit_staged(Ring, State=#state{changes=Changes, seed=Seed}) ->
     Changes2 = filter_changes(Changes, Ring),
     case compute_next_ring(Changes2, Seed, Ring) of
-        {legacy, _} ->
-            {ignore, legacy};
+        {error, invalid_resize_claim} ->
+            lager:error("invalid_resize_claim BUG");
```
jrwest: this probably doesn't need to log. just surface the error to the user? or does this path not bubble back up that far...need to look...

seancribbs:
- [x] Revert `lists:foreach` into LCs which are faster, underscore match result.
- [x] Break out `riak_core_vnode:vnode_forward`.
- [x] Address https://github.com/basho/riak_core/pull/528#discussion_r9837539
- [x] Restore deleted annotations to riak_core_apl functions.

src/riak_core_claimant.erl
jrwest: ok we should be able to bubble this up to the user like so (sorry this is going to be a little verbose because im working through it as i type):
- `maybe_commit_staged/2` where this change has been made is called within a `riak_core_ring_manager:ring_trans/2` so instead of logging we can return `{ignore, invalid_resize_claim}` on L402. the current implementation is actually an invalid return value for the `ring_trans/2` call (although the ring manager handles this)
- the `{ignore, invalid_resize_claim}` will be transformed to `{error, invalid_resize_claim}` here https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core_claimant.erl#L389-L390 and this ultimately bubbles up to the caller as the return value of `riak_core_claimant:commit/1`
- we will have to add a new case to https://github.com/basho/riak_core/blob/develop/src/riak_core_console.erl#L742-L761 to display the error to the user. something like https://github.com/basho/riak_core/blob/develop/src/riak_core_console.erl#L610-L614

seancribbs:
Sorry for the commit rearrangement, I interactively removed the big `foreach` commit to change them back to LCs, then added a commit to remove the warnings.

src/riak_core_claimant.erl
seancribbs: How's this?

``` erlang
io:format("Unable to commit staged ring changes while ring is resizing.~n"
          "Check that there are no pending changes in 'riak-admin ring-status'~n"
          "If there are, try again once they are completed,~n"
          "Otherwise try again shortly.~n");
```

jrwest: sorry for the delay:

"Unable to commit staged ring changes while ring is resizing" -> "Unable to commit staged ring changes"

The issue is not that the ring is resizing but that another operation took place (e.g. unstaged join) between the plan and the commit. Not sure how to best explain that, so figure its easier just to be short.

src/hashtree.erl
```diff
@@ -631,7 +631,7 @@ get_disk_bucket(Level, Bucket, #state{id=Id, ref=Ref}) ->
 set_disk_bucket(Level, Bucket, Val, State=#state{id=Id, ref=Ref}) ->
     HKey = encode_bucket(Id, Level, Bucket),
     Bin = term_to_binary(Val),
-    eleveldb:put(Ref, HKey, Bin, []),
+    ok = eleveldb:put(Ref, HKey, Bin, []),
```
jrwest, line 634: @jtuple are we ok w/ semantic change here that we may crash if failing to write to leveldb instead of silently ignoring it? I _think_ its ok, but wanted to double check. we could also just change `ok =` to `_ =`...

jrwest:
there are a few cases still where legacy gossip doesn't seem to be fully removed. were these left in intentionally?

gist from @seancribbs of all the mentions of "legacy" in `riak_core`: https://gist.github.com/seancribbs/9145331

https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core.erl#L74-L78
https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core.erl#L108-L115
https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core.erl#L153-L168
https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core_ring.erl#L192-L196
https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core_claim.erl#L119-L133
https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core_console.erl#L42-L50

src/riak_core_ring.erl
```diff
@@ -1095,15 +1095,10 @@ complete_resize_transfers(State, Source, Mod) ->
 deletion_complete(State, Idx, Mod) ->
     transfer_complete(State, Idx, Mod).
 
--spec resize_transfers(chstate(), {integer(), term()}) ->
```
jrwest, line 1098: i'd rather see these specs corrected if they are wrong. as someone who wrote these functions and forgets what they return regularly the spec, as stands, is extremely helpful.

jrwest:
i'd like to see https://github.com/basho/riak_core/commit/2de80a14a89eb2dae73056017a6ffd7684aaf1fd reverted. Most of these functions are exported and many of those specs are useful. Some of them should be modified to not include the variable names for brevity.

src/hashtree.erl
seancribbs, line 634: Indeed, my main reason for introducing this is to raise the questions: 
- Should the process continue if the put fails? Will this lead to incorrect results?
- If it is safe to ignore error conditions, why?

src/riak_core_security.erl
```diff
@@ -409,12 +410,7 @@ add_revoke(all, Bucket, Revokes) ->
     %% all is always valid
     case validate_permissions(Revokes) of
         ok ->
-            case add_revoke_int([all], Bucket, revokes) of
```
jrwest, line 412: these changes are no longer valid i don't think. perhaps one more rebase?

seancribbs:
Reverting 2de80a1 introduces some more warnings, will fix.

```
riak_core_capability.erl:412: Invalid type specification for function riak_core_capability:preferred_modes/4. The success typing is ([any()],[{atom(),[any()]}],[{atom(),{_,_,_,_}}],_) -> [{{_,_},_}]
riak_core_stat_cache.erl:87: The created fun has no local return
riak_core_stat_cache.erl:88: The call riak_core_stat_cache:register_mod(App::any(),{{_,'produce_stats',[]},non_neg_integer()},Registered::any()) breaks the contract (atom(),registered_app(),orddict:orddict()) -> orddict:orddict()
riak_core_vnode_manager.erl:371: Expression produces a value of type pid() | [pid()], but this value is unmatched
riak_core_vnode_manager.erl:854: Expression produces a value of type pid() | [pid()], but this value is unmatched
Unknown functions:
  cluster_info:format/3
  cluster_info:register_app/1
Unknown types:
  ordsets:ordset/0
```

seancribbs:
So I was unable to rebase cleanly again with the security changes, but I just copied the `riak_core_security` module over from `develop`. After reinstating some of my previous changes, here's a warning I can't untangle, which seems to have been introduced in the separation of roles/groups and users:

```
riak_core_security.erl:489: The call riak_core_security:add_grant_int(['all'],Bucket::any(),Grants::maybe_improper_list()) will never return since it differs in the 1st argument from the success typing arguments: ([{_,'group' | 'user'}],any(),maybe_improper_list())
riak_core_security.erl:516: The call riak_core_security:add_revoke_int(['all'],Bucket::any(),'revokes') will never return since it differs in the 1st argument from the success typing arguments: ([{_,'group' | 'user'}],any(),'revokes' | maybe_improper_list())
```

Essentially, calling `add_grant_int/3` or `add_revoke_int/3` with `[all]` as the first argument will never match. Is this code path even valid?

/cc @Vagabond @macintux

macintux:
Huh. I kept focusing on users & groups and forgot to watch carefully what happened to ‘all’, and a quick experiment does lead me to believe I screwed that up nicely.

-John

seancribbs:
There seems to be a race or something in [this test](https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core_vnode_proxy.erl#L357-L372). From buildbot:

```
  riak_core_vnode_proxy: overload_test_ (should not discard in normal operation)...*failed*
in function riak_core_vnode_proxy:'-overload_test_/0-fun-5-'/2 (src/riak_core_vnode_proxy.erl, line 368)
**error:{assertEqual_failed,[{module,riak_core_vnode_proxy},
                     {line,368},
                     {expression,"Count"},
                     {expected,50001},
                     {value,47842}]}
```

andrewjstone:
Yep. We definitely discussed removing that since it's covered by r_t. No
idea why it fails in particular locally. After 2-3 weeks of not touching
it, I think I'm just going to remove it.

src/riak_core_dtrace.erl
```diff
@@ -127,12 +127,7 @@ put_tag(Tag) ->
         true ->
             FTag = iolist_to_binary(Tag),
             put(?DTRACE_TAG_KEY, FTag),
-            case get(?MAGIC) of
-                dtrace ->
-                    dtrace:put_utag(FTag);
```
jrwest, line 132: this change seems to be removing some functionality? `put_utag` vs. `put_tag`?

seancribbs, line 132: `dtrace` module was from Scott's patch before it was fixed-up/merged into OTP. There is no such module anymore.

src/gen_nb_server.erl
```diff
@@ -111,7 +125,7 @@ handle_info({inet_async, ListSock, _Ref, {ok, CliSocket}}, #state{cb=Callback, s
   inet_db:register_socket(CliSocket, inet_tcp),
   case Callback:new_connection(CliSocket, ServerState) of
     {ok, NewServerState} ->
-      prim_inet:async_accept(ListSock, -1),
+      {ok, _} = prim_inet:async_accept(ListSock, -1),
```
jrwest, line 128: I'm not sure it's ok to crash here if `async_accept/2` returns `{error, _}`.

note: this behaviour is used by `riak_core_handoff_listener`.

jrwest, line 128: if it does crash, the handoff listener, for example, would be restarted and all would be fine. but previously we ignored the error, so just want to make sure this semantic change is ok.

src/riak_core.erl
```diff
@@ -81,15 +81,10 @@ join(false, _, Node, Rejoin, Auto) ->
...
-            case false of
-                true ->
-                    legacy_join(Node);
-                _ ->
-                    %% Failure due to trying to join older node that
-                    %% doesn't define legacy_gossip will be handled
-                    %% in standard_join based on seeing a legacy ring.
-                    standard_join(Node, Rejoin, Auto)
-            end
+            %% Failure due to trying to join older node that
```
jrwest: if we remove all of legacy gossip this comment no longer applies

src/gen_nb_server.erl
seancribbs, line 128: I do remember looking into `prim_inet`, I did not see any reason why an async accept would not return `{ok, integer()}`, but dialyzer complained because the `integer()` would vary.

src/riak_core_apl.erl
```diff
@@ -42,16 +42,17 @@
 -type preflist2() :: [{{index(), node()}, primary|fallback}].
 -type iterator() :: term().
 -type chashbin() :: term().
+-type docidx() :: <<_:160>>.
```
jrwest, line 45: we have a lot of types defined for this same thing. one, less accurate example, https://github.com/basho/riak_core/blob/develop/src/chash.erl#L68

src/riak_core_bg_manager.erl
```diff
@@ -403,10 +403,10 @@ query_resource(Resource, Types) ->
 %%%
 
 -record(state,
-        {info_table:: ets:tid(),          %% TableID of ?BG_INFO_ETS_TABLE
-         entry_table:: ets:tid(),         %% TableID of ?BG_ENTRY_ETS_TABLE
-         enabled :: boolean(),            %% Global enable/disable switch, true at startup
-         bypassed:: boolean()             %% Global kill switch. false at startup
+        {info_table:: ets:tid() | atom(),  %% TableID of ?BG_INFO_ETS_TABLE
```
jrwest: iirc `ets:tab()` is equivalent to `ets:tid() | atom()`

src/riak_core_gossip_legacy.erl
```diff
@@ -153,7 +153,8 @@ code_change(_OldVsn, State, _Extra) ->
 schedule_next_gossip() ->
```
jrwest: when legacy gossip is gone maybe this file can be removed?

seancribbs: Indeed.

src/riak_core_node_watcher.erl
```diff
@@ -325,8 +325,7 @@ handle_info({'DOWN', Mref, _, _Pid, _Info}, State) ->
 
             %% Update our list of active services and ETS table
             Services = ordsets:del_element(Id, State#state.services),
-            S3 = S2#state { services = Services },
-            local_update(S3),
+            S3 = local_update(S2#state { services = Services }),
```
jrwest, line 328: I think there is a case here where we can lose the correct state because `local_update/1` may return `ok` [1]

[1] https://github.com/basho/riak_core/blob/bugfix/realtalk/dialyze/src/riak_core_node_watcher.erl#L487-L490

src/riak_core_bg_manager.erl
seancribbs: Addressed in e320072

src/riak_core_node_watcher.erl
seancribbs, line 328: Notice that `broadcast` is called after the `case` statement in that function.

jrwest, line 328: ah, yes. my mistake

evanmcc, line 328: Was just looking at the top of the thread, and wanted to note that you can
add -compile(inline_list_funcs). to the top to get the LC optimizations
added to standard lists module calls.

seancribbs, line 328: @evanmcc Interesting. Should we be doing that globally?

seancribbs, line 328: Also, the other argument for LCs even when the return value is not used is clarity/visual noise reduction.

evanmcc, line 328: I don't know.  It's something that I'd need to test.  Maybe we can enable
it at the start of the next dev cycle so we have a long time to hopefully
run into any issues.

src/riak_core_apl.erl
seancribbs, line 45: Addressed in ceec2d9

src/hashtree.erl
jtuple, line 634: I'm generally against changing behavior that works today and has been in production for a year. So, I'm a bit torn.

In any case, AAE is designed to be lossy. Dropping updates/etc is no big deal. However, if I can't put to a LevelDB instance, something's seriously wrong. I imagine future attempts to iterate over the LevelDB instance (eg. for an exchange) would fail as well. Hard to imagine a scenario where `eleveldb:put` fails, but other operations against the tree work just fine. So, this is seems to be an error we might as well fail against.

If it somehow ends up breaks things in production, it's a 1 line change in a 2.0.1, right? ;)

reiddraper, line 634: We can always pattern match on the output, or put it in a try/catch too.

seancribbs, line 634: I'm happy to switch it to a case clause and have it spew the error to the logs.

src/riak_core_apl.erl
```diff
@@ -35,23 +35,24 @@
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
--type index() :: non_neg_integer().
+-type index() :: chash:index_as_int().
 -type n_val() :: non_neg_integer().
 -type ring() :: riak_core_ring:riak_core_ring().
 -type preflist() :: [{index(), node()}].
 -type preflist2() :: [{{index(), node()}, primary|fallback}].
```
jrwest: we're probably nearing the point where we should try to stop making new changes, but couldn't help but comment that i really wish this were something like `preflist_ann()` or `annotated_preflist()` rather than `preflist2()`.

seancribbs: That makes sense, will fix.

Sean Cribbs

seancribbs: Addressed in 00086a7

src/riak_core_ring.erl
```diff
@@ -185,65 +175,23 @@
...
 %% Public API
 %% ===================================================================
 
-%% @doc Returns true if the given ring is a legacy ring.
-legacy_ring(#chstate{}) ->
-    true;
-legacy_ring(_) ->
-    false.
-
 %% @doc Upgrade old ring structures to the latest format.
```
jrwest, line 184: it might be worth adding a comment to `upgrade/1` and `downgrade/1` as to why the functions do basically nothing. long term it would probably be nice to remove them (as any future upgrade/downgrade may work differently)

src/riak_core_ring_manager.erl
```diff
@@ -625,8 +618,8 @@ prune_write_notify_ring(Ring, State) ->
 
 prune_write_ring(Ring, State) ->
     riak_core_ring:check_tainted(Ring, "Error: Persisting tainted ring"),
-    riak_core_ring_manager:prune_ringfiles(),
-    do_write_ringfile(Ring),
+    ok = riak_core_ring_manager:prune_ringfiles(),
+    _ = do_write_ringfile(Ring),
```
jrwest, line 622: nitpick note: above we match on `ok =` and here we do not. seems like in this case its equally reasonable to crash.

src/riak_core_vnode.erl
```diff
@@ -375,6 +376,11 @@ vnode_handoff_command(Sender, Request, ForwardTo,
             riak_core_vnode_worker_pool:handle_work(Pool, Work, From),
             continue(State, NewModState);
         {forward, NewModState} ->
+            %% FIXME: vnode_forward returns a gen_fsm state return
```
jrwest, line 379: i believe this was discussed and addressed. does this comment still apply?

jrwest:
seeing some issues running `verify_dynamic_ring` against this branch. lots of log messages like the one below. investigating...

```
20:20:07.354 [error] Error: riak_core_ring/ring_ready called on tainted ring
```

update 1:

same issue with `verify_build_cluster`, in this case, the log messages are in the riak logs only.

update 2:

I think i've isolated the issue to some of the recent legacy gossip removal changes. the errors do not appear when rolling back to a5c244319

seancribbs:
Closing this in favor of a PR that excludes:
- Changes to `riak_core_security`, since that changed several times during the course of this work.
- Removal of anything related to legacy gossip, which should be separate.

PR #529, opened 2014-02-13
Treat users and groups as distinct concepts

macintux:
To make it easier to reason about and document the new security behavior, this PR splits roles into users and groups.
- Users (but not groups) can be given authentication information (sources).
- Groups (but not users) can be used to cascade privileges to other groups and users.

Where the code does not or cannot distinguish between users and groups, I continue to refer to roles, but the user interface is constrained to users and groups.

There are 3 projects impacted by this change, all with the same branch name (`jrd-security-role-bifurcation`):
- riak
- riak_core
- riak_test

All should be merged at the same time. I will be filing PRs for each.

/cc @Vagabond

### DIFF
--- a/src/riak_core_console.erl
+++ b/src/riak_core_console.erl
@@ -24,6 +24,7 @@
          stage_force_replace/1, print_staged/1, commit_staged/1,
          clear_staged/1, transfer_limit/1, pending_claim_percentage/2,
          transfers/1, add_user/1, alter_user/1, del_user/1,
+         add_group/1, alter_group/1, del_group/1,
          add_source/1, del_source/1, grant/1, revoke/1,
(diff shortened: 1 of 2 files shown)


src/riak_core_console.erl
```diff
@@ -825,8 +827,14 @@ check_limit(Str) ->
     end.
 
 add_user([Username|Options]) ->
-    try riak_core_security:add_user(list_to_binary(Username),
-                                     parse_options(Options)) of
+    add_role(Username, Options, add_user).
```
Vagabond: You could also do fun add_user/2 and not use the apply()

macintux: Good point, cleaner, will update

macintux:
Superseded by #534, closing

PR #530, opened 2014-02-13
Bugfix/mw/ease bg manager test running

lordnull:
Added pulse test for potential race condition with table manager and resolution there-of.


lordnull:
This is now a pointless pr, and thus can be closed.

PR #531, opened 2014-02-13
Removed table manager.

lordnull:
The background manager tables are now created in the start/1 and
start_link/1 tables as public. This puts the owner as the supervisor.
If the supervisor dies, the system needs a clean restart anyway.
Minimal changes to the tests were needed.

This supersedes #530 .


evanmcc:
Code looks good.  All riak_core tests pass before and after merged.  Manually ran the eqc test for a while with no errors.

:+1:

PR #532, opened 2014-02-13
Attempt to enforce role name uniqueness

macintux:
One proposal for coping with naming conflicts between groups and users for purposes of granting permissions: don't allow conflicts (so far as we can determine).

When strong consistency comes to cluster metadata, we can enforce this more strongly, but for now we can supplement this by adding documentation instructing ops not to create groups and users by the same name.

/cc @Vagabond

### DIFF
--- a/src/riak_core_security.erl
+++ b/src/riak_core_security.erl
@@ -374,13 +374,18 @@ authenticate(Username, Password, ConnInfo) ->
 add_user(Username, Options) ->
     case user_exists(Username) of
         false ->
-            case validate_options(Options) of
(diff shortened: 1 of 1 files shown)


macintux:
This PR will probably be rejected in favor of #533

jrwest:
For the same reason as other security/bucket-type buckets found by @macintux, this has been marked for milestone `2.0-RC`

EDIT: wrong milestone, was `2.0.1`, meant `2.0-RC`

macintux:
This can be closed now. #533 has separated the concepts so name uniqueness is no longer required.

PR #534, opened 2014-02-19
Treat users and groups as distinct concepts (extended)

macintux:
(This consolidates #529 and #533.)

To make it easier to reason about and document the new security behavior, this PR splits roles into users and groups.
- Users (but not groups) can be given authentication information (sources).
- Groups (but not users) can be used to cascade privileges to other groups and users.

Where the code does not or cannot distinguish between users and groups, I continue to refer to roles, but the user interface is constrained to users and groups.

There are 3 projects impacted by this change:
- riak (branch `jrd-security-role-bifurcation`) basho/riak#489
- riak_core (branch `jrd-security-grant-namespace`)
- riak_test (branch `jrd-security-role-bifurcation`) basho/riak_test#538

All should be merged at the same time.

/cc @Vagabond

### DIFF
--- a/src/riak_core_security.erl
+++ b/src/riak_core_security.erl
@@ -20,10 +20,12 @@
 -module(riak_core_security).
 
 %% printing functions
--export([print_users/0, print_sources/0, print_user/1]).
+-export([print_users/0, print_sources/0, print_user/1,
+         print_groups/0, print_group/1]).
 
 %% API
 -export([authenticate/3, add_user/2, alter_user/2, del_user/1,
+         add_group/2, alter_group/2, del_group/1,
          add_source/4, del_source/2,
          add_grant/3, add_revoke/3, check_permission/2, check_permissions/2,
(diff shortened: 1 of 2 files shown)


src/riak_core_security.erl
```diff
@@ -591,31 +708,29 @@ status() ->
 
 add_revoke_int([], _, _) ->
     ok;
-add_revoke_int([User|Users], Bucket, Permissions) ->
-    UserGrants = riak_core_metadata:get({<<"security">>, <<"grants">>}, {User,
-                                                                         Bucket}),
+add_revoke_int([{Name, RoleType}|Roles], Bucket, Permissions) ->
+    Prefix = {<<"security">>, list_to_binary(atom_to_list(RoleType) ++ "grants")},
```
Vagabond: Maybe a pattern matching function for converting the roletype atom to a binary here, instead of this?

Vagabond:
Can you squash and rebase this off of develop, please?

Vagabond:
+1 to merge.

PR #535, opened 2014-02-20
Change multiple LCs into a single ets:match_delete/2.

seancribbs:
Cleans up that one obnoxious function clause in the background manager.

PR #538, opened 2014-02-28
Enchance/mw/bg manager

lordnull:
Add tests for coverage of exported api, and removed unused/old code in the process.

### DIFF
--- a/src/riak_core_bg_manager.erl
+++ b/src/riak_core_bg_manager.erl
@@ -167,22 +167,42 @@ enabled() ->
 %%      has not already been registered, this will have no effect.
 -spec enable(bg_resource()) -> enabled | unregistered | bypassed.
 enable(Resource) ->
-    gen_server:call(?SERVER, {enable, Resource}).
+    case gen_server:call(?SERVER, {enable, Resource}) of
+        {unregistered, Resource} ->
+            unregistered;
+        Else ->
+            Else
+    end.
 
 %% @doc Disable handing out resource of the given kind.
 -spec disable(bg_resource()) -> disabled | unregistered | bypassed.
 disable(Resource) ->
-    gen_server:call(?SERVER, {disable, Resource}).
+    case gen_server:call(?SERVER, {disable, Resource}) of
+        {unregistered, Resource} ->
+            unregistered;
+        Else ->
+            Else
+    end.
 
 -spec enabled(bg_resource()) -> enabled | disabled | bypassed.
 enabled(Resource) ->
-    gen_server:call(?SERVER, {enabled, Resource}).
+    case gen_server:call(?SERVER, {enabled, Resource}) of
+        {unregistered, Resource} ->
+            unregistered;
+        Else ->
+            Else
+    end.
 
 %% @doc Disable handing out resource of the given kind. If kill == true,
 %%      processes that currently hold the given resource will be killed.
 -spec disable(bg_resource(), boolean()) -> disabled | unregistered | bypassed.
 disable(Resource, Kill) ->
-    gen_server:call(?SERVER, {disable, Resource, Kill}).
+    case gen_server:call(?SERVER, {disable, Resource, Kill}) of
+        {unregistered, Resource} ->
+            unregistered;
+        Else ->
+            Else
+    end.
 
 %%%%%%%%%%%
 %% Lock API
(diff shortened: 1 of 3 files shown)


cmeiklejohn:
Is this targeted for 2.0 or 2.1?

src/riak_core_bg_manager.erl
```diff
@@ -176,22 +167,42 @@ enabled() ->
...
+            Else
+    end.
 
 %% @doc Disable handing out resource of the given kind.
 -spec disable(bg_resource()) -> disabled | unregistered | bypassed.
 disable(Resource) ->
-    gen_server:call(?SERVER, {disable, Resource}).
+    case gen_server:call(?SERVER, {disable, Resource}) of
+        {unregistered, Resource} ->
+            unregistered;
```
jrwest, line 182: what's the benefit of changing the API here. I understand the information is redundant but is there a reason? Also, this code path seems repeated in many places.

src/riak_core_bg_manager.erl
```diff
@@ -141,11 +137,6 @@ start_link() ->
     _ = maybe_create_ets(),
     gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
 
-%% Test entry point to start stand-alone server
-start() ->
```
jrwest, line 145: this is useful for local testing

src/riak_core_bg_manager.erl
```diff
@@ -421,11 +440,7 @@ query_resource(Resource, Types) ->
                   {stop, term()}.
 init([]) ->
     lager:debug("Background Manager starting up."),
-    State = #state{info_table=?BG_INFO_ETS_TABLE,
```
jrwest, line 424: i actually find this more confusing. having the defaults in `init` makes it more clear imo

src/riak_core_bg_manager.erl
```diff
@@ -758,8 +760,6 @@ update_resource_info(Resource, Fun, Default, State=#state{info_table=TableId}) -
     State.
 
 %% @doc Throws unregistered for unknown Resource
-resource_info(_Resource, State) when ?NOT_TRANSFERED(State) ->
```
jrwest, line 761: ah good call. all the `?NOT_TRANSFERED` stuff is dead code since removing the table manager

test/bg_manager_tests.erl
```diff
@@ -65,35 +65,199 @@ bg_mgr_test_() ->
...
+                         ?assertEqual(max_concurrency, ?BG_MGR:get_token(T)),
+                         ?assertEqual(1, length(?BG_MGR:all_locks(lock_a))),
+                        Locker ! ok
                  end},
 
-                {"bypass API",
-                 fun() ->
-                         %% bypass API
-                         riak_core_bg_manager:bypass(true),
+                {"lock api", [
```
jrwest, line 85: i dont think we should be adding to these tests. I'm not even sure why they are still here. Can this be extended to the eqc test?

src/riak_core_bg_manager.erl
lordnull, line 182: The spec did not match the actual return value. Looking at where this was used, it seemed the better option was to change the code so it matched the spec rather than the other way around. This was discovered while writing the unit test for the function.

src/riak_core_bg_manager.erl
lordnull, line 145: As is using the function that will be used in production and unlinking to it. The fewer differences between production and testing, the better. If this is useful in testing, and intended only for testing, it should be wrapped in an -ifdef.

jrwest, line 145: the `ifdef` is more correct but means you have to do testing using beams from the `.eunit` directory. thats fine, just something to remember. fwiw, this is a pattern we use in several places -- but again I think wrapping it in the `ifdef` makes sense

src/riak_core_bg_manager.erl
jrwest, line 182: hrm, neither dialyzer [1] nor the eqc test caught this. one should...once again, i'd like to move away from the unit tests (and see them removed once any non-redundant parts are part of the eqc test). the unit tests caused several problems along the way during review.

[1] https://github.com/basho/riak_core/pull/528/files#diff-9

lordnull, line 182: Double checking, this breaks the eqc, so fixing the spec would been the right way to go.

jrwest, line 182: ok wfm. i'm guessing btw that dialyzer can't figure it out bc of the gen_server message boundary.

lordnull:
This has been punted upon until at least 2.1.

jrwest:
related PR: https://github.com/basho/riak_core/pull/535

jrwest:
For now marking as milestone `2.0.1`.

@lordnull can this be closed? some of these changes overlapped with other PRs and others, and iirc broke tests or were not necessary.

lordnull:
This can be closed mainly since it ended up duplicating some effort.

PR #539, opened 2014-02-28
Fixes for YZ dialzyer warnings

bowrocker:
- Replace the function() type definitions with atom() in the multi-call\* family of functions which were causing dialyzer warnings in yokozuna (and possibly other places). Atoms are being passed in, not functions.
- Introduce vlock:sort/1, which simply does a lists:sort on a vlock. Because the vclock type is an opaque type, doing the sort outside the vlock module causes dialyzer to warn as it requires looking at the internals of the vlock type, which is exactly what the -opaque keyword forbids.

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -401,7 +401,7 @@ rpc_every_member(Module, Function, Args, Timeout) ->
 
 %% @doc Same as rpc_every_member/4, but annotate the result set with
 %%      the name of the node returning the result.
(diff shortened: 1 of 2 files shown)


seancribbs:
Many of these are already covered by #528. Can we get on the same page?

jrwest:
yes please

bowrocker:
Closing as these are covered in https://github.com/basho/riak_core/pull/540.

PR #541, opened 2014-03-01
Remove legacy gossip

seancribbs:
This was extracted from #528.


jrwest:
moved to `2.0.1` since we ran into complications and decided to punt for `2.0`

jonmeredith:
Seems too major for a point - do early in 2.1 cycle.

jrwest:
IMO this is great debt to remove in a point. Doesn't change the API and most of the work had already been done. We just didn't have time to chase down a bug for the 2.0 release.

rzezeski:
I would also agree to removing in point.  This code has been dead for a while now.  Down with the tech debt!

jonmeredith:
What's the rationale for doing in a point?  I'd like to minimize the amount of work/diffs done on points and be focussed going forwards.  It seems like additional work to apply it to the 2.0 branch, and then reapply it to develop, given the code doesn't overly hurt anything, and we're not planning on refactoring anything on the point branch unless we have to.

seancribbs:
Realizing this is a refactor, I agree with @jonmeredith and think it should be 2.0.0 or 2.1.0. I'm happy to rebase and fix it for 2.0.0 if needed.

hmmr:
Rebased onto current develop, filed separately as #737 (RIAK-1799).

hmmr:
I believe this PR (and the branch) can be deleted, too?

cmeiklejohn:
Yeah, I believe so.  Let's just close for now and wait to delete until we are sure.

PR #544, opened 2014-03-04
Remove those obnoxious debug statements from bucket_fixup_test

seancribbs:
...and reformat the file

### DIFF
--- a/test/bucket_fixup_test.erl
+++ b/test/bucket_fixup_test.erl
@@ -44,40 +44,35 @@ fixup(_, Props) ->
 
 fixup_test_() ->
     {setup,
-        fun() ->
-                application:set_env(riak_core,ring_creation_size, 4),
-                application:set_env(riak_core, bucket_fixups, [{someapp, ?MODULE}]),
-                application:set_env(riak_core, default_bucket_props, [])
-        end,
-        fun(_) ->
-                application:unset_env(riak_core, ring_creation_size),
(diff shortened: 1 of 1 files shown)


test/bucket_fixup_test.erl
```diff
@@ -89,68 +84,48 @@ io:format(user, "BBOT DBG: ~p ~p ~p\n", [?MODULE, ?LINE, Ls2]),
...
-io:format(user, "BBOT DBG: ~p ~p\n", [?MODULE, ?LINE]),
                              receive
                                  done ->
-io:format(user, "BBOT DBG: ~p ~p\n", [?MODULE, ?LINE]),
                                      ok
                              end
                      end),
-io:format(user, "BBOT DBG: ~p ~p\n", [?MODULE, ?LINE]),
              receive
                  ready -> ok
```
jrwest: While testing this branch on my machine the test hung indefinitely on this `receive`. This happens on every run of this branch for some reason but not on develop even though the code is the same (lol scheduling). Adding an `after` to the receive makes it easy to track down. We should at least not hang up a test runner indefinitely and add a timeout. We probably don't need to make any other fixes to the test in this PR. If it continues to fail (timeout) then that can be addressed separately -- unless you feel like fixing it now @seancribbs ;)

jrwest:
+1 merge. test passing w/o the hang now.

PR #546, opened 2014-03-04
Improve output from (primarily) security commands

macintux:
- Fix a few missing EOLs on output, some in security, some in other commands
- Fix `check_grant_blockers` to discard the `error` tuple label from `verify_permissions` before adding the message to the collected errors
- Stop swallowing errors in `riak_core_console`
- Add success messages
- Translate errors in `riak_core_console`
- Drop problematic single-role clause for `add_grant`/`add_revoke`, was only used by unit tests, could result in an infinite loop if invoked incorrectly from shell
  - Fix corresponding unit tests
- Unintentionally drop some spurious whitespace

### DIFF
--- a/src/riak_core_console.erl
+++ b/src/riak_core_console.erl
@@ -332,7 +332,7 @@ print_arrowbox(SrcAtom, TargetAtom, Objs, Bytes, Progress) ->
                  "" -> string:centre(Target0, 25);
                  _ -> Target0
              end,
-    
+
     ToFrom = riak_core_format:fmt("~25s ~10s ~25s",
                                   [Src, ?ARROW, Target]),
     Width = length(ToFrom),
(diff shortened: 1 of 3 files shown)


macintux:
Forgot to /cc @Vagabond

src/riak_core_console.erl
```diff
@@ -839,9 +895,12 @@ add_group([Groupname|Options]) ->
 
 add_role(Name, Options, Fun) ->
     try Fun(list_to_binary(Name), parse_options(Options)) of
-        ok -> ok;
+        ok ->
+            io:format("Succesfully added ~s~n", [Name]),
```
Vagabond, line 899: http://www.linfo.org/rule_of_silence.html

src/riak_core_security.erl
```diff
@@ -1088,8 +1080,8 @@ flatten_once(List) ->
 
 check_grant_blockers([], [], ok) ->
     none;
-check_grant_blockers([], [], Error) ->
-    Error;
+check_grant_blockers([], [], {error, Error}) ->
```
Vagabond, line 1083: Isn't the pattern match here a little redundant?

macintux, line 1083: I debated on that, but since I screwed up the follow-on function heads by not matching on just the error message itself, I liked the symmetry here of doing the same.

Vagabond:
+1ed while I mourn the death of UNIX.

src/riak_core_security.erl
reiddraper, line 1083: fwiw, dialyzer is going to complain about this. the pattern will never match

reiddraper, line 1083: ignore me, i misread the pattern match as still having the removed code. carry on

PR #551, opened 2014-03-07
Make print-user/group look like print-users/groups

macintux:
- `print-user` (likewise `-group`) is just a single-entity version of `print-users` now
- Create `print-grants <user|group>` to replace the old `print-user/group`
- Make `print_grants` much less noisy by making the `console_table` code smarter
- Allow `user/` and `group/` prefixes for print-grants, also fix old render bug with `print-grants` and groups
- Extend unit tests

PRs exist for the related repositories:
- basho/riak#504
- basho/riak_ee#227
- basho/riak_test#548

/cc @Vagabond


Vagabond:
+1

PR #552, opened 2014-03-08
Add hashtree function for returning all buckets.

cmeiklejohn:
Prototype a hashtree function for returning all buckets as an orddict to allow for batching of requests which need to iterate all of the buckets in the tree.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -493,20 +493,23 @@ esha_final(Ctx) ->
 -endif.
 
 -spec get_buckets(integer(), orddict(), hashtree()) -> orddict().
-get_buckets(Level, Results, #state{width=W,levels=L}=State) ->
-    ByLevel = erlang:trunc(W / (L - 1)),
-    Buckets = case Level of
-        1 ->
-            [0];
-        X ->
-            lists:seq(0, erlang:trunc(ByLevel * (X - 1)))
-    end,
+get_buckets(1, Results, State) ->
+    get_buckets(1, [0], Results, State);
+
+get_buckets(Level, Results,
+            #state{levels=L, segments=S, width=W}=State) ->
+    Multiple = math:pow(W, (L - Level + 1)),
+    Upper = erlang:trunc(S / Multiple),
+    Buckets = lists:seq(0, Upper - 1),
+    get_buckets(Level, Buckets, Results, State).
+
+-spec get_buckets(integer(), list(), orddict(), hashtree()) -> orddict().
+get_buckets(Level, Buckets, Results, State) ->
     lists:foldl(fun(Bucket, Orddict) ->
                 orddict:store({Level, Bucket},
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
cc: @jtuple

src/hashtree.erl
```diff
@@ -479,6 +492,22 @@ esha_final(Ctx) ->
     crypto:sha_final(Ctx).
 -endif.
 
+-spec get_buckets(integer(), orddict(), hashtree()) -> orddict().
+get_buckets(Level, Results, #state{width=W,levels=L}=State) ->
```
Vagabond: So when level=3 that means grab the first three levels, not just the third level?

Also, the math below is a little obtuse, maybe @jtuple can eyeball it.

cmeiklejohn: No, it's just get that level for this function.  The other function generates a sequence up to the provided level, and then folds the results into an orddict.  

See: https://github.com/basho/riak_core/pull/552/files#diff-61653aa3ba5c9c8b67d8e8953ea7e8c4R452

Yes, @jtuple, please check this math.

Vagabond: Oh yes, I forgot that.

Vagabond:
I'm +1 on this, but I want @jtuple to sign off on that tree math before it merges.

cmeiklejohn:
@jtuple thoughts?

src/hashtree.erl
```diff
@@ -479,6 +492,22 @@ esha_final(Ctx) ->
...
+    ByLevel = erlang:trunc(W / (L - 1)),
+    Buckets = case Level of
+        1 ->
+            [0];
+        X ->
+            lists:seq(0, erlang:trunc(ByLevel * (X - 1)))
```
engelsanchez: So ByLevel here is definitely a whole number. If you are worried about Level not being a whole number maybe add a guard instead? If not worried about that, the truncation here is just distracting.

cmeiklejohn: How is it definitely a whole number?

cmeiklejohn: Oh wait, I thought this comment was on the other line.

cmeiklejohn:
Once merged, please merge: basho/riak_kv#863 and basho/riak_repl#557.

cmeiklejohn:
Updated PR with commit https://github.com/basho/riak_core/commit/2c488b7017596df73a9e230116af34bb0ba179a7 which addresses the incorrect math in the original commit.

cmeiklejohn:
Considering we are not merging the `riak_repl` and `riak_kv` changes which require this support in `riak_core`, we're not going to merge this for the 2.0 RC.  Closing.

cmeiklejohn:
Please do not delete this branch, as this code may be retargeted against the 2.1 release.

PR #553, opened 2014-03-17
Allow handoff sender to abort handoff by throw'ing from fold fun

slfritchie:
Depends upon: https://github.com/basho/riak_kv/pull/869

There's no reason why the handoff sender fold fun has to continue
folding when there's a TCP error.  Instead, throw the #ho_acc{}
record with its error={error,Reason} field intact.

If we force the handoff receiver to exit (using exit(HO_pid, kill)),
then we simply see the following logged on the sender side:

```
2014-03-17 15:53:30.949 [error] <0.27207.2>@riak_core_handoff_sender:start_fold:266 hinted_handoff transfer of riak_kv_vnode from 'dev1@127.0.0.1' 1415829711164312202009819681693899175291684651008 to 'dev2@127.0.0.1' 1415829711164312202009819681693899175291684651008 failed because of closed
2014-03-17 15:53:30.950 [error] <0.220.0>@riak_core_handoff_manager:handle_info:289 An outbound handoff of partition riak_kv_vnode 1415829711164312202009819681693899175291684651008 was terminated for reason: {shutdown,{error,closed}}
```

Also, change the lager:info() message when the handoff receiver
gets a tcp_error message that _includes_ the reason and
distiguishes itself from other handoff receiver log messages.

Also, one whitespace correction.

### DIFF
--- a/src/riak_core_handoff_receiver.erl
+++ b/src/riak_core_handoff_receiver.erl
@@ -81,10 +81,10 @@ handle_info({tcp_closed,_Socket},State=#state{partition=Partition,count=Count,
     lager:info("Handoff receiver for partition ~p exited after processing ~p"
                           " objects from ~p", [Partition, Count, Peer]),
     {stop, normal, State};
-handle_info({tcp_error, _Socket, _Reason}, State=#state{partition=Partition,count=Count,
-                                                        peer=Peer}) ->
+handle_info({tcp_error, _Socket, Reason}, State=#state{partition=Partition,count=Count,
+                                                       peer=Peer}) ->
     lager:info("Handoff receiver for partition ~p exited after processing ~p"
-                          " objects from ~p", [Partition, Count, Peer]),
+                          " objects from ~p: TCP error ~p", [Partition, Count, Peer, Reason]),
     {stop, normal, State};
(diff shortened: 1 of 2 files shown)


jrwest:
also an existing issue, so I've marked for `2.0.1`. Since we have a backport for `1.4` already though maybe we should consider moving to `2.0-RC`.

slfritchie:
I am fully in favor of 2.0-RC.

jrwest:
@jonmeredith opinions?

macintux:
Simple testing is going well using iptables to force connection resets.

Without the throw, fold continues; with the throw, fold also continues, but just for a short time until the error propagates. Removing the iptables rule allows the handoff to automatically restart and successfully complete.

macintux:
Scott, if you happen to be around, we'll need this retargeted for 2.0. I was thinking about doing the branch rebase myself, but realized this morning that the PR itself will need to be edited or redone entirely.

macintux:
Ran more testing by killing vnode processes on the target system, fold canceled quickly and everything recovered gracefully.

This is :+1: once the branch and PR are rebased/rejiggered.

macintux:
Never mind Scott, #626 replaces this. Thanks.

PR #555, opened 2014-03-17
Fix overload test

andrewjstone:
- Wait for all messages to be processed by fake vnode in order to get
  accurate count
- Abstract out some common code to update_msg_counter()


Vagabond:
+1, lets merge this thing.

PR #556, opened 2014-03-18
Prevent a bucket type named 'any' from being created

macintux:
In order for the security CLI to be able to grant permissions to any bucket type, we need to prevent the administrator from creating a bucket type named "any".

This is dependent on an unfinished PR that would change the grant syntax to use lower-case keywords; this may need to be changed to "ANY" before approval and merging.

/cc @jrwest

### DIFF
--- a/src/riak_core_bucket_type.erl
+++ b/src/riak_core_bucket_type.erl
@@ -143,9 +143,14 @@ defaults() ->
 %% calls and the type is not active. An error will also be returned if the properties
 %% are not valid. Properties not provided will be taken from those returned by
 %% {@see defaults/0}.
+%%
+%% `default` and `any` are reserved names; `any` because the security subsystem needs a
+%% way to refer to any bucket type when granting permissions.
 -spec create(bucket_type(), bucket_type_props()) -> ok | {error, term()}.
 create(?DEFAULT_TYPE, _Props) ->
     {error, default_type};
+create(<<"any">>, _Props) ->
+    {error, reserved_name};
 create(BucketType, Props) when is_binary(BucketType) ->
     ?IF_CAPABLE(riak_core_claimant:create_bucket_type(BucketType,
                                                       riak_core_bucket_props:merge(Props, defaults())),


jrwest:
milestone `2.0-RC` because its a bug in an existing feature

src/riak_core_bucket_type.erl
```diff
@@ -143,9 +143,14 @@ defaults() ->
 %% calls and the type is not active. An error will also be returned if the properties
 %% are not valid. Properties not provided will be taken from those returned by
 %% {@see defaults/0}.
+%%
+%% `default` and `any` are reserved names; `any` because the security subsystem needs a
+%% way to refer to any bucket type when granting permissions.
 -spec create(bucket_type(), bucket_type_props()) -> ok | {error, term()}.
 create(?DEFAULT_TYPE, _Props) ->
     {error, default_type};
+create(<<"any">>, _Props) ->
```
jrwest, line 152: this isn't really how I imagined this. I had expected we would extend the [bucket validation code in riak_core](https://github.com/basho/riak_core/blob/develop/src/riak_core_bucket_props.erl#L42-L45). Now that I see it I'm not too opposed, but at a minimum we need to open an issue to extend [btypes_eqc](https://github.com/basho/riak_core/blob/develop/test/btypes_eqc.erl) or [its successor](https://github.com/basho/riak_core/issues/563) to cover this case.

jrwest, line 152: actually I do prefer the validator code I think because technically there is a [code path](https://github.com/basho/riak_core/blob/develop/src/riak_core_claimant.erl#L171-L172) that can circumvent this validation as implemented now.

macintux, line 152: The current property validation code is heavily tied to properties, in fact IIRC it returns processed properties after. I didn't see any clean way to validate anything other than props.

I'll take another look.

jrwest, line 152: [the first and second arguments to `validate/4`](https://github.com/basho/riak_core/blob/develop/src/riak_core_bucket_props.erl#L38-L39) will give you what you need

macintux:
I was looking at `riak_kv_bucket:validate`; `riak_core_bucket_props:validate` is much easier to figure out how to reject creating the "any" bucket type.

I am a bit puzzled: `riak_kv_bucket:validate` is supposed to return `{error, [{atom(), atom()}]}` per its spec, but I see several `{atom(), string()}` tuples originating in `validate_core_prop` for error conditions.

Anyway, new PR should be forthcoming.

jrwest:
closing since this is superseded by #567

Issue #557, opened 2014-03-18
Handoff key filtering (repair or ring resize handoff) does not send periodic sync messages to keep TCP connection alive


slfritchie:
Confirmed in the lab: while doing repair handoff, the filtering done appears to prevent sending batches of handoff data to the receiver.  The receiver's default timeout is 5 minutes, but it's see-it-with-my-own-eyes possible to fold over Bitcask data at over 150MBytes per second and not have enough data to send to trigger the receiver's 5 minute idle timeout.

After the reciever has timed out, the #ho_acc record looks like this (from a stack trace from `riak-admin cluster-info`:

```
y(0)     {ho_acc,6,ok,#Fun&lt;riak_core_repair.0.61411323>,riak_kv_vnode,&lt;0.588.0>,#Port&lt;0.8040>,{593735040165679310520246963290989976735222595584,616571003248974668617179538802181898917346541568},{ho_stats,{1395,157067,590863},undefined,125,128161535},gen_tcp,286632,158788528786,[&lt;&lt;803113 bytes>>],1,803113,25,repair,undefined,undefined}
```

Note that element 2, the ack number, is 6.  And it remains at 6 for minutes at a time, confirming the hypothesis that the batch isn't big enough to send yet to the remote side.

When the batch finally becomes big enough, we see that the ack number only reaches 7: only one send is required to change the element 3 from `ok` to either `{error,closed}` or `{error,enotconn}`, depending on the timing.

```
{ho_acc,7,{error,closed},#Fun<riak_core_repair.0.61411323>,riak_kv_vnode,<0.588.0>,#Port<0.8040>,{593735040165679310520246963290989976735222595584,616571003248974668617179538802181898917346541568},{ho_stats,[....]
```

Packet capture by tcpdump confirms that there is 5 minutes of inactivity before the receiver closes/sends FIN.

jrwest:
changed to milestone `2.0.1` for same reason as https://github.com/basho/riak_core/pull/560#issuecomment-38460264. Happy to revist again.

martinsumner:
https://github.com/basho/riak_core/pull/995

PR #560, opened 2014-03-20
Handoff sender sends sync periodically

slfritchie:
If we can discover the `handoff_receive_timeout value`, then use 1/3rd that interval to send a sync/heartbeat thingie, otherwise 60 seconds (for infinity & undefined).

Purpose: to avoid unhelpful timeouts when the handoff sender is filtering (e.g. repair or ring resizing handoff) and encounters zillions of keys to filter and thus doesn't send anything to the receiver for a while.  A workaround is to set the receiver's timeout to be extremely long, but that's a kludge IMHO.

### DIFF
--- a/src/riak_core_handoff_sender.erl
+++ b/src/riak_core_handoff_sender.erl
@@ -63,6 +63,7 @@
           item_queue_byte_size :: non_neg_integer(),
 
           acksync_threshold    :: non_neg_integer(),
+          acksync_timer        :: timer:tref(),
 
           type                 :: ho_type(),
 
@@ -216,8 +217,10 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
            total_objects=TotalObjects,
            total_bytes=TotalBytes,
            stats=FinalStats,
+           acksync_timer=TRef,
            notsent_acc=NotSentAcc} = AccRecord,
 
+         _ = timer:cancel(TRef),
          case ErrStatus of
              ok ->
                  %% One last sync to make sure the message has been received.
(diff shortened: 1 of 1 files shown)


slfritchie:
Thoughts @jrwest @engelsanchez @rzezeski?  Since starting this, I've decided that "feature" isn't strong enough: this is a bug serious enough to cause problems for Riak's timely replica reconstruction and (from what I've heard indirectly) ring resizing.

jrwest:
I agree this a bug and that raising the timeout is a workaround (read: kludge). However, we have to draw the line at some point and this bug has existed through several versions of Riak (since repair was introduced) and we are only running into it now. My vote would be to include it in a 2.0.x, x > 0 and if necessary deliver it as a patch in the meantime (although tbh at that point raising the timeouts is probably an equal amount of work for the user).

slfritchie:
@jrwest @jtuple @jonmeredith @andrewjstone To follow up on last week's core/KV meeting, I'm willing to open a PR for riak & riak EE config to raise the receiver timeout, but if I do that, I'm going to open it to something conservative, e.g. 24 hours.

Why?  Because it's possible for a double-node failure & vnode repair scenario to have the sender folding & doing nothing for a long long time.
1. Double failure somewhere in the cluster.
2. Force replace new machines to replace the dead ones.
3. Use vnode repair on the first new machine.
4. Use vnode repair on the 2nd new machine.  If any pair-wise vnode repair transfers to the 2nd machine are source from the 1st new machine, it's very possible for the 1st machine to fold over lots of data inside Bitcask that was written in arrival order, where all of the keys in that order were for a preflist that included the 1st new machine but do not include the 2nd.  So they'll all be filtered out by the sender's fold.

This scenario has already happened.  I'd like to see as few repeats as possible.

slfritchie:
Fixes https://github.com/basho/riak_core/issues/557

jrwest:
I still see this as an issue that has existed for several releases and can be fixed in a `2.0.1`. The milestone should either be `2.0-RC` (its not a data loss bug, which the `2.0` milestone indicates) or `2.0.1`. I've moved it to `2.0.1` for now, happy to revist.

Also, this has been seen in the wild with ring resizing, and in fact its probably more likely than repair, and I think a solution like this one is correct. However, I don't see changing the handoff protocol this late in the game as necessary, given known (kludgy) workarounds (raising the timeout and [this PR](https://github.com/basho/riak_core/pull/514)).

engelsanchez:
Wouldn't having a very long timeout also cause problems? Somebody's handoff will get stuck for way too long. Until the real fix comes in, why not set it to a reasonable time and expect CSEs will be involved with anybody that might possibly hit it for that long (that's a lot of data) and can recommend tune it up.  Don't ignore the damage done on the other side of the spectrum.

jrwest:
Oh I'm not advocating raising the default timeout. I agree it should only be done on suggestion in cases where we know this is the issue. These handoffs ultimately complete, when the timeout is raised.

kellymclaughlin:
We probably need to open a new PR for this change against the `2.0` branch if it is destined for `2.0.x`.

kellymclaughlin:
Opened #624 against the `2.0` branch.

kellymclaughlin:
Going through the nonsensical exercise of commenting on both PRs just so everyone sees them.

src/riak_core_handoff_sender.erl
```diff
@@ -281,11 +284,40 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
              gen_fsm:send_event(ParentPid, {handoff_error, Err, Reason})
      end.
 
+start_visit_item_timer() ->
+    Ival = case app_helper:get_env(riak_core, handoff_receive_timeout) of
```
kellymclaughlin, line 288: I think this `case` should just be replaced with a call to `get_handoff_receive_timeout/0`. It uses a default of `?TCP_TIMEOUT` which is defined as 60000.

src/riak_core_handoff_sender.erl
```diff
@@ -281,11 +284,40 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
...
+                   60*1000
+           end,
+    timer:send_interval(Ival, tick_send_sync).
+
+visit_item(K, V, Acc0 = #ho_acc{acksync_threshold = AccSyncThreshold}) ->
+    %% Eventually, a vnode worker proc will be doing this fold, but we don't
+    %% know the pid of that proc ahead of time.  So we have to start the
+    %% timer some time after the fold has started execution on that proc
+    %% ... like now, perhaps.
+    Acc = case get(is_visit_item_timer_set) of
```
kellymclaughlin, line 301: It seems like overkill to have to execute this `case` statement for every item in the fold during handoff. Since `start_visit_item_timer` calls `timer:send_interval`, it seems like we could call `start_visit_item_timer` right before starting the fold and cut this function down to just the `receive` block.

src/riak_core_handoff_sender.erl
kellymclaughlin, line 288: I retract my previous comment about this. I missed the fact that the function `get_handoff_receive_timeout` actually retrieves the value for `handoff_timeout`, not `handoff_receive_timeout`.

src/riak_core_handoff_sender.erl
kellymclaughlin, line 301: Sorry, misunderstood how this flowed at first glance. I retract my statement.

kellymclaughlin:
I modified the `verify_handoff` test to verify the efficacy of this change: https://github.com/basho/riak_test/pull/681

kellymclaughlin:
Once I get a review on my test changes and get that merged in I'm +1 on this change.

kellymclaughlin:
#627 is merged into the `2.0` branch, closing

PR #558, opened 2014-03-19
Make utf8 bucket types grantable

macintux:
Related to basho/riak_kv#868. This should be superseded by a broader utf8 security fix soon (branch `jrd-security-encoding`, under development).

/cc @jrwest

PR #562, opened 2014-03-21
Be more rigorous with security command-line arguments

macintux:
- Remove upper-case requirements on grant command
- Lower-case (at least for Latin-1 character set) all user/group names
- Restrict names from containing control characters, whitespace, most latin1 punctuation
- Block creation of names that would conflict with grant syntax
- Make `riak_core_console:grant/1` (and `revoke`) easier to read/modify by extracting common logic to `grant_int`
- Stop converting names to binary in `riak_core_console` so we can centrally apply lower-case, unicode conversions in `riak_core_security`
- Allow unicode names, render unicode output properly
- Catch and interpret badarg exception when a character outside latin1 space is specified as part of a source name (atoms won't allow unicode until R18)
- Lower-case all keys in options
- Be explicit with `unicode:characters_to_binary` at @jrwest's suggestion: we're converting from utf8 lists to utf8 binaries, so use the 3-ary version
- Convert (potentially utf-8) error messages from `check_permissions` to binary so the protocol buffers stack doesn't choke on them

Several other PRs are necessary to complete this:
- basho/riak_ee#229
- basho/riak#513
- basho/riak_test#555
- basho/riak_kv#874
- basho/node_package#119

Also related:
- #567 (prevent `any` from being used as a bucket type name, conflicting with granting permissions)

Should be able to be closed without merging if this goes through first:
- #558 

/cc @Vagabond @jrwest

### DIFF
--- a/src/riak_core_console.erl
+++ b/src/riak_core_console.erl
@@ -841,55 +841,54 @@ security_error_xlate({error, unknown_user}) ->
 security_error_xlate({error, unknown_group}) ->
     "Group not recognized";
 security_error_xlate({error, {unknown_permission, Name}}) ->
-    io_lib:format("Permission not recognized: ~s", [Name]);
+    io_lib:format("Permission not recognized: ~ts", [Name]);
 security_error_xlate({error, {unknown_role, Name}}) ->
-    io_lib:format("Name not recognized: ~s", [Name]);
+    io_lib:format("Name not recognized: ~ts", [Name]);
 security_error_xlate({error, {unknown_user, Name}}) ->
-    io_lib:format("User not recognized: ~s", [Name]);
+    io_lib:format("User not recognized: ~ts", [Name]);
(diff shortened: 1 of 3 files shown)


macintux:
One important note: today we don't accept non-ascii usernames via HTTP because there's no defined standard apparently and `riak_api_web_security.erl` assumes base64 encoding.

https://tools.ietf.org/id/draft-reschke-basicauth-enc-00.html offers a mechanism for supporting it, but haven't done much digging yet.

jrwest:
moved to milestone `2.0-RC` instead of `2.0-BETA`. These are bugs, mostly around input handling, found in something that is feature complete.

lordnull:
With all the pr's listed in this one used, a riak_ee install passes the http_security r_t.

lordnull:
:+1: cbdae61

macintux:
@borshop merge

PR #567, opened 2014-04-03
Do not allow a bucket type named 'any'

macintux:
In order for the security CLI to be able to grant permissions to any bucket type, we need to prevent the administrator from creating a bucket type named "any".

This is dependent on an unfinished PR that would change the grant syntax to use lower-case keywords; this may need to be changed to "ANY" before approval and merging.

**Note**: I don't have a registered copy of QuickCheck, nor would I know what to do with it if I had it, so the eqc fix is untested. Seems self-evidently correct based on my reading, but yolo.

/cc @jrwest 

(Supersedes #556)

### DIFF
--- a/test/btypes_eqc.erl
+++ b/test/btypes_eqc.erl
@@ -279,7 +279,7 @@ existing_type_name(#state{types=Types}) ->
     oneof(ExistingTypes).
 
 new_type_name() ->
-    binary(10).
+    ?SUCHTHAT(X, binary(10), X =/= <<"any">>).
 
 bucket_name() ->
     binary(10).
--- a/test/btypes_eqc.erl
+++ b/test/btypes_eqc.erl
@@ -315,6 +315,10 @@ prop_btype_invariant() ->
             aggregate(command_names(Cmds),
                       ?TRAPEXIT(
                          begin
+                             meck:new(riak_core_capability, []),
+                             meck:expect(riak_core_capability, get,
+                                         fun({riak_core, bucket_types}) -> true;
+                                            (X) -> meck:passthrough([X]) end),
                              os:cmd("rm -r ./btypes_eqc_meta"),
                              application:set_env(riak_core, claimant_tick, 4294967295),
                              application:set_env(riak_core, broadcast_lazy_timer, 4294967295),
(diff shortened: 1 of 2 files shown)


jrwest:
assigned to `2.0-RC` milestone

test/btypes_eqc.erl
```diff
@@ -279,7 +279,7 @@ existing_type_name(#state{types=Types}) ->
     oneof(ExistingTypes).
 
 new_type_name() ->
-    binary(10).
+    ?SUCHTHAT(X, binary(10), X =/= <<"any">>).
```
jrwest, line 282: it would actually be nice to extend the test to make sure the validation on `<<"any">>` works correctly, but for this PR this is sufficient. We should open an issue to try and get the work into https://github.com/basho/riak_test/pull/562

jrwest:
kind of overloading this PR but `btypes_eqc` has been failing since https://github.com/basho/riak_core/commit/c9aa007d09. I've fixed it here since its a minor change, and necessary to review this PR.

jrwest:
Instead of returning the single reserved name error, its preferable to continue validation and accumulate other errors. 3089d2a changes things around a bit to do this.

before 3089d2a:

```
$ riak-admin bucket-type create any '{"props": {"n_val": "abc", "consistent": "abc"}}'
Error creating bucket type any: [{reserved_name,
                                     "The name 'any' may not be used for bucket types"}]
```

after:

```
$ riak-admin bucket-type create any '{"props": {"n_val": "abc", "consistent": "abc"}}'
Error creating bucket type any: [{n_val,not_integer},
                                 {consistent,
                                     "<<\"abc\">> is not a valid value for consistent. use \"true\" or \"false\""},
                                 {reserved_name,
                                     "The name 'any' may not be used for bucket types"}]
```

side note: this output is ugly (EDIT: so much so I opened an issue https://github.com/basho/riak_kv/issues/893)

coderoshi:
+1. Agreed on @jrwest's points, but this is mergable in its current state.

```
OK, passed 1000 tests

29.86% {btypes_eqc,create_type,2}
29.68% {btypes_eqc,activate_type,1}
10.30% {btypes_eqc,get_type,1}
10.21% {btypes_eqc,type_status,1}
10.09% {btypes_eqc,update_type,2}
9.87% {btypes_eqc,set_bucket,2}
```

jrwest:
@coderoshi mind giving @borshop the +1 on the commmit?

coderoshi:
+1 3089d2a

jrwest:
@borshop merge please

PR #577, opened 2014-04-22
Update core:security:bucket() spec

reiddraper:
Buckets are binaries, not strings. See basho/riak_kv#920 for more
detail. This was discovered by the way kv was calling
riak_core_security:check_permission/2.


seancribbs:
+1 19be73a

reiddraper:
@borshop merge

PR #581, opened 2014-05-19
Correct return type information on riak_core_bucket:set_bucket/2

seancribbs:
In basho/riak_api#62, the [dialyzer check fails](http://buildbot.bos1/builders/test-riak_api/builds/13/steps/shell_5/logs/stdio) because it believes the spec on riak_core_bucket:set_bucket/2, meaning that `{error, no_type}` would completely cover any returns matching `{error, _}`. However, validation errors may also be returned from `set_bucket/2`, as a list  of atom pairs that resulted from the call to riak_core_bucket_props:validate/4.

### DIFF
--- a/src/riak_core_bucket.erl
+++ b/src/riak_core_bucket.erl
@@ -56,7 +56,7 @@ append_bucket_defaults(Items) when is_list(Items) ->
 %% @doc Set the given BucketProps in Bucket or {BucketType, Bucket}. If BucketType does not
 %% exist, or is not active, {error, no_type} is returned.
 -spec set_bucket(binary() | {riak_core_bucket_type:bucket_type(), binary()}, [{atom(), any()}]) ->
-                        ok | {error, no_type}.
+                        ok | {error, no_type | [{atom(), atom()}]}.
 set_bucket({<<"default">>, Name}, BucketProps) ->
     set_bucket(Name, BucketProps);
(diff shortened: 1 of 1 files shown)


reiddraper:
I'm not finding which code path will return the list of `[{atom(), atom()}]` from `set_bucket/2`. It looks like all of the `StoreFun`s return `ok`. What am I missing?

seancribbs:
@reiddraper `riak_core_bucket_props:validate/4` is where that comes from.

reiddraper:
Right, but we further process that return-value:

``` erlang
case riak_core_bucket_props:validate(update, Bucket, OldBucket, BucketProps0) of
    {ok, BucketProps} ->
        NewBucket = merge_props(BucketProps, OldBucket),
        StoreFun(Bucket, NewBucket);
    {error, Details} ->
        lager:error("Bucket properties validation failed ~p~n", [Details]),
        {error, Details}
end.
```

seancribbs:
Correct, it's the second clause there I'm concerned with.

reiddraper:
But the second clause returns `{error, Something}`, not `[{atom(), atom()}]`.

reiddraper:
Derp, my misreading of this. +1

seancribbs:
@borshop: retry

seancribbs:
@borshop merge

PR #582, opened 2014-05-20
Update lager version and relax version check

cristianberneanu:
Update the version for the lager dependency to 2.0.3 and relax the
version check to permit overriding it from the top-level rebar.config.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -5,7 +5,7 @@
 {eunit_opts, [verbose]}.
 
 {deps, [
-  {lager, "2.0.0", {git, "git://github.com/basho/lager", {tag, "2.0.0"}}},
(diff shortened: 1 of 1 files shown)


reiddraper:
Hey @cristianberneanu, we're using `develop` as our target branch for pull-requests. Can you make a new PR with that branch as the target?

cristianberneanu:
Isn't the develop branch for version 2.0 ? Is there any way I can get those changes into 1.4?

reiddraper:
There's a [1.4 branch](https://github.com/basho/riak_core/tree/1.4) you can target. I don't know whether we intend to cut a 1.4.9, however.

Issue #585, opened 2014-05-22
hashtree tests time out


reiddraper:
In recent builds, I've seen at least two of the `hashtree` tests fail with a timeout.
- [One](http://buildbot.bos1/builders/test-riak_core/builds/983/steps/shell_3/logs/stdio) `hashtree: eqc_test_...*timed out*`
- [Two](http://buildbot.bos1/builders/test-riak_core/builds/984/steps/shell_3/logs/stdio) `hashtree: sha_test_...*timed out*`

This is preventing work from being merged.

cc @seancribbs

reiddraper:
cc @vinoski too. I've been taking a look at the `sha_test_`, and should have a fix soonish.

reiddraper:
#587 should fix issue 2. @kellymclaughlin word is you have a fix for 1.? If we like/agree, we can bypass bors and merge those in manually, and then when leveldb is fixed, we'll be green again.

reiddraper:
@kellymclaughlin has been working on the others in #589

reiddraper:
Closing as #587 and #589 have both been merged.

PR #587, opened 2014-05-22
Fix hashtree:sha_test_ from timing out

reiddraper:
EQC takes a long time, and uses gigabytes of memory to generate 1MB
random binaries, as evidenced by:

```
eqc_gen:sample(eqc_gen:binary(1024 * 1024)).
```

We're still able to write a solid test, by ensuring that we always chunk
the binary into between 1 and 16 chunks. So instead of generating the
chunk size, we now generate the number of chunks, and derive chunk size
from that.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -1129,10 +1129,18 @@ sha_test_() ->
     }.
 
 prop_sha() ->
-    ?FORALL(Size, choose(256, 1024*1024),
-            ?FORALL(Chunk, choose(1, Size),
+    %% NOTE: Generating 1MB (1024 * 1024) size binaries is incredibly slow
+    %% with EQC and was using over 2GB of memory
+    ?FORALL({Size, NumChunks}, {choose(1, 1024), choose(1, 16)},
                     ?FORALL(Bin, binary(Size),
-                            sha(Chunk, Bin) =:= esha(Bin)))).
+                            begin
+                                %% we need at least one chunk,
+                                %% and then we divide the binary size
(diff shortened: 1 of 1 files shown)


reiddraper:
Fixes part of #585

vinoski:
I'm +1 on this change. Originally on the slow build machines this test would take upwards of 90 seconds to execute, would use excessive memory, and would often core dump as a result. With this change the test reliably takes only 4 seconds.

seancribbs:
Nitpick: there doesn't seem to be a reason for nested ?FOR_ALLs there. Won't shrinking be better if they are all in one generation step (perhaps using ?LET)?

reiddraper:
> Nitpick: there doesn't seem to be a reason for nested ?FOR_ALLs there. Won't shrinking be better if they are all in one generation step (perhaps using ?LET)?

Indeed, you're exactly right. The nesting was required before, but in my change, I removed the need for nesting, without realizing it. I'll force-push a change up to here to amend that.

reiddraper:
We'll still need one level of nesting, but that shouldn't be an issue.

reiddraper:
I force-pushed 863baf0e8b4eb7d3c8ce9e77f62608411460020d.

reiddraper:
Derp, I didn't `git add` before ammending, pushed another fix. 602e3aba69231ce636dbd04d6dd0f14ed61b069c.

kellymclaughlin:
+1  602e3ab

reiddraper:
Merging manually since develop is not passing buildbot tests yet.

PR #589, opened 2014-05-23
Extend hashtree eqc test timeouts

kellymclaughlin:
The sha_test_ and eqc_test_ test suites in the hashtree module have
had trouble with slowness and timeouts on slower test machines.  Use
spawn so each test runs in its own process and extend the timeouts to
120 seconds to avoid these spurious failures.


kellymclaughlin:
@borshop merge

kellymclaughlin:
Local result for that test puts indicates it probably needs a higher timeout for slow machines:

```
riak_core_security_tests: security_test_ (user grant/revoke on type/bucket works)...[2.398 s] ok
```

kellymclaughlin:
Extended the timeouts for the `riak_core_security_tests` tests to 60 seconds. There was a lot of whitespace shifting, but the only net change is really adding `{timeout, 60,` to each test case.

kellymclaughlin:
@borshop: retry  fe640a3

reiddraper:
I opened #590 about the particular failure we're seeing now. I vote for merging this in manually for now.

kellymclaughlin:
Agreed, merging now

reiddraper:
This just came back :(. [Here](http://buildbot.bos1/builders/test-riak_core/builds/1043/steps/shell_3/logs/stdio).

reiddraper:
That build above is from #596.

PR #588, opened 2014-05-23
Error handling in riak_core_coverage_fsm:init

jburwell:
Adds error handling the riak_core_coverage_fsm:init to stop the FSM when the Mod:init function call returns {stop, _}.

PR #594, opened 2014-05-27
Error handling in riak_core_coverage_fsm:init

jburwell:
Adds error handling the riak_core_coverage_fsm:init to stop the FSM when the Mod:init function call returns {stop, _}.  This change supports the for https://github.com/basho/riak_kv/issues/875 (PB List Keys to missing bucket type causes crash, hangs client).

This PR replaces https://github.com/basho/riak_core/pull/588 to squash commits.  Due to a mistake rebasing to squash, I simply created a new branch with the few relevant changes.

### DIFF
--- a/src/riak_core_coverage_fsm.erl
+++ b/src/riak_core_coverage_fsm.erl
@@ -158,25 +158,7 @@ init([Mod,
       From={_, ReqId, _},
       RequestArgs]) ->
     Exports = Mod:module_info(exports),
-    {Request, VNodeSelector, NVal, PrimaryVNodeCoverage,
-     NodeCheckService, VNodeMaster, Timeout, ModState} =
-        Mod:init(From, RequestArgs),
-    maybe_start_timeout_timer(Timeout),
-    PlanFun = plan_callback(Mod, Exports),
-    ProcessFun = process_results_callback(Mod, Exports),
-    StateData = #state{mod=Mod,
-                       mod_state=ModState,
-                       node_check_service=NodeCheckService,
-                       vnode_selector=VNodeSelector,
-                       n_val=NVal,
-                       pvc = PrimaryVNodeCoverage,
-                       request=Request,
-                       req_id=ReqId,
-                       timeout=infinity,
-                       vnode_master=VNodeMaster,
-                       plan_fun = PlanFun,
-                       process_fun = ProcessFun},
-    {ok, initialize, StateData, 0};
+    maybe_transition_to_initialize(Mod, Exports, ReqId, Mod:init(From, RequestArgs));
 init({test, Args, StateProps}) ->
     %% Call normal init
     {ok, initialize, StateData, 0} = init(Args),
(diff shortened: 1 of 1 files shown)


reiddraper:
@jburwell can you comment on [@seancribb's comment](https://github.com/basho/riak_kv/pull/955#issuecomment-44068962) about doing this error handling before the FSM is started?

jburwell:
@seancribbs @reiddraper When an FSM transitions to its initial state (starts), it means that the it is in a consistent state to operate.  In some cases (e.g. FSMs working with bucket types), the state provided at initialization maybe invalid -- preventing the FSM from operating properly.  In these scenarios (e.g. providing an undefined bucket type to the riak_kv_keys_fsm), the FSM should abort before starting.  This approach is supported by OTP (see notes regarding {error, Reason} from init on http://www.erlang.org/doc/man/gen_fsm.html).  It DRYs code by consolidating error checking, as well as, ensuring that the FSM does not lie about its state. 

/cc @kellymclaughlin

reiddraper:
@jburwell let me try and rephrase what I think @seancribbs is saying, and Sean can correct me if I'm wrong. The suggestion is to not even start the FSM is the bucket is invalid/doesn't-exist. That way, the FSM is only ever started with valid parameters. This will isolate the change to a smaller set of code, and the FSM can then assume it is only ever started with valid bucket types. Sean can correct me, but I believe this would just mean a small code change in the riak_client.

seancribbs:
Yes, that is what I was trying to say, not so elegantly. We've already made adjustments in the webmachine resources and the PB services (for other endpoints) to check for existence of the bucket type, that's all that's needed to resolve the original issue. If those checks can be consolidated/deduplicated, that'd be great, but isn't strictly necessary this late in the game.

kellymclaughlin:
I do not see that advantage of doing this check in the client versus having each fsm perform the check and just return an error. I suppose having the check in `riak_client` is fine, but it requires checking in at least as many places as checking in the fsms and requires the client module to be aware of details that seem like it shouldn't have to be concerned with. 

The point about it being late in the release cycle is a good one though. I propose we make the change to check for the bucket type in `riak_client` for this release to keep the change footprint as small as possible, but that we consider revisiting the approach after 2.0.

reiddraper:
> The point about it being late in the release cycle is a good one though. I propose we make the change to check for the bucket type in riak_client for this release to keep the change footprint as small as possible, but that we consider revisiting the approach after 2.0.

Seems fair to me.

russelldb:
Was that change made to `riak-client` as suggested? If so can this be either closed or at least milestone reassigned to a later milestone?

jburwell:
Per [riak kv ticket 965](https://github.com/basho/riak_kv/issues/965), moved to 2.1 when we have more time to consider the impact of the change and thoroughly test any side effects.

The fix implemented for 2.0 is captured in [riak kv ticket 875](https://github.com/basho/riak_kv/issues/875).

Issue #590, opened 2014-05-23
bucket_fixup_test load_test occasional failure


reiddraper:
See [riak_core build 1001](http://buildbot.bos1/builders/test-riak_core/builds/1001/steps/shell_3/logs/stdio). Relevant lines here:

```
module 'bucket_fixup_test'
  bucket_fixup_test: fixup_test_...[0.020 s] ok
  bucket_fixup_test: fixup_test_...[0.028 s] ok
  bucket_fixup_test: fixup_test_...[0.027 s] ok
  bucket_fixup_test: fixup_test_...[0.027 s] ok
  bucket_fixup_test: fixup_test_...[0.015 s] ok
  bucket_fixup_test: fixup_test_...[0.203 s] ok
  bucket_fixup_test:92: load_test_...*failed*
in function gen_server:call/3 (gen_server.erl, line 188)
in call from riak_core_bucket:set_bucket_in_ring/2 (src/riak_core_bucket.erl, line 90)
in call from bucket_fixup_test:'-load_test_/0-fun-4-'/0 (test/bucket_fixup_test.erl, line 94)
**exit:{noproc,{gen_server,call,
                    [riak_core_ring_manager,
                     {ring_trans,#Fun<riak_core_bucket.3.106885211>,undefined},
                     infinity]}}
```

This test appears to pass/fail non-deterministically.

reiddraper:
Closing now that #592 has been merged. We'll re-open if this comes back.

PR #598, opened 2014-05-30
Make riak_core_util:safe_rpc catch exit correctly

bowrocker:
The function was attempting to catch exit exception with 'EXIT':{noproc, ...}; should be exit:{noproc, ...}

http://erlang.org/doc/reference_manual/errors.html

Related riak_test PR:

https://github.com/basho/riak_test/pull/624

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -401,7 +401,7 @@ safe_rpc(Node, Module, Function, Args) ->
         Result ->
             Result
     catch
-        'EXIT':{noproc, _NoProcDetails} ->
+        exit:{noproc, _NoProcDetails} ->
             {badrpc, rpc_process_down}
     end.
 


reiddraper:
Thanks for catching this, I've assigned the 2.0-RC milestone to both of these issues.

reiddraper:
cc @lordnull

lordnull:
:+1: as the riak_test proves it's correctness.

Of course, bors isn't happy due to a spurious eunit failure (sha_test timeout).

lordnull:
@borshop merge

lordnull:
:+1:  b01f400

lordnull:
@borshop merge

reiddraper:
The failure we're seeing here is #590.

cmeiklejohn:
Can we manually merge, @reiddraper?

reiddraper:
Sure, and I appreciate that we talked about it in the comments. I just want there to at least be an explanation in the comments every time we decide to manually merge.

In this case, the explanation is that another test is blocking us, and we'd like to get a better handle on the outstanding issues, so having this closed will help us gain better visibility into what's remaining.

Issue #593, opened 2014-05-27
hashtree prop_correct cleanup test failure


jrwest:
see http://buildbot.bos1/builders/test-riak_core/builds/1005

```
module 'hashtree'
  hashtree: snapshot_test...[0.188 s] ok
  hashtree: delta_test...[0.011 s] ok
  hashtree: sha_test_...[8.136 s] ok
  hashtree: eqc_test_...*failed*
in function hashtree:'-eqc_test_/0-fun-0-'/0 (src/hashtree.erl, line 1150)
**error:{assertion_failed,[{module,hashtree},
                   {line,1150},
                   {expression,"eqc : quickcheck ( eqc : testing_time ( 4 , prop_correct ( ) ) )"},
                   {expected,true},
                   {value,false}]}
  output:<<"Failed! Reason: 
{'EXIT',
    {{badmatch,
         {error,
             {error_db_destroy,
                 "IO error: lock /tmp/anti/level/198388418127771313527061732379278910321/LOCK: already held by process"}}},
     [{hashtree,destroy,1,[{file,"src/hashtree.erl"},{line,266}]},
      {hashtree,'-prop_correct/0-fun-4-',2,
          [{file,"src/hashtree.erl"},{line,1230}]}]}}
After 1 tests.
[{<<"2">>,<<180,46,7,151,8,82,248,159>>},
 {<<"3">>,<<48,134,157,69,146,192,46,106>>},
 {<<"1">>,<<31,120,226,110,149,197,32,218>>}]
{1,1,0}
">>

  [done in 58.690 s]
```

reiddraper:
@jtuple any idea what may be going on here? Is the same leveldb instance being opened before the previous one has finished closing? Or something to that effect?

reiddraper:
Closing for now, since we're hopeful that #600 has fixed the issue. If this crops up again, we can re-open this ticket.

PR #600, opened 2014-06-03
attempt to isolate hashtree tests more by using a reference

jrwest:
#593

one idea -- maybe it won't work. `hashtree` tests still pass and I am still unable to repro the failure.

cc/ @jtuple @reiddraper


reiddraper:
@borshop merge

PR #602, opened 2014-06-11
Fix riak_core_util:pmap/2 infinite stall

jtuple:
This commit changes `riak_core_util:pmap/2` to use `spawn_link` rather than `spawn` to create the asynchronous processes. Without this change, `pmap/2` can stall forever if any of these processes crashes -- eg. by using a map function that generates an error.

This commit also fixes the function -spec as well as optimizes the final stage of the pmap by using a list comprehension rather than `lists:unzip` (this is faster and generates less garbage).

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -300,19 +300,22 @@ ensure_started(App) ->
 
 %% @doc Invoke function `F' over each element of list `L' in parallel,
 %%      returning the results in the same order as the input list.
--spec pmap(function(), [node()]) -> [any()].
+-spec pmap(F, L1) -> L2 when
+      F :: fun((A) -> B),
+      L1 :: [A],
+      L2 :: [B].
 pmap(F, L) ->
     Parent = self(),
     lists:foldl(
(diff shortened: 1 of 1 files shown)


engelsanchez:
I'm taking this review.

engelsanchez:
:+1: e8d604c

Added a unit test to verify the stall doesn't happen on a crashed worker. This test would time out on develop and pass with this branch.  One thing: if the calling process is trapping exits, the pmap can stall anyway.  If we want to prevent these things, we may have to re-implement both versions of pmap to spawn a separate process to make sure trap_exit is false.  This is still an improvement, so we can do that later if we wish to do it at all.

Another minor thing to watch out for: this pmap is mostly used in the family of multi-rpc helpers in the module.  There was some effort made to make RPC calls robust to rex being down.  Assuming we are catching the all the possible exceptions in the safe wrapper, things should be good. If not, this change will propagate a crash, which is better than stalling forever anyway.

jtuple:
@borshop merge

PR #609, opened 2014-06-30
Use a proxy process in claimant when joining/removing SC nodes

andrewjstone:
We use a proxy process in the claimant to handle errors when calling `riak_ensemble_manager:{join/remove}`. These errors can result from the ensemble manager crashing or strong consistency not being enabled on some nodes during a rolling upgrade. This change also prevents late responses from getting sent to the claimant in the case that it already got a timeout for the given operation.
### TESTING

Perform the following steps with and without the bugfix to see crashes
in the log on dev2 without the fix and no crashes with the fix. The
crashes should look like this:

```
12:46:47.992 [error] Supervisor riak_core_sup had child
riak_core_claimant started with riak_core_claimant:start_link() at
<0.12836.0> exit with reason no such process or port in call to
gen_server:call({riak_ensemble_manager,'dev1@127.0.0.1'},
{join,'dev2@127.0.0.1'}, infinity) in context child_terminated
12:46:57.993 [error] gen_server riak_core_claimant terminated with
reason: no such process or port in call to
gen_server:call({riak_ensemble_manager,'dev1@127.0.0.1'},
{join,'dev2@127.0.0.1'}, infinity)
```
- Build a devrel
- Config one node (dev2) to turn on strong consistency
  - Edit `/dev/dev2/etc/riak.conf`
  - Set `strong_consistency = on`
  - Start 3 nodes
    - `/dev/dev{1,2,3}/bin/riak start`
  - Join nodes 1 and 3 to node 2
    - `dev/dev3/bin/riak-admin cluster join dev2@127.0.0.1`
    - `dev/dev1/bin/riak-admin cluster join dev2@127.0.0.1`
    - `dev/dev2/bin/riak-admin cluster plan`
    - `dev/dev2/bin/riak-admin cluster commit`
  - Wait for the cluster to become ready by checking the `riak-admin status`
  - Check the ensemble status - `dev/dev2/bin/riak-admin ensemble-status`
  - Notice that the root ensemble only has 1 node and the others don't have leaders 

```
============================== Consensus System
===============================
Enabled:     true
Active:      true
Ring Ready:  true
Trust:       medium (AAE syncing required)
AAE enabled: true

==================================
Ensembles
==================================
Ensemble     Quorum   Nodes      Leader
-------------------------------------------------------------------------------
root       1 / 1     1 / 1      dev2@127.0.0.1

```
- On nodes 1 and 3, turn on strong consistency and restart the nodes.
- Check status after about a minute to verify 3 root nodes. It will take a
  little while longer, but other ensembles will come online
  eventually.

```
============================== Consensus System
===============================
Enabled:     true
Active:      true
Ring Ready:  true
Trust:       medium (AAE syncing required)
AAE enabled: true

================================== Ensembles ==================================
 Ensemble     Quorum        Nodes      Leader
-------------------------------------------------------------------------------
     root      3 / 3         3 / 3      dev2@127.0.0.1
     2         0 / 3         2 / 2      --
     3         3 / 3         3 / 3      dev3@127.0.0.1
     4         0 / 3         2 / 2      --
```

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -623,6 +623,7 @@ proxy_spawn(Fun) ->
     Pid ! {proxy, MRef},
     receive
         {proxy_reply, MRef, Result} ->
+            demonitor(MRef, [flush]),
             Result;
         {'DOWN', MRef, _, _, Reason} ->
             {error, Reason}
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -853,5 +853,21 @@ make_fold_req_test_() ->
      ]
     }.
 
+proxy_spawn_test() ->
+    A = proxy_spawn(fun() -> a end),
+    ?assertEqual(a, A),
+    B = proxy_spawn(fun() -> exit(killer_fun) end),
+    ?assertEqual({error, killer_fun}, B),
+
+    %% Ensure no errant 'DOWN' messages
+    receive
+        {'DOWN', _, _, _, _}=Msg ->
+            throw({error, {badmsg, Msg}});
+        _ ->
+            ok
+    after 1000 ->
+        ok
+    end.
+
 -endif.
 
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -60,7 +60,9 @@
          make_fold_req/1,
          make_fold_req/2,
          make_fold_req/4,
-         make_newest_fold_req/1
+         make_newest_fold_req/1,
+         proxy_spawn/1,
+         proxy/2
         ]).
 
 -include("riak_core_vnode.hrl").
@@ -86,7 +88,7 @@
 %%      number of seconds from year 0 to now, universal time, in
 %%      the gregorian calendar.
 
-moment() -> 
+moment() ->
     {Mega, Sec, _Micro} = os:timestamp(),
     (Mega * 1000000) + Sec + ?SEC_TO_EPOCH.
 
@@ -238,8 +240,8 @@ mkclientid(RemoteNode) ->
 chash_key({Bucket,_Key}=BKey) ->
     BucketProps = riak_core_bucket:get_bucket(Bucket),
     chash_key(BKey, BucketProps).
-    
-%% @spec chash_key(BKey :: riak_object:bkey(), [{atom(), any()}]) -> 
+
+%% @spec chash_key(BKey :: riak_object:bkey(), [{atom(), any()}]) ->
 %%          chash:index()
 %% @doc Create a binary used for determining replica placement.
 chash_key({Bucket,Key}, BucketProps) ->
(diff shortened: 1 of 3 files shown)


andrewjstone:
Fixes basho/riak#559

lordnull:
Without patch, crash log is full and there's a single ensemble on dev2, no leaders for others.

With patch, crash log is empty, no ensembles created on dev2 until 1 and 3 have SC enabled.

Code looks clean and is pretty easy to grok.

Not sure if we can create an automated test for this.

andrewjstone:
The simplest test I can think of would be to use an intercept to make the proxy catch any errors instead of actually proxying, with `strong_consistency = on` on only one node. Then remove the intercept, change the config to enabling SC on all nodes and ensure operations complete successfully. Honestly though I'm not sure it's worth it this fix.

lordnull:
:+1: 68e25d4

jtuple:
Can we please not have a module called `proxy` in `riak_core`? Perhaps `riak_core_proxy`? Or just move the proxy functions into `riak_core_util`?

Yes, I'm aware I named things `proxy.erl` in my gist. But, that was gist-quality code ;)

andrewjstone:
@jtuple I meant to ask that actually but somehow magically did not. I support putting it in riak_core_util since there are already other files with the name 'proxy' in them. I'll make that change and push.

andrewjstone:
@jtuple @lordnull rebased to use riak_core_util:proxy_spawn/1 in the claimant.

lordnull:
:+1: 8cefb59

It took some time for the ensembles to come back, but they do come.

andrewjstone:
:+1: 016b22a

Needed to update dialyzer.ignore-warnings line numbers

src/riak_core_util.erl
```diff
@@ -611,6 +613,22 @@ make_newest_fold_req(#riak_core_fold_req_v1{foldfun=FoldFun, acc0=Acc0}) ->
...
+%%      calls.
+proxy_spawn(Fun) ->
+    %% Note: using spawn_monitor does not trigger selective receive
+    %%       optimization, but spawn + monitor does. Silly Erlang.
+    Pid = spawn(?MODULE, proxy, [self(), Fun]),
+    MRef = monitor(process, Pid),
+    Pid ! {proxy, MRef},
+    receive
+        {proxy_reply, MRef, Result} ->
+            Result;
```
jtuple, line 627: Just remembered we need to `demonitor` here. Above this line, add a `demonitor(MRef, [flush])`; otherwise, the `DOWN` message will arrive after we return from this function, potentially confusing/crashing the calling process.

lordnull:
Yeah, demonitor w/ flush is a good plan. The eunit test passes. The only case it doesn't cover is if the parent process (the one making the request) goes down.

Rest of the manual testing still goes as before.

jonmeredith:
@andrewjstone @lordnull what's the plan for this PR?

andrewjstone:
I think it's good to go. @lordnull want to give it the :+1: ?

lordnull:
:+1: c8ba479

andrewjstone:
@borshop merge

PR #612, opened 2014-07-08
Update tools.mk to v0.5.5

reiddraper:


reiddraper:
See basho/riak_kv#999, which was closed to make a minor fix. And basho/riak_kv#998, where @engelsanchez originally introduced the tools.mk change.

russelldb:
+1 c206922

Verified that the `tools.mk` is the same as in `basho/tools.mk` tagged `0.5.5`. Given the review and testing that got these changes merged there, that seems to be enough to pass this.

reiddraper:
@borshop merge

PR #617, opened 2014-08-01
Alert riak_kv_ensembles server of possible change

andrewjstone:
The riak_kv_ensembles gen_server determines which ensembles need to be
created based on the current ring and n_vals corresponding to both
existing buckets and active bucket types. It only runs this check on
ring changes though. We need to alert it to check again in case a
strongly consistent bucket type was activated.

Part of a fix for basho/riak_kv#1007
Requires https://github.com/basho/riak_kv/pull/1008

Corresponding riak_test coming soon


andrewjstone:
Ahh shit. Closing this as it breaks the abstraction between riak_core and riak_kv. I guess I'll have to call `riak_core_ring_manager:force_update()` instead. I was trying to be more lightweight. Oh well.

PR #618, opened 2014-08-01
Refresh the ring after activating a bucket type

andrewjstone:
This is a somewhat heavy hammer, but it is expected that bucket types
are created somewhat infrequently. It is required to notify other
applications on top of riak core that may need this changing
information. In particular it's required by riak_kv to bootstrap riak
ensembles.

Supercedes #617
Part of a fix for basho/riak_kv#1007
Requires basho/riak_kv#1008

### DIFF
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -734,7 +734,9 @@ maybe_activate_type(_BucketType, active, _Props) ->
     ok;
 maybe_activate_type(BucketType, ready, Props) ->
     ActiveProps = lists:keystore(active, 1, Props, {active, true}),
-    riak_core_metadata:put(?BUCKET_TYPE_PREFIX, BucketType, ActiveProps).
+    riak_core_metadata:put(?BUCKET_TYPE_PREFIX, BucketType, ActiveProps),
+    riak_core_ring_manager:force_update(),
+    ok.
 
 %% @private
 type_active(Props) ->


jrwest:
`force_update` and cluster metadata are entirely independent? The `claimant` should be using `prefix_hash` or something to that effect if it needs to determine when bucket types change.

EDIT: unless im entirely misunderstanding what this does (totally possible)

andrewjstone:
@jrwest This is just there to alert riak_kv_ensembles gen_server next time it polls that the ringId has changed. https://github.com/basho/riak_core/pull/617 was the first attempt that I immediately rejected. Regardless I want to discuss the whole issue with Joe when he's back on Monday, so this isn't going in without some discussion.It's possible I'm doing something in a manner more difficult than needed or incorrectly, since I figured out how to do this by reading the code and testing the results essentially. The corresponding k/v pr has more detail. It's linked in the PR description.

jrwest:
The ring id shouldn't have changed because a bucket type is created. It sounds like this may work because its racing with something and regularly winning?

andrewjstone:
@jrwest You are correct in that the metadata should be checked for changes by the consumer of this info instead of forcing a ring change, which is quite hacky. After talking with @jtuple I've made the following KV change to remove the need for this. If we decide we need the optimization I'll add a check for cluster metadata instead of making this change. Either way, it can be closed.

https://github.com/basho/riak_kv/commit/59883e44c1fe7882e8b6999ddaafcb6f46a109bb

PR #624, opened 2014-09-08
Handoff sender sends sync periodically

kellymclaughlin:
Equivalent to #560, but targetted at the 2.0 branch.
The following description was copied over from #560:

If we can discover the `handoff_receive_timeout value`, then use 1/3rd that interval to send a sync/heartbeat thingie, otherwise 60 seconds (for infinity & undefined).

Purpose: to avoid unhelpful timeouts when the handoff sender is filtering (e.g. repair or ring resizing handoff) and encounters zillions of keys to filter and thus doesn't send anything to the receiver for a while.  A workaround is to set the receiver's timeout to be extremely long, but that's a kludge IMHO.

### DIFF
--- a/src/riak_core_handoff_sender.erl
+++ b/src/riak_core_handoff_sender.erl
@@ -63,6 +63,7 @@
           item_queue_byte_size :: non_neg_integer(),
 
           acksync_threshold    :: non_neg_integer(),
+          acksync_timer        :: timer:tref(),
 
           type                 :: ho_type(),
 
@@ -216,8 +217,10 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
            total_objects=TotalObjects,
            total_bytes=TotalBytes,
            stats=FinalStats,
+           acksync_timer=TRef,
            notsent_acc=NotSentAcc} = AccRecord,
 
+         _ = timer:cancel(TRef),
          case ErrStatus of
              ok ->
                  %% One last sync to make sure the message has been received.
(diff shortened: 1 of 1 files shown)


src/riak_core_handoff_sender.erl
```diff
@@ -281,11 +284,40 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
              gen_fsm:send_event(ParentPid, {handoff_error, Err, Reason})
      end.
 
+start_visit_item_timer() ->
+    Ival = case app_helper:get_env(riak_core, handoff_receive_timeout) of
```
kellymclaughlin, line 288: I think this `case` should just be replaced with a call to `get_handoff_receive_timeout/0`. It uses a default of `?TCP_TIMEOUT` which is defined as 60000.

src/riak_core_handoff_sender.erl
```diff
@@ -281,11 +284,40 @@ start_fold(TargetNode, Module, {Type, Opts}, ParentPid, SslOpts) ->
...
+                   60*1000
+           end,
+    timer:send_interval(Ival, tick_send_sync).
+
+visit_item(K, V, Acc0 = #ho_acc{acksync_threshold = AccSyncThreshold}) ->
+    %% Eventually, a vnode worker proc will be doing this fold, but we don't
+    %% know the pid of that proc ahead of time.  So we have to start the
+    %% timer some time after the fold has started execution on that proc
+    %% ... like now, perhaps.
+    Acc = case get(is_visit_item_timer_set) of
```
kellymclaughlin, line 301: It seems like overkill to have to execute this `case` statement for every item in the fold during handoff. Since `start_visit_item_timer` calls `timer:send_interval`, it seems like we could call `start_visit_item_timer` right before starting the fold and cut this function down to just the `receive` block.

src/riak_core_handoff_sender.erl
kellymclaughlin, line 288: I retract my previous comment about this. I missed the fact that the function `get_handoff_receive_timeout` actually retrieves the value for `handoff_timeout`, not `handoff_receive_timeout`.

src/riak_core_handoff_sender.erl
kellymclaughlin, line 301: Sorry, misunderstood how this flowed at first glance. I retract my statement.

kellymclaughlin:
I modified the `verify_handoff` test to verify the efficacy of this change: https://github.com/basho/riak_test/pull/681

kellymclaughlin:
Once I get a review on my test changes and get that merged in I'm +1 on this change.

andrewjstone:
This needs to be rebased onto 2.0. Otherwise you will end up with the following error:

```
[error] riak_kv_vnode command failed {undef,[{vclock,get_dot,[<<197,82,177,11,84,24,93,163>>,[{<<197,82,177,11,84,24,93,163>>,{1,63578102181}}]],[]},{riak_object,increment_vclock,3,[{file,"src/riak_object.erl"},{line,638}]}
```

kellymclaughlin:
> This needs to be rebased onto 2.0. Otherwise you will end up with the following error:

Ah yes, John must have merged the PR he was reviewing and that caused a conflict in the last few days. 

@andrewjstone, I pushed a rebased branch for merge, but due to the conflict do you mind just glancing over the diff the verify my resolution looks correct?

https://github.com/basho/riak_core/compare/2.0...feature/handoff-sender-sync-tick20-rebased?expand=1

kellymclaughlin:
Just went ahead and created a PR for simplicity: #627

PR #628, opened 2014-09-18
Add breadth-first AAE exchange

jtuple:
This commit ports the breath-first exchange algorithm from the synctree code in riak_ensemble to the hashtree code here in riak_core. This level-by-level exchange is necessary to support streaming and/or pipelined AAE exchange -- an approach that underpins the new AAE-based fullsync replication protocol in Riak Enterprise.

Sibling pull-requests: basho/riak_kv#1023, basho/riak_repl#616

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -124,6 +124,7 @@
          segments/1,
          width/1,
          mem_levels/1]).
+-export([compare2/4]).
 
 -ifdef(TEST).
 -export([local_compare/2]).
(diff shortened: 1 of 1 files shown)


src/hashtree.erl
```diff
@@ -805,6 +806,59 @@ iterate({ok, K, V}, IS=#itr_state{itr=Itr,
...
+    end.
+
+exchange_level(Level, Buckets, Local, Remote, _Opts) ->
+    Remote(start_exchange_level, {Level, Buckets}),
+    lists:flatmap(fun(Bucket) ->
+                          A = Local(get_bucket, {Level, Bucket}),
+                          B = Remote(get_bucket, {Level, Bucket}),
+                          Delta = riak_ensemble_util:orddict_delta(lists:keysort(1, A),
+                                                                   lists:keysort(1, B)),
+                          Diffs = Delta,
```
jonmeredith, line 841: Why have `Diffs`, why not do the same as `exchange_final`?

krestenkrab, line 841: Good observation.  `riak_ensemble_util:orddict_delta/2` returns `[ {'$none', any()} | {any(), '$none'} | {any(), any()} ]`.  Thus, if the local cluster has some missing sub-buckets or segments as compared to the remote cluster; the result of this function (line below) would include some `'$none'`s.  How does the next level in the recursion handle that?

jonmeredith:
Ran coverage tests and confirmed compare2 was covered by EQC.

I did notice that there was a local `orddict_delta` as well as the version called from ensemble.  Seems confusing to keep both around.  Also seems weird that it's in ensemble rather than a utils library.

Not critical for merge, but if left as it is should add a comment saying what the plan for it is.

src/hashtree.erl
jonmeredith, line 841: I just meant there was an unnecessary variable being set, it looks like the `exchange_level/6` usage strips the difference on L842 and knows that it needs to recurse to the next level if there are any differences.

jonmeredith:
I suppose we have lost test coverage for `compare/4` which is still called by `yz_index_hastree`, `riak_core_metadata_exchange_fsm` and `riak_kv_2i_aae`.

src/hashtree.erl
```diff
@@ -805,6 +806,59 @@ iterate({ok, K, V}, IS=#itr_state{itr=Itr,
...
+                          [BK || {BK, _} <- Diffs]
```
krestenkrab, line 842: What I mean is that maybe this line should read

``` erlang
 [BK || {BK, _} <- Diffs, BK =/= '$none']
```

jonmeredith, line 842: BK is the key in the orddict, so I think it will be keyed as Bucket. Here's an example of orddict_delta showing how the key is separate from the `'$none'` so you can see BK will not be set to it.

```
(riak@127.0.0.1)87> riak_ensemble_util:orddict_delta(orddict:from_list([{a,1},{b,2},{c,3}]),orddict:from_list([{b,2},{c,5},{d,4}])).
[{a,{1,'$none'}},{c,{3,5}},{d,{'$none',4}}]
```

jonmeredith:
Code inspection looks good - only issues are minor.  Will complete other PR reviews before giving final approval

jonmeredith:
+1 to merge - but this should be against 2.0 so not telling bors to do it.

Ran with this and sibling branches and got through riak_tests.

jonmeredith:
@jtuple I think this should be closed against develop and just merge the 2.0 branch in.  Can you confirm?

PR #629, opened 2014-09-22
Add breadth-first AAE exchange

jtuple:
This commit ports the breath-first exchange algorithm from the
synctree code in riak_ensemble to the hashtree code here in riak_core.
This level-by-level exchange is necessary to support streaming and/or
pipelined AAE exchange -- an approach that will soon underpin the new
AAE-based fullsync replication protocol in Riak.

This is the 2.0 rebased version of #628

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -124,6 +124,7 @@
          segments/1,
          width/1,
          mem_levels/1]).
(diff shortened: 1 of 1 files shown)


jtuple:
@borshop merge

Issue #564, opened 2014-03-25
Handoff exceeding the transfer limit should not throw an error


Vagabond:
Right now, when riak tries to handoff more than the limit allows, you get scary messages like this:

```
2014-03-25 16:49:58.988 [error] <0.142.0> Supervisor riak_core_handoff_sender_sup had child 
riak_core_handoff_sender started with {riak_core_handoff_sender,start_link,undefined} at 
<0.18847.7518> exit with reason max_concurrency in context child_terminated
```

Instead of exiting with max_concurrency, we should exit normal (possibly logging something to info/notice about exceeding the transfer limit). The supervisor has the children flagged as temporary anyway, so this will not impact retrying handoff later.

bearcage:
Started on this: lacks tests still, but it unpacks a the meanings behind {shutdown, max_concurrency}.

https://github.com/basho/riak_core/compare/bugfix/atb/remove_max_concurrency_log_spam?expand=1

PR #630, opened 2014-09-29
Refactored handoff sender max_concurrency to reduce overloaded meanings.

bearcage:
Addresses #564 [JIRA](https://bashoeng.atlassian.net/browse/RIAK-1189) — still needs tests, just opening this to get review started.
- Separated handoffs terminated for concurrency limit violations from those
  terminated for real errors -- concurrency violations no longer cause
  abnormal process exits.
  - Split up the three sources of max_concurrency exits in handoff_sender:
    1. vnode lock max_concurrency
    2. remote node hitting its concurrency limit
    3. concurrency limit lowered after handoffs already started
  - Added a mechanism for marking a transfer as "rescheduled" -- considered
    failed on completion.
  
  This is used to notify the handoff manager that a given handoff, while
  exiting normal, should not signal xfer_complete to the corresponding
  vnode. NB this logic is basically working around the relatively tight
  coupling between handoff_manager and vnode_manager; while there's some
  groundwork laid for decoupling them in the form of the optional
  `handoff_started/2', the handoff manager still special-cases a bit
  just for vnode_manager.

### DIFF
--- a/src/riak_core_handoff_manager.erl
+++ b/src/riak_core_handoff_manager.erl
@@ -44,7 +44,8 @@
          set_concurrency/1,
          get_concurrency/0,
          set_recv_data/2,
-         kill_handoffs/0
+         kill_handoffs/0,
+         reschedule/2
         ]).
 
 -include("riak_core_handoff.hrl").
@@ -122,6 +123,13 @@ status(Filter) ->
 -spec status_update(mod_src_tgt(), ho_stats()) -> ok.
 status_update(ModSrcTgt, Stats) ->
     gen_server:cast(?MODULE, {status_update, ModSrcTgt, Stats}).
+
+%% @doc Permit the handoff identified by `ModSrcTgt' to be rescheduled
+%%      (assume it was unsuccessful on completion), and store the reason
+%%      why retrying is necessary.
+-spec reschedule(mod_src_tgt(), any()) -> ok.
+reschedule(ModSrcTgt, Reason) ->
+    gen_server:call(?MODULE, {reschedule, ModSrcTgt, Reason}, infinity).
 
 set_concurrency(Limit) ->
     gen_server:call(?MODULE,{set_concurrency,Limit}, infinity).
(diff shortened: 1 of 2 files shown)


macintux:
+R

src/riak_core_handoff_manager.erl
```diff
@@ -203,21 +211,34 @@ handle_call({set_concurrency,Limit},_From,State=#state{handoffs=HS}) ->
...
 
 handle_call(get_concurrency, _From, State) ->
     Concurrency = get_concurrency_limit(),
-    {reply, Concurrency, State}.
+    {reply, Concurrency, State};
+
+handle_call({reschedule, ModSrcTgt, Reason}, _From, State=#state{handoffs=HS}) ->
+    case lists:keyfind(ModSrcTgt, #handoff_status.mod_src_tgt, HS) of
+        false ->
+            lager:error("Tried mark a nonexistent handoff as rescheduled: ~p~n", [ModSrcTgt]),
```
macintux: Grammar oops, nbd

src/riak_core_handoff_manager.erl
```diff
@@ -123,6 +124,13 @@ status(Filter) ->
 status_update(ModSrcTgt, Stats) ->
     gen_server:cast(?MODULE, {status_update, ModSrcTgt, Stats}).
 
+%% @doc Permit the handoff identified by `ModSrcTgt' to be rescheduled
+%%      (assume it was unsuccessful on completion), and store the reason
+%%      why retrying is necessary.
+-spec reschedule(mod_src_tgt(), any()) -> ok.
+reschedule(ModSrcTgt, Reason) ->
+    gen_server:call(?MODULE, {reschedule, ModSrcTgt, Reason}, infinity).
```
macintux, line 132: Any particular reason this is a call instead of a cast? Doesn't seem like a caller is likely to care (and in fact the places where `reschedule` is invoked disregard the return value).

src/riak_core_handoff_manager.erl
```diff
@@ -295,13 +311,28 @@ handle_info({'DOWN', Ref, process, _Pid, Reason}, State=#state{handoffs=HS}) ->
...
                         _ ->
-                            %% Use proplist instead so it's more
-                            %% flexible in future, or does
-                            %% capabilities nullify that?
-                            Msg = {M, S, I},
-                            riak_core_vnode_manager:xfer_complete(Origin, Msg)
-                    end,
-                    ok
+                            case Status of
+                                {rescheduled, module_max_concurrency} ->
```
macintux: I struggle remembering what `module_max_concurrency` is; is there any reason `sender_max_concurrency` wouldn't work?

bearcage: module_max_concurrency is a kind of odd one — the only way that gets thrown is when the module owning the partition under handoff errors with max concurrency.

There's an optional handoff_started interface, that only riak_kv implements right now, which can throw max_concurrency. It seems like a sort of ill-defined interface right now, but I didn't want to get into more clearly specifying it in this PR since that'd further widen the surface area, but if you think it's worth doing now I'm happy to.

bearcage: After talking about this on the call I'm inclined to just ignore this case and not catch it — if/when we get back to background_mgr we can clean it up, since this is sort of out of scope.

src/riak_core_handoff_manager.erl
bearcage, line 132: Going to postpone addressing this one until after the handoff_manager tests are at least partially written

Basho-JIRA:
Reassigning from Alex Berghage (no longer with Basho). 

**_[posted via JIRA by Derek Somogyi]**_

PR #632, opened 2014-10-01
Add support for 0-based partition IDs

macintux:
Initial goal is to allow user commands to specify partition IDs as short integers `[0, ring_size)` instead of hash values. May consolidate other features such as Yokozuna's logical partitions later.


src/riak_core_ring_util.erl
```diff
@@ -56,3 +68,94 @@ check_ring(Ring, Nval) ->
...
+                                 riak_core_ring:riak_core_ring() |
+                                 pos_integer()) ->
+                                        boolean().
+%% @doc For user-facing tools, indicate whether a specified hash value
+%% is a valid "boundary" value (first hash in some partition)
+hash_is_partition_boundary(CHashKey, Ring) when is_binary(CHashKey), is_tuple(Ring)->
+    <<CHashInt:160/integer>> = CHashKey,
+    hash_is_partition_boundary(CHashInt, Ring);
+hash_is_partition_boundary(CHashInt, Ring) when is_tuple(Ring) ->
+    PartitionCount = riak_core_ring:num_partitions(Ring),
```
JeetKunDoug: Trying to figure out how this works - it appears that we should really be passing chash:ring_increment(PartitionCount) to the base case, not ring size. If the hash is divisible by ring_increment, it is a boundary. But there could (and are) many values that are divisible by ring_size, most of which aren't boundaries.

macintux: Yep, I screwed that up pretty good. Made a similar mistake elsewhere.

Great catch, thanks, will extend testing.

seancribbs:
My drive-by-review: Seems like this is a good candidate for getting your feet wet in EQC.

macintux:
Yep, I'd considered that. Unfortunately the functionality is almost pure math, so the same description I'd use for testing is effectively the code itself, so I was afraid I'd just repeat the logic failures when writing the tests.

Clearly, though, this latest bug is a stupid one, so maybe I'd not be quite so stupid twice...wait, did make the same mistake twice, maybe I'd not quite so stupid 3 times...

macintux:
@JeetKunDoug Fixed. Will tackle EQC before considering this complete, however

bearcage:
Reading

macintux:
Reasonably happy with the QuickCheck coverage at this point, although I'm going to be chatting with Reid tomorrow about improving them stylistically.

macintux:
On further reflection, don't think I'm adequately testing via eqc to make sure `hash_is_partition_boundary/2` doesn't have false positives.

**Update**: addressed in latest commit

src/riak_core_ring_util.erl
```diff
@@ -56,3 +73,229 @@ check_ring(Ring, Nval) ->
                                 ordsets:add_element(PL, Acc)
                         end
                 end, [], Preflists).
+
+-spec hash_to_partition_id(chash:index() | chash:index_as_int(),
+                riak_core_ring:riak_core_ring() | pos_integer()) ->
+                       partition_id().
+%% @doc Map a key hash (as binary or integer) to a partition ID [0, ring_size)
+hash_to_partition_id(CHashKey, Ring) when is_binary(CHashKey), is_tuple(Ring)->
```
jrwest: a few comments about this function:
1. I think it would be better if it lived in `riak_core_ring`. Then it could properly check that the only tuple its getting is the `riak_core_ring` record. As this function stands calls like `hash_to_partition_id(0, {something, that, is, not, a, ring})` will crash despite the function guards. 
2. Instead of throwing a custom function, we should just rely on erlang's support for partial functions and the error thrown there
3. I initially assumed we were only building this function w/ the assuming that the index passed in was a vnode identifier (a specific subset of hashes). Having it support for is probably better, since there is no way to nicely enforce what integers can be passed it, just curious why you decided to extend the scope?

src/riak_core_ring_util.erl
```diff
@@ -56,3 +73,229 @@ check_ring(Ring, Nval) ->
...
+    CHashInt div chash:ring_increment(RingSize).
+
+-spec partition_id_to_hash(partition_id(),
+                           riak_core_ring:riak_core_ring() | pos_integer()) ->
+                                  chash:index_as_int().
+%% @doc Identify the first key hash (integer form) in a partition ID [0, ring_size)
+partition_id_to_hash(Id, Ring) when is_tuple(Ring) ->
+    partition_id_to_hash(Id, riak_core_ring:num_partitions(Ring));
+partition_id_to_hash(Id, RingSize) when Id < 0; Id >= RingSize ->
+    throw(invalid_partition_id);
```
jrwest: regarding the `throw` here I have the same comment as `partition_id_to_hash`

src/riak_core_ring_util.erl
jrwest: An alternative to number one would be to not support passing in the ring here. Only the ring size. Callers would then have to call `riak_core_ring:num_partitions` themselves....

src/riak_core_ring_util.erl
jrwest: re: point 3 in my comment above, on what hashes are passed to `hash_to_partition_id`, I suppose one downside of taking any hash is that `partition_id_to_hash` is no longer an inverse _shrug_.

src/riak_core_ring_util.erl
```diff
@@ -56,3 +73,229 @@ check_ring(Ring, Nval) ->
...
+partition_id_to_hash(Id, RingSize) ->
+    Id * chash:ring_increment(RingSize).
+
+
+-spec hash_is_partition_boundary(chash:index() | chash:index_as_int(),
+                                 riak_core_ring:riak_core_ring() |
+                                 pos_integer()) ->
+                                        boolean().
+%% @doc For user-facing tools, indicate whether a specified hash value
```
jrwest, line 97: What tools were you picturing using this? Is the idea that we are curious if a hash is a vnode a valid vnode identifier in the ring?

src/riak_core_ring_util.erl
macintux: 1. I'm not terribly averse to ditching support for passing the ring. Passing the ring seems to aid with abstraction, but it's not going to be used very often either way.
2. The downside, of course, is that whatever error message you receive is going to be far less explicit and harder to troubleshoot. Is there a reason you'd rather go the implicit route?
3. My original code checked to make sure the hashes were indeed partition boundaries, but I decided that it made more sense to make this a general hash -> partition mapper because it made it more useful and removed a throw. The `hash_is_partition_boundary/2` function was introduced so that code that needed to do this mapping could make an intelligent decision about what to do with hash values that aren't at the top of a partition.

src/riak_core_ring_util.erl
macintux, line 97: I see this being useful for CLI tools that want to take partition hash values and convert them to IDs for whatever reason. The calling code gets to decide what to do if the hash doesn't map properly rather than me throwing an error in `hash_to_partition_id/2`

src/riak_core_ring_util.erl
jrwest: 1. I'm fine if we want to keep passing in the ring, but then the function should live in `riak_core_ring` imo, so we can do the proper guards on the function.
2. Partial functions are standard erlang imo. At some point, even when you don't have types, we need to consider the input sanitized. Sprinkling checks through every layer of the codebase is a peril of dynamic langs imo. The first place we plan to use this function we know we will have well formed partition identifiers (from the vnode themselves)
3. That may be helpful but again I think is maybe a little too defensive programming. Like I mentioned in my reply to 2, the first place we will probably get these identifiers is from the vnode itself (or the ring), in which case we know they are valid and lie on a "boundary".

src/riak_core_ring_util.erl
jrwest, line 97: I think it may be useful but not sure we have a use-case for it to exist yet, which is what I was try to dig into. See my response to 2 & 3 in the previous comment.

macintux:
@jrwest Now ready for another look per our discussion. Undecided about the cost:benefit ratio for the tagged tuples, so that may not be my best idea.

src/riak_core_ring_util.erl
```diff
@@ -56,3 +73,200 @@ check_ring(Ring, Nval) ->
...
+%% Partition IDs should map to hash values which map back to the same partition IDs
+prop_reverse() ->
+    ?FORALL(RingPower, choose(2, ?RINGSIZEEXPMAX),
+            ?FORALL(PartitionId, choose(0, ?RINGSIZE(RingPower) - 1),
+                    riak_core_ring_util:hash_to_partition_id(
+                      riak_core_ring_util:partition_id_to_hash(PartitionId,
+                                                               ?RINGSIZE(RingPower)),
+                      ?RINGSIZE(RingPower)) =:= {partition_id, PartitionId})).
+
+%% Hash values map to discrete partitions (this test is of dubious value)
```
jrwest: this test is the same as `prop_ids_are_boundaries`. The `HashValue` is projected to the partition id that owns it, and then we execute the same code.

src/riak_core_ring_util.erl
```diff
@@ -56,3 +73,200 @@ check_ring(Ring, Nval) ->
...
+                                    riak_core_ring_util:hash_to_partition_id(GreaterHash,
+                                                                             ?RINGSIZE(RingPower)),
+                                LowerPartition =< GreaterPartition
+                            end
+                           ))).
+
+
+%% Only hash values evenly divisible by partition size respond true to
+%% hash_is_partition_boundary/2.
+boundary_helper(Hash, PartitionSize) when Hash rem PartitionSize =:= 0 ->
```
jrwest: In this case our model is identical to the implementation. On one hand, that is ok -- the model should be ignorant of the implementation. On the other hand it could also be said that we are simply testing that erlang executes the same math in two different code blocks the same. It might be helpful to change the model to generate a ring of `?RINGSIZE(RingPower)` using `chash` and define `boundary_helper` by checking whether `HashValue` is one of the partition identifiers in the `chash`.

src/riak_core_ring_util.erl
```diff
@@ -56,3 +73,200 @@ check_ring(Ring, Nval) ->
...
+    CHashInt rem chash:ring_increment(RingSize) =:= 0.
+
+
+%% ===================================================================
+%% EUnit tests
+%% ===================================================================
+-ifdef(TEST).
+
+%% Partition boundaries are reversable.
+reverse_test() ->
```
jrwest, line 116: these eunits are covered by the eqcs and can be removed?

src/riak_core_ring_util.erl
```diff
@@ -24,7 +24,24 @@
...
+-ifdef(EQC).
+-export([prop_ids_are_boundaries/0, prop_reverse/0, prop_discrete_partitions/0,
+         prop_monotonic/0, prop_only_boundaries/0]).
+-include_lib("eqc/include/eqc.hrl").
+-endif.
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
+-type partition_id() :: {partition_id, non_neg_integer()}. %% Integer represents
+%% a value [0, ring_size)
```
jrwest: style nitpick: move this comment above the type instead of this odd line wrapping?

jrwest:
A rebase/squash would also be good.

src/riak_core_ring_util.erl
macintux: I was attempting to make `edoc` treat that comment properly. It turns out I can move it to the line below the `-type` and `edoc` will still include it in the generated HTML.

src/riak_core_ring_util.erl
macintux, line 116: In general I prefer to retain unit tests that test a specific historic failure case, but since the failures were stupidity on my part, the code has not yet made it into a release, and the EQC tests do cover them pretty thoroughly, I'll drop them.

src/riak_core_ring_util.erl
macintux: Hence the "this test is of dubious value" comment. :-) Dropped.

src/riak_core_ring_util.erl
macintux: Good idea, implementing.

macintux:
Getting test failures now that I'm explicitly comparing against ring hash values for `boundary_helper/2` because I'm generating hash values for testing greater than the hash space.

The most logical explanation is that `find_near_boundaries/2` is ill-defined:

``` erlang
find_near_boundaries(RingSize, PartitionSize) ->
    ?LET({Id, Offset}, {choose(1, RingSize-1), choose(-RingSize, RingSize)},
          Id * PartitionSize + Offset).
```

I'm calling it quits for the moment, will get back to this later tonight to figure out what I'm doing wrong.

macintux:
Forgot to submit this update. That code is fine, it was the macros that were causing problems. Fixed, now looking into `chashbin` to improve performance.

macintux:
Closing in favor of #638.

PR #633, opened 2014-10-07
Implement key count estimation via AAE trees

jtuple:
This pull-request adds `hashtree:estimate_keys/1` which can be used to estimate the number of keys in a partition by sampling the segments in the relevant AAE tree.

The estimation is calculated by computing a keys-per-segment estimate and then multiplying that estimate by the total number of segments.

The keys-per-segment estimate is determined by traversing over segments until either 1% of segments have been sampled or 1000 keys have been seen.

This key count estimation is used as part of an improvement to AAE-based fullsync replication.

Sibling pull-requests: basho/riak_kv#1030, basho/riak_repl#623

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -153,7 +153,6 @@
 -define(MEM_LEVELS, 0).
 
 -define(NUM_KEYS_REQUIRED, 1000).
--define(MAX_NUM_SEGMENTS, 10000).
 
 -type tree_id_bin() :: <<_:176>>.
 -type segment_bin() :: <<_:256, _:_*8>>.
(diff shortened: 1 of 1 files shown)


src/hashtree.erl
```diff
@@ -428,6 +432,32 @@ read_meta(Key, State) when is_binary(Key) ->
...
+%%       choice since that ensures segments are updated giving a better
+%%       estimate.
+-spec estimate_keys(hashtree()) -> {ok, integer()}.
+estimate_keys(State) ->
+    estimate_keys(State, 0, 0, ?NUM_KEYS_REQUIRED).
+
+estimate_keys(#state{segments=Segments}, CurrentSegment, Keys, MaxKeys)
+  when (CurrentSegment * 100) >= Segments;
+       Keys >= MaxKeys ->
+    {ok, (Keys * Segments) div (CurrentSegment + 1)};
```
engelsanchez: It seems that we add 1 to CurrentSegment here purely to avoid dealing with a divide by zero corner case.  I think we should not. If we want to avoid dividing by zero given bad input (0 segments, or zero MaxKeys, etc), perhaps just add a clause to detect that.

lixen: We start at the first segment 0. We could start with 1 instead and remove the +1

engelsanchez: When segment 0 is read in line 458, the next step is to increment its number and call estimate_keys again in line 459. If at that point the limits have been hit, CurrentSegment is already 1. The only case you will match the condition on 452 with CurrentSegment equals to zero is if Segments is zero, or MaxKeys is zero, or if a silly user starts with Keys set to something that already hits MaxKeys.

engelsanchez: In other words, in the normal case you are dividing by the number of segments you have processed so far + 1.  This introduces a bias.

lixen: You're right. I only looked at the diff shown by github. I will fix it.

engelsanchez:
:+1: 23b8c80
Review finished and AAE FS tests pass when using this riak_test branch: https://github.com/basho/riak_test/pull/714

engelsanchez:
@borshop merge

PR #635, opened 2014-10-15
Add status types and implement transfers-limit

andrewjstone:
For all status types, we should define an explicit data format. The
purpose of this format is to allow different front ends to render the
output. The status types currently defined live in
riak_core_status_types.hrl.

The generic parser of status types lives in riak_core_status:parse/3.
The specific callback functions that do the actual formatting of the
output for the console live in riak_core_console_writer. The functions
riak_core_console:transfers_limit/{0,1} actually use the writer and
print the output for the new implementation of
`riak-admin transfers limit [Node]`.

Additionally add autosizing to riak_core_console_table

When calling riak_core_console_table:print/2 without a spec including
sizes, the table elements will be sized according to the maximum length
of element in each row. If that all fits within the console width,
including borders, then it is printed. If it does not fit, the elements
are reduced in size until everything fits on screen. If the console
width is not detected with io:columns() then the default value,
?MAX_LINE_LEN=100 is used instead.

### DIFF
--- a/src/riak_core_status.erl
+++ b/src/riak_core_status.erl
@@ -25,6 +25,8 @@
          transfers/0,
          transfer_limit/0,
          transfer_limit/1,
+         rpc_transfer_limit/0,
+         rpc_transfer_limit/1,
          ring_status/0]).
 
 %% Status writer API
@@ -64,24 +66,46 @@ ringready() ->
 
 -spec transfer_limit() -> status().
 transfer_limit() ->
-    {Limits, Down} =
-    riak_core_util:rpc_every_member_ann(riak_core_handoff_manager,
-                                        get_concurrency, [], 5000),
-    transfer_limit_status(Limits, Down).
+    Limits0 = riak_core_handoff_manager:get_all_concurrency(),
+    Limits = [{Node, Limit} || {Node, [Limit]} <- Limits0],
+    transfer_limit_status(Limits).
 
 -spec transfer_limit(node()) -> status().
 transfer_limit(Node) ->
-    case riak_core_util:safe_rpc(Node, riak_core_handoff_manager,
-                                 get_concurrency, [], 5000) of
-        {badrpc, rpc_process_down} ->
-            transfer_limit_status([], [Node]);
+    case riak_core_handoff_manager:get_concurrency(Node) of
+        undefined ->
+            [{alert, [{text, io_lib:format("Invalid/unknown node '~s'.", [Node])}]}];
         Limit ->
-            transfer_limit_status([{Node, Limit}], [])
+            transfer_limit_status([{Node, Limit}])
     end.
 
--spec transfer_limit_status([{node(), non_neg_integer()}], [node()]) ->
-    status().
-transfer_limit_status(Limits, Down) ->
+-spec transfer_limit_status([{node(), non_neg_integer()}]) -> status().
+transfer_limit_status(Limits0) ->
+    Schema = [node, limit],
+    Limits = lists:keysort(1, Limits0),
+    Rows = [[Node, Limit] || {Node, Limit} <- Limits],
+    [{table, Schema, Rows}].
+
+-spec rpc_transfer_limit() -> status().
+rpc_transfer_limit() ->
+    {Limits, Down} =
+        riak_core_util:rpc_every_member_ann(riak_core_handoff_manager,
+                                            get_concurrency, [], 5000),
+    rpc_transfer_limit_status(Limits, Down).
+
+-spec rpc_transfer_limit(node()) -> status().
+rpc_transfer_limit(Node) ->
+    case riak_core_util:safe_rpc(Node, riak_core_handoff_manager,
+                                 get_concurrency, [], 5000) of
+        {badrpc, _} ->
+            rpc_transfer_limit_status([], [Node]);
+        Limit ->
+            rpc_transfer_limit_status([{Node, Limit}], [])
+    end.
+
+-spec rpc_transfer_limit_status([{node(), non_neg_integer()}], [node()]) ->
+                                       status().
+rpc_transfer_limit_status(Limits, Down) ->
     Schema = [node, limit],
     Rows = [[Node, Limit] || {Node, Limit} <- Limits],
     Table = {table, Schema, Rows},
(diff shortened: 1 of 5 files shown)


src/riak_core_status.erl
```diff
@@ -23,26 +23,75 @@
 -export([ringready/0,
          all_active_transfers/0,
          transfers/0,
+	 transfer_limit/0,
```
macintux: Isn't it a crime to use tabs? If it isn't, it should be

src/riak_core_status.erl
```diff
@@ -23,26 +23,75 @@
...
-                true ->
-                    Nodes = [N || {N,_} <- Rings],
-                    {ok, Nodes};
-
+		true ->
+		    Nodes = [N || {N,_} <- Rings],
+		    {ok, Nodes};
                 {false, N2} ->
-                    {error, {different_owners, N1, N2}}
+		    {error, {different_owners, N1, N2}}
```
macintux: Even moar criminal, replacing spaces with tabs

src/riak_core_status.erl
```diff
@@ -23,26 +23,75 @@
...
+    status().
+transfer_limit_status(Limits, Down) ->
+    Schema = [node, limit],
+    Rows = [[Node, Limit] || {Node, Limit} <- Limits],
+    Table = {table, Schema, Rows},
+    case Down of 
+	[] ->
+	    [Table];
+	_ ->
+	    NodesDown = {alert, [{column, "(offline)", Down}]},
```
macintux: Having done similar work in the past, I can understand why we're embedding column/row concepts this deeply into the status code, but I'm not sold on error conditions being coerced into the same representations.

src/riak_core_console_table.erl
```diff
@@ -62,15 +81,76 @@ create_table(Spec, [], _Length, IoList) ->
...
-                        Longest = find_longest_field(Rows, length(Total)+1),
+			Longest = find_longest_field(Rows, length(Total)+1),
                         Size = erlang:max(MinSize, Longest),
                         [Size | Total]
                 end, [], Spec),
     lists:reverse(Res).
 
+-spec find_longest_field(list(), pos_integer()) -> non_neg_integer().
+find_longest_field(Rows, ColumnNo) ->
+        lists:foldl(fun(Row, Longest) ->
```
macintux: Sorry, more whitespace nitpicks, indentation level is twice that of the rest of the file here

src/riak_core_console_table.erl
```diff
@@ -85,7 +165,7 @@ row(Spec, Row0) ->
 titles(Spec) ->
     [ $| | lists:reverse(
         ["\n" | lists:foldl(fun({Title, Size}, TitleRow) ->
-                               [align(atom_to_list(Title), Size) | TitleRow]
+                               [align(Title, Size) | TitleRow]
                             end, [], Spec)])].
 
 -spec align(string(), non_neg_integer()) -> iolist().
```
macintux: Looks like this type spec has been wrong all along (Dialyzer, y u no love us anymore?) but the first argument apparently should be `term()` or similar.

include/riak_core_status_types.hrl
```diff
@@ -0,0 +1,19 @@
+%% @doc The following types describe an abstract format for status information.
+%% Each type has a semantic, organizational meaning in a way similar to an html
+%% document. The difference here is that we want our format to use erlang
+%% data structures and types and be able to generate human readable output, json,
+%% csv and a subset of html, as well as other possible output.
+-type value() :: {value, term()}.
+-type text() :: {text, iolist()}.
+-type schema() :: [any()].
+%% A row where the type of each element matches it's place in the schema. A row
```
macintux: "its" (you _have_ seen my Twitter handle, right?)

include/riak_core_status_types.hrl
```diff
@@ -0,0 +1,19 @@
...
+%% is unlabeled data and is only contained inside a table.
+-type row() :: [any()].
+-type rows() :: [row()].
+%% A list of the same type of data. It's really just a labelled list. 
+-type column() :: {column, iolist(), [iolist()]}.
+-type table() :: {table, schema(), rows()} | {table, [column()]}.
+-type alert() :: {alert, [column() | table() | text()]}.
+-type ratio() :: {ratio, integer(), pos_integer() | neg_integer()}.
+-type elem() :: text() | schema() | row() | column() | table() | alert() | ratio() | value.
```
macintux: `value` should be `value()`, neh?

macintux:
Maybe....some tests?

src/riak_core_console_table.erl
```diff
@@ -39,12 +46,23 @@ print(Header, Spec, Rows) ->
     Table = create_table(Spec, Rows),
     io:format("~ts~n~n~ts~n", [Header, Table]).
 
+-spec autosize_create_table([any()], [[any()]]) -> iolist().
+autosize_create_table(Schema, Rows) ->
+    BorderSize = 1 + length(hd(Rows)),
+    MaxLineLen = case io:columns() of
```
macintux: Unless we're sure this will be the only place we'll need to pick a line length in this module, this seems like a natural function

src/riak_core_status.erl
```diff
@@ -23,58 +23,107 @@
...
+                Acc1 ++ Acc;
+            _ ->
+                Acc1 ++ [{stopped, N, length(Stopped)} | Acc]
+        end
+    end,
     {Down, lists:foldl(F, [], Rings)}.
 
 %% @doc Produce status for all active transfers in the cluster.
--spec all_active_transfers() -> {Xfers::list(), Down::list()}.
+-spec all_active_transfers() -> {Xfers :: list(), Down :: list()}.
```
macintux: Particularly given how unhelpful the contents of this function are in visually determining what kind of data is coming out of this, I'd like to see the spec fleshed out with the type of data embedded in those lists.

src/riak_core_status.erl
```diff
@@ -23,58 +23,107 @@
...
 all_active_transfers() ->
     {Xfers, Down} =
-        riak_core_util:rpc_every_member(riak_core_handoff_manager,
-                                        status,
-                                        [{direction, outbound}],
-                                        5000),
+    riak_core_util:rpc_every_member(riak_core_handoff_manager,
```
macintux: Should be indented further

src/riak_core_status.erl
```diff
@@ -82,13 +131,13 @@ ring_status() ->
     %% are running on each node.
     {ok, Ring} = riak_core_ring_manager:get_raw_ring(),
     {AllMods, Down} =
-        riak_core_util:rpc_every_member_ann(riak_core, vnode_modules, [], 1000),
+    riak_core_util:rpc_every_member_ann(riak_core, vnode_modules, [], 1000),
```
macintux: indent

src/riak_core_status.erl
```diff
@@ -23,58 +23,107 @@
...
+        [] ->
+            [Table];
+        _ ->
+            NodesDown = {alert, [{column, "(offline)", Down}]},
+            [Table, NodesDown]
+    end.
 
 -spec(transfers() -> {[atom()], [{waiting_to_handoff, atom(), integer()} |
-                                 {stopped, atom(), integer()}]}).
+{stopped, atom(), integer()}]}).
```
macintux: indent

src/riak_core_status.erl
JeetKunDoug: Unfortunately, what comes out right now is a list of tagged tuple that looks like this:

``` erlang
{status_v2, [{mod, Mod},
                 {src_partition, SrcP},
                 {target_partition, TargetP},
                 {src_node, SrcNode},
                 {target_node, TargetNode},
                 {direction, Dir},
                 {status, Status},
                 {start_ts, StartTS},
                 {sender_pid, TPid},
                 {stats, calc_stats(HO)},
                 {type, Type}]}
```

Not sure how to type that, short of turning it into a record. In fact, we already removed it _from_ a record to make this tagged tuple. Unfortunately, we add a "stats" field to the tuple that is derived from the stats field in the handoff record, but does some calculations, so I don't think we can simply return the riak_core_handoff:handoff_status records.

For context, take a look at:
- `riak_core_handoff_manager:build_status/1`\- https://github.com/basho/riak_core/blob/feature/status-types/src/riak_core_handoff_manager.erl#L47
- `riak_core_handoff_manager:calc_stats/1` -  https://github.com/basho/riak_core/blob/feature/status-types/src/riak_core_handoff_manager.erl#L367

I'm open to suggestions.

macintux:
@borshop merge

macintux:
@borshop merge

PR #637, opened 2014-10-16
Use cluster metadata to get/set transfer limits

andrewjstone:
riak_core_handoff_manager now uses cluster metadata to get and set
transfer limits per node. The configuration in the file overwrites the
current cluster metadata value for that node on startup of the handoff
manager. All sets still use RPC calls to the given node, but then add
them to cluster metadata instead of changing the configured environment
variable. The environment variable now always matches what was in the
config file at node startup.


JeetKunDoug:
We need to take into account cluster leaves with this - when a node leaves a cluster, we're still leaving the metadata in place.

macintux:
@borshop merge

macintux:
@borshop merge

PR #638, opened 2014-10-17
Introduce partition IDs

macintux:
Partition IDs are particularly useful for user-facing code but may have value elsewhere.
- Translation functions from hash (integer or binary) to partition ID and back. Partition IDs are zero-based up to (but not including) ring
  size.
- For user-facing tools, hash_is_partition_boundary/2 can be used to verify whether a hash provided by the user is actually the first in a partition.
- EQC and unit tests.

Supersedes #632 (this is squashed and rebased to our integration branch).

### DIFF
--- a/src/riak_core_ring_util.erl
+++ b/src/riak_core_ring_util.erl
@@ -179,11 +179,14 @@ test_monotonic(TestTimeSecs) ->
         eqc:quickcheck(eqc:testing_time(TestTimeSecs, ?QC_OUT(prop_monotonic()))).
 
 
+%% `prop_only_boundaries' should run a little longer: not quite as
+%% fast, need to scan a larger portion of hash space to establish
+%% correctness
 only_boundaries_test_() ->
-    {timeout, ?TEST_TIME_SECS+5, [?_assert(test_only_boundaries() =:= true)]}.
+    {timeout, ?TEST_TIME_SECS+15, [?_assert(test_only_boundaries() =:= true)]}.
 
 test_only_boundaries() ->
-    test_only_boundaries(?TEST_TIME_SECS).
+    test_only_boundaries(?TEST_TIME_SECS+10).
 
 test_only_boundaries(TestTimeSecs) ->
         eqc:quickcheck(eqc:testing_time(TestTimeSecs, ?QC_OUT(prop_only_boundaries()))).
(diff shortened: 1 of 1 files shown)


src/riak_core_ring_util.erl
```diff
@@ -56,3 +74,185 @@ check_ring(Ring, Nval) ->
                                 ordsets:add_element(PL, Acc)
                         end
                 end, [], Preflists).
+
+-spec hash_to_partition_id(chash:index() | chash:index_as_int(),
+                           pos_integer()) ->
```
andrewjstone, line 79: Can we line this spec up better?

macintux, line 79: That's the default style in Emacs Erlang mode. What would you prefer?

src/riak_core_ring_util.erl
```diff
@@ -56,3 +74,185 @@ check_ring(Ring, Nval) ->
...
+test_only_boundaries() ->
+    test_only_boundaries(?TEST_TIME_SECS).
+
+test_only_boundaries(TestTimeSecs) ->
+        eqc:quickcheck(eqc:testing_time(TestTimeSecs, ?QC_OUT(prop_only_boundaries()))).
+
+%% Partition IDs should map to hash values which are partition boundaries
+prop_ids_are_boundaries() ->
+    ?FORALL(RingPower, choose(2, ?RINGSIZEEXPMAX),
+            ?FORALL(PartitionId, choose(0, ?RINGSIZE(RingPower) - 1),
```
JeetKunDoug, line 197: Andrew and I were both a bit confused about this for a bit just because of the formatting and nesting of calls - any way we can pull:

``` erlang
 riak_core_ring_util:hash_is_partition_boundary(
                      riak_core_ring_util:partition_id_to_hash(PartitionId,
                                                               ?RINGSIZE(RingPower)),
                      ?RINGSIZE(RingPower)) 
```

and similar constructs into their own, well-named functions to make it easier to understand?

jrwest, line 197: fwiw when reading this a week or so back I initially got caught up on the same issue...was scratching my head on why it passed until I opened it in an editor.

JeetKunDoug, line 197: Or, even something like:

``` erlang
prop_ids_are_boundaries() ->
    ?FORALL(RingPower, choose(2, ?RINGSIZEEXPMAX),
            ?FORALL(PartitionId, choose(0, ?RINGSIZE(RingPower) - 1),
            begin
                          PartitionIdHash =  riak_core_ring_util:partition_id_to_hash(PartitionId,
                                                               ?RINGSIZE(RingPower)),   
                          IsBoundary = riak_core_ring_util:hash_is_partition_boundary(PartitionIdHash,
                                    ?RINGSIZE(RingPower)),
                          IsBoundary =:= true
            end
        )).
```

Formatting in a comment is hard - ignore large tabs.
I know it's a bit more verbose, but is much easier to read/understand.

src/riak_core_ring_util.erl
```diff
@@ -56,3 +74,185 @@ check_ring(Ring, Nval) ->
...
+                            end
+                           ))).
+
+%% Using a dictionary lookup to identify partition boundaries is quite
+%% slow, especially given the large hash space we'd like to test.
+%%
+%% Replacing the use of `dict' with something based on `chashbin' may
+%% help significantly, but it's not vital.
+
+%% Hash values which are listed in the ring structure are boundary
```
andrewjstone, line 242: For code below this line a dict with dummy value is used. Instead you could have used a set and not have to fill in dummy. Not a big deal, and no need to change it here, just figured I'd mention it.

macintux, line 242: Good point, will test to see if it's any faster with an ordset.

macintux, line 242: Uh. Yeah. `ordsets` is dramatically faster than `orddict` or `dict` (or, obviously, `proplists`).

macintux:
Used `begin`/`end` for all statements to make for better readability.

macintux:
Thanks @seancribbs, I keep forgetting that exists. Pushed.

JeetKunDoug:
@borshop merge

PR #643, opened 2014-11-06
Patch management

macintux:
Underpinnings of a mechanism for automated patch management. This will play a purely advisory role for operators looking for obsolete patches.


macintux:
May move this out of `riak_core` into its own library sooner rather than later, so don't merge this.

PR #644, opened 2014-11-06
Patch management

macintux:
Same as #643 but for the next 1.4 point release.


macintux:
Build failure appears to be because the builder is using R16. Waiting for advice.

jonmeredith:
You can add an 'erlang_version' file to the root dir of that branch
containing 'r15b01' and it will pick up the right version of erlang.

https://github.com/basho/basho_buildbot/blob/master/master.cfg#L585

On Thu, Nov 6, 2014 at 5:00 PM, John Daily notifications@github.com wrote:

> Build failure appears to be because the builder is using R16. Waiting for
> advice.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/basho/riak_core/pull/644#issuecomment-62074276.

## 

Jon Meredith
Chief Architect
Basho Technologies, Inc.
jmeredith@basho.com

macintux:
Closing this. We don't particularly need this code in 1.4.

PR #646, opened 2014-11-11
Tell the caller when you fail to open a file, too

russelldb:
If you can't open a file, that should be an error too!


slfritchie:
+1

russelldb:
+1 c77b0c7

(Inferred from above)

slfritchie:
@borshop merge

PR #647, opened 2014-11-11
Handoff manager EQC testing

macintux:
Starting with code written by @jrwest long ago, @aberghage and I have created a more comprehensive test suite using `eqc_statem` for `riak_core_handoff_manager`.

Tests include:
- Add outbound and inbound transfers
- Disable future transfers using `riak_core` environment variables
- Control concurrency level using `set_concurrency`
- Retrieve concurrency level using `get_concurrency`

Near-term plans for extending this:
- Test transfer lifecycle (errors at start or during transfer) as part of Alex's improvements to error status management #630 
- Test the more sophisticated enabling/disabling transfers as part of Alex's branch `feature/atb/handoff-enable-disable`
- Test resize transfers

### DIFF
--- a/test/handoff_manager_eqc.erl
+++ b/test/handoff_manager_eqc.erl
@@ -13,13 +13,29 @@
 
 -compile(export_all).
 
--record(state,{handoffs=[],max_concurrency=0}).
+-define(QC_OUT(P),
+        eqc:on_output(fun(Str, Args) -> io:format(user, Str, Args) end, P)).
+
+%% How many times we should sleep before giving up on handoff
+%% processes that we expected to terminate
+-define(MAX_EXIT_WAITS, 5).
+
+%% How many milliseconds we should sleep waiting on processes to die
+-define(EXIT_WAIT, 50).
+
+-record(state,{handoffs=[],max_concurrency=2,inbound_count=0}).
+
+
+check() ->
+    eqc:check(eqc_statem:show_states(?QC_OUT(prop_handoff_manager()))).
+
+
 
 %% @doc Returns the state in which each test case starts. (Unless a different
 %%      initial state is supplied explicitly to, e.g. commands/2.)
 -spec initial_state() -> eqc_statem:symbolic_state().
 initial_state() ->
-    #state{handoffs=[],max_concurrency=2}.
+    #state{}.
 
 %% ------ Common pre-/post-conditions
 %% @doc General command filter, checked before a command is generated.
@@ -44,73 +60,44 @@ postcondition_common(_S, _Call, _Res) ->
 %% @doc add_inbound_command - Command generator
 %% We can send an empty list as the `SSLOpts' argument to
 %% `add_inbound/1` since we mock `handoff_receiver'
--spec add_inbound_command(S :: eqc_statem:symbolic_state()) ->
-        eqc_gen:gen(eqc_statem:call()).
-add_inbound_command(_S) ->
-    {call, riak_core_handoff_manager, add_inbound, [[]]}.
+add_inbound(SSLOpts) ->
+    riak_core_handoff_manager:add_inbound(SSLOpts).
 
-%% @doc add_inbound_pre - Precondition for generation
--spec add_inbound_pre(S :: eqc_statem:symbolic_state()) -> boolean().
-add_inbound_pre(_S) ->
-    true. %% Condition for S
-
-%% @doc add_inbound_pre - Precondition for add_inbound
--spec add_inbound_pre(S :: eqc_statem:symbolic_state(),
-                      Args :: [term()]) -> boolean().
-add_inbound_pre(_S, _Args) ->
-    true. %% Condition for S + Args
+-spec add_inbound_args(S :: eqc_statem:symbolic_state()) ->
+                                 list().
+add_inbound_args(_S) ->
+    [[]].
 
 %% @doc add_inbound_next - Next state function
 -spec add_inbound_next(S :: eqc_statem:symbolic_state(),
                        V :: eqc_statem:var(),
                        Args :: [term()]) -> eqc_statem:symbolic_state().
-add_inbound_next(S, _Value, _Args) ->
-    S.
+add_inbound_next(S=#state{handoffs=Handoffs,max_concurrency=MaxConcurrency}, _Value, _Args) when length(Handoffs) >= MaxConcurrency ->
+    S;
+add_inbound_next(S=#state{handoffs=Handoffs,inbound_count=IC}, Value, _Args) ->
+    S#state{handoffs=Handoffs++[{{inbound, IC}, Value}],
+            inbound_count=IC+1}.
+
 
 %% @doc add_inbound_post - Postcondition for add_inbound
 -spec add_inbound_post(S :: eqc_statem:dynamic_state(),
                        Args :: [term()], R :: term()) -> true | term().
-add_inbound_post(_S, _Args, _Res) ->
-    true.
-
-%% @doc add_inbound_blocking - Is the operation blocking in this State
-%% -spec add_inbound_blocking(S :: eqc_statem:symbolic_state(),
-%%         Args :: [term()]) -> boolean().
-%% add_inbound_blocking(_S, _Args) ->
-%%   false.
-
-%% @doc add_inbound_adapt - How to adapt a call in this State
-%% -spec add_inbound_adapt(S :: eqc_statem:symbolic_state(),
-%%     Args :: [term()]) -> boolean().
-%% add_inbound_adapt(_S, _Args) ->
-%%   exit(adapt_not_possible).
-
-%% @doc add_inbound_dynamicpre - Dynamic precondition for add_inbound
-%% -spec add_inbound_dynamicpre(S :: eqc_statem:dynamic_state(),
-%%         Args :: [term()]) -> boolean().
-%% add_inbound_dynamicpre(_S, _Args) ->
-%%   true.
-
+add_inbound_post(#state{handoffs=Handoffs}, _Args, {ok, _Pid}) ->
+    eq(current_concurrency(inbound), length(valid_handoffs(Handoffs, inbound)) + 1);
+add_inbound_post(#state{handoffs=Handoffs}, _Args, {error,max_concurrency}) ->
+    eq(current_concurrency(inbound), length(valid_handoffs(Handoffs, inbound))).
 
 %% ------ Grouped operator: kill_handoffs
 %% kill_handoffs is a function of limited utility, only invoked by the
 %% object version upgrade/downgrade via riak_kv_reformat:run/2.
 %% @doc kill_handoffs_command - Command generator
--spec kill_handoffs_command(S :: eqc_statem:symbolic_state()) ->
-        eqc_gen:gen(eqc_statem:call()).
-kill_handoffs_command(_S) ->
-    {call, riak_core_handoff_manager, kill_handoffs, []}.
+kill_handoffs() ->
+    riak_core_handoff_manager:kill_handoffs().
 
-%% @doc kill_handoffs_pre - Precondition for generation
--spec kill_handoffs_pre(S :: eqc_statem:symbolic_state()) -> boolean().
-kill_handoffs_pre(_S) ->
-    true. %% Condition for S
-
-%% @doc kill_handoffs_pre - Precondition for kill_handoffs
--spec kill_handoffs_pre(S :: eqc_statem:symbolic_state(),
-                        Args :: [term()]) -> boolean().
-kill_handoffs_pre(_S, _Args) ->
-    true. %% Condition for S + Args
+-spec kill_handoffs_args(S :: eqc_statem:symbolic_state()) ->
+                                 list().
+kill_handoffs_args(_S) ->
+    [].
 
 %% @doc kill_handoffs_next - Next state function
 -spec kill_handoffs_next(S :: eqc_statem:symbolic_state(),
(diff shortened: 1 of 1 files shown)


test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+%% How many times we should sleep before giving up on handoff
+%% processes that we expected to terminate
+-define(MAX_EXIT_WAITS, 5).
+
+%% How many milliseconds we should sleep waiting on processes to die
+-define(EXIT_WAIT, 50).
+
+-record(state,{handoffs=[],max_concurrency=2,inbound_count=0,
+               outbound_disabled=false, inbound_disabled=false}).
+
```
andrewjstone, line 45: There should be an eunit function wrapping the property so that it will run with `make test` and `rebar eunit skip_deps=true suites=handoff_manager_eqc` like this: https://github.com/basho/riak_core/blob/develop/test/hashtree_eqc.erl#L15

andrewjstone, line 45: The module should be wrapped with -ifdef(TEST) and -ifdef(EQC) like this: https://github.com/basho/riak_core/blob/develop/test/hashtree_eqc.erl#L4

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+        eqc:on_output(fun(Str, Args) -> io:format(user, Str, Args) end, P)).
+
+%% How many times we should sleep before giving up on handoff
+%% processes that we expected to terminate
+-define(MAX_EXIT_WAITS, 5).
+
+%% How many milliseconds we should sleep waiting on processes to die
+-define(EXIT_WAIT, 50).
+
+-record(state,{handoffs=[],max_concurrency=2,inbound_count=0,
```
andrewjstone, line 43: Can you change the formatting here to each member on a separate line? Also add type specs. That makes it easier to read IMO.

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+            %% Don't allow rolling start of vnodes as it will cause a
+            %% race condition with `all_nodes'.
+            {vnode_rolling_start, 0}],
+    OldVars = [begin
+                   Old = app_helper:get_env(riak_core, AppKey),
+                   ok = application:set_env(riak_core, AppKey, Val),
+                   {AppKey, Old}
+               end || {AppKey, Val} <- Vars],
+    %% our mock vnode don't need async pools for this test
+    application:set_env(riak_core, core_vnode_eqc_pool_size, 0),
```
andrewjstone, line 446: This setting should just move into `Vars` list.

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+    riak_core_vnode_sup:start_link(),
+    riak_core_eventhandler_sup:start_link(),
+    riak_core_vnode_proxy_sup:start_link(),
+    riak_core_vnode_manager:start_link(),
+    riak_core_handoff_manager:start_link(),
+    riak_core_handoff_sender_sup:start_link(),
+    riak_core_handoff_receiver_sup:start_link(),
+    riak_core:register([{vnode_module, mock_vnode}]),
+
+    meck:new(riak_core_handoff_sender),
```
andrewjstone, line 459: I think all this meck stuff for handoff sender and receiver has to come before you call 
`riak_core_handoff_{sender | receiver}_sup:start_link()`

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+    meck:new(riak_core_handoff_receiver),
+    meck:expect(riak_core_handoff_sender, start_link,
+                fun(_TargetNode, _Mod, _TypeAndOpts, _Vnode) ->
+                        Pid = spawn_link(fun() ->
+                                           timer:sleep(20000)
```
andrewjstone, line 464: As long as this timeout is greater than the runtime of the test this should work fine. It's a bit hacky though. It's better to have a receive statement without a timeout and to return the pids from setup. After the test completes you can call `teardown(Pids)` and send the message to cause them to exit or just call `exit/2`.

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+         {eqc_gen:oneof([ownership_handoff, hinted_handoff, repair]),
+          eqc_gen:oneof([0, 1, 2, 3, 4, 5, 6, 7]),
+          node(),
+          eqc_gen:nat(), eqc_gen:oneof([bytes, objects])},
+         [ HandoffType,
+           mock_vnode,
+           PartID * chash:ring_increment(8),
+           Node,
+           {call,?MODULE,get_vnode_pid,
+            [PartID * chash:ring_increment(8), mock_vnode]},
```
andrewjstone, line 100: The magic number `8` should be pulled out of here and `setup/0` and put in a macro or function.

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+    {ok, Pid} =
+        riak_core_vnode_master:get_vnode_pid(HashId, Module),
+    Pid.
+
+-spec add_outbound_args(S :: eqc_statem:symbolic_state()) ->
+                               list().
+add_outbound_args(_S) ->
+    ?LET({HandoffType, PartID, Node, Size, SizeType},
+         {eqc_gen:oneof([ownership_handoff, hinted_handoff, repair]),
+          eqc_gen:oneof([0, 1, 2, 3, 4, 5, 6, 7]),
```
andrewjstone, line 92: This is tied to the ring size (the magic number 8 below). Can do something like `lists:seq(0, ?RING_SIZE)` instead.

test/handoff_manager_eqc.erl
andrewjstone, line 459: Maybe it doesn't actually. I guess those processes are all started dynamically.

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+%% No existing (or no successful existing), max concurrency reached
+add_outbound_if(S, _ListOfResults, false, true, _Result, _Idx) ->
+    S;
+%% Existing, has been started successfully
+add_outbound_if(S, _ListOfResults, true, _Max, _Result, _Idx) ->
+    S;
+%% No match
+add_outbound_if(#state{handoffs=Handoffs}=S, ListOfResults, false, false, Result, Idx) ->
+    S#state{handoffs=lists:keystore({outbound, Idx}, 1,
+                                    Handoffs, {{outbound, Idx}, [Result] ++ ListOfResults})}.
```
andrewjstone, line 126: Better to use `[Result | ListOfResults]`

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+
+%% @doc add_outbound_next - Next state function
+
+-spec add_outbound_next(S :: eqc_statem:symbolic_state(),
+                        V :: eqc_statem:var(),
+                        Args :: [term()]) -> eqc_statem:symbolic_state().
+add_outbound_next(#state{handoffs=Handoffs,max_concurrency=MaxConcurrency}=S, Result, Args) ->
+    Idx = lists:nth(3, Args),
+    PreviousResults = proplists:get_value({outbound, Idx}, Handoffs, []),
+    PreviousOk = has_outbound(PreviousResults),
```
andrewjstone, line 112: This is one of those cases where the function is less clear than the line it wraps. maybe call it `has_ok/1`, or just call `lists:keymember/3` directly

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+    add_outbound_if(S, PreviousResults, PreviousOk,
```
andrewjstone, line 113: It's hard to tell what this function is supposed to do. I think the following looks clearer. You could put it in it's own function. Our typical naming convention would also call the function `maybe_add_outbound` instead of `add_outbound_if`.

``` erlang
ReachedMaxConcurrency =  length(valid_handoffs(Handoffs)) >= MaxConcurrency,
case PreviousOk orelse ReachedMaxConcurrency of
    true ->
        S;
    false ->
        Key = {outbound, Idx},
        Results = [Result | ListOfResults],
        S#state{handoffs=lists:keystore(Key, 1, Handoffs, {Key, Results})}
end
```

andrewjstone, line 113: Despite my refactoring above, I have to wonder why we are keeping a list of results at all if we are just checking that a single outbound connection was setup at one point. In that case a boolean should suffice.

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+
+
+
+
+%% @doc add_outbound_post - Postcondition for add_outbound
+-spec add_outbound_post(S :: eqc_statem:dynamic_state(),
+                        Args :: [term()], R :: term()) -> true | term().
+add_outbound_post(#state{handoffs=Handoffs,max_concurrency=MaxConcurrency}, _Args, {error, max_concurrency}) ->
+    outbound_disabled() orelse
+        length(valid_handoffs(Handoffs)) >= MaxConcurrency;
```
andrewjstone, line 136: When would this ever occur? Don't we prevent adding it to the state in the next_state functions of add_inbound and add_outbound if we are already past max concurrency. It seems like if we ever go over MaxConcurrency we have a bug.

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+add_inbound_args(_S) ->
+    [[]].
+
+%% @doc add_inbound_next - Next state function
+-spec add_inbound_next(S :: eqc_statem:symbolic_state(),
+                       V :: eqc_statem:var(),
+                       Args :: [term()]) -> eqc_statem:symbolic_state().
+
+add_inbound_next(S=#state{handoffs=Handoffs,max_concurrency=MaxConcurrency,inbound_count=IC}, Value, _Args) ->
+    case length(valid_handoffs(Handoffs)) >= MaxConcurrency of
```
andrewjstone, line 161: All the places where the following is used, should probably be replaced with a function.
`length(valid_handoffs(Handoffs)) >= MaxConcurrency`

 I also think we should be specific and only check for  `=:= MaxConcurrency`. If we are trying to model never adding more handoffs than the max concurrency in the `next_state` functions we should fail in the `postcondition` if that actually occurs.

test/handoff_manager_eqc.erl
```diff
@@ -0,0 +1,556 @@
...
+                         V :: eqc_statem:var(),
+                         Args :: [term()]) -> eqc_statem:symbolic_state().
+kill_handoffs_next(S, _Value, _Args) ->
+    S#state{handoffs=[],max_concurrency=0}.
+
+%% @doc kill_handoffs_post - Postcondition for kill_handoffs
+-spec kill_handoffs_post(S :: eqc_statem:dynamic_state(),
+                         Args :: [term()], R :: term()) -> true | term().
+kill_handoffs_post(_S, _Args, ok) ->
+    try_wait(current_concurrency() =< 0, 0, ?MAX_EXIT_WAITS).
```
andrewjstone, line 203: current_concurrency() will never return a negative. We should be precise in our expectations. This occurs elswhere in the test.

test/handoff_manager_eqc.erl
macintux, line 459: tbh, unless Alex did this on his own, we didn't look closely at `setup`. Ripped that straight from Jordan's work.

test/handoff_manager_eqc.erl
macintux, line 113: We do need to keep the pid, at least, but I don't remember whether we had a good reason to maintain the full list. Another remnant of Jordan's approach that we didn't rework.

test/handoff_manager_eqc.erl
macintux, line 161: As with the `current_concurrency() =< 0` line below, I never use equality when checking for boundary conditions because dragons. I'd be willing to reconsider, at least for test purposes.

test/handoff_manager_eqc.erl
bearcage, line 459: I did review this stuff, yeah -- I haven't pared down to the minimal necessary set though, which is perhaps something we should consider. I am pretty confident that we don't have ordering problems as-written, though.

test/handoff_manager_eqc.erl
bearcage, line 126: Is this a style guideline or a performance one? I've not looked at the implementation difference between single-element concat and cons, but I do agree it reads clearer with the latter.

test/handoff_manager_eqc.erl
macintux, line 113: On further review, we're ok with keeping `error` (or `{error, Reason}`) vs `{ok, pid}` until proven otherwise.

test/handoff_manager_eqc.erl
macintux, line 464: We'd like to leave this as is: a real-soon-now branch testing Alex's termination status changes will rework entirely how we manage process lifecycle.

test/handoff_manager_eqc.erl
bearcage, line 113: This isn't strictly speaking necessary for the level of testing done right now, but will be when we add sender termination case testing for #630 we'll need at least the pid of a running transfer. I do agree though that we can discard the list of failed ones, no need to track it at present.

test/handoff_manager_eqc.erl
andrewjstone, line 126: I was thinking there was a performance issue, but you are correct in this instance I think. Using `cons` is `O(1)` by just adding to the head of a linked list.  Using `FrontList ++ ListOfResults` is `O(length(FrontList))` which is 1 in this case.

test/handoff_manager_eqc.erl
andrewjstone, line 161: I strongly disagree here. 'm saying you should always be precise in what you check for. Otherwise you have let unknown bugs creep in. What is the type returned by current_concurrency() here? Is it the same type as what it's declared as elsewhere? My guess is that it's defined (or should be) as non_neg_int(). You want to know if current_concurrency ever goes below 0. That's an invariant violation. You have no idea how that will affect you. The code (and dialyzer) should be preventing that, but if it's not it should crash if it ever sees that value because it's unexpected. Accepting that value leads to it being used in other code where the assumption that it's never less than 0 could be made. It doesn't matter if the code is test or production, you still want to type things properly and check for values that match that type.

test/handoff_manager_eqc.erl
andrewjstone, line 464: That's fine by me.

test/handoff_manager_eqc.erl
andrewjstone, line 459: No worries, you can leave it for now. I don't think it's an issue either.

macintux:
@aberghage Are you still working on this?

bearcage:
Yep, now that I've got some time to clean this up I'll address the rest of ajs' comments and bring this up to date. Did you have any other concerns not currently reflected here?

macintux:
@aberghage Way too far in the past for me to have any meaningful input. If you need help bootstrapping yourself back into this code, I'd be happy to collaborate (assuming SH still works, dammit).

Issue #634, opened 2014-10-08
handoff.ip is missing from riak_core.schema


jrwest:
when it is added we should add a validation to ensure people _do not_ set it to `127.0.0.1` -- it can cause problems.

andrewjstone:
Fixed by #653

Basho-JIRA:
This was resolved a while ago by https://github.com/basho/riak_core/pull/653 — time to learn how to mark a jira issue closed. 

**_[posted via JIRA by Alex Berghage]**_

Basho-JIRA:
Reopening b/c JIRA workflow was screwy when I closed this, so it never got a resolution field and still shows in my dash. 

**_[posted via JIRA by Alex Berghage]**_

PR #653, opened 2014-11-18
Add handoff.ip to core cuttlefish schema [JIRA: RIAK-3058]

bearcage:
Items for discussion:
- Included a validator prohibiting loopback per issue report — seems like a sane choice to me since handoff on the loopback interface is only useful for devrels, and in that case 0.0.0.0 is fine too.
- I'm not clear on if/how validators are currently tested, so I've neglected to write a test for that as yet.

### DIFF
--- a/priv/riak_core.schema
+++ b/priv/riak_core.schema
@@ -71,6 +71,23 @@
   {commented, "$(platform_etc_dir)/cacertfile.pem"}
 ]}.
 
+%% @doc handoff.ip is the network address that Riak binds to for
+%% intra-cluster data handoff.
+{mapping, "handoff.ip", "riak_core.handoff_ip", [
+  {default, "{{handoff_ip}}" },
+  {datatype, string},
+  {validators, ["not_loopback_address"]},
+  hidden
+]}.
+
+{validator,
+  "not_loopback_address",
+  "IP address must not be a loopback interface",
+  fun(Address) -> Address /= "127.0.0.1"
+          andalso Address /= "0:0:0:0:0:0:0:1"
+          andalso Address /= "::1"
+  end}.
+
 %% @doc handoff.port is the TCP port that Riak uses for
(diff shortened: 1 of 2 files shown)


andrewjstone:
@aberghage I'm against the validator prohibiting loopback. I don't want to open up a handoff port publicly if I'm just running tests on my laptop.

bearcage:
I don't have a strong feeling there; I added it since it was requested in 634, but I can see the argument either way.

seancribbs:
Agree with @andrewjstone, validating that it's a correct IPv4/6 address should be sufficient.

`inet_parse:address/1` is what you want.

jrwest:
fwiw, since I left the original comment in #634, I think its more important to prevent the user from doing something silly. I understand the concern while developing but I think its better to put the burden on the riak developer to protect their machine than the user from ending up in a confusing config state.

bearcage:
Now that we're out of the collective turkey coma, any objections to merging this?

andrewjstone:
:+1:  73a0bbd 
sorry @aberghage for the delayed +1

seancribbs:
Should have a :+1: from another CSE.

pyrrho:
:+1:
I mentioned to Alex in HipChat that I was surprised that the `invalid_states_test` is new. I'm assuming it's named as such to allow it to expand into a more general test.

bearcage:
Yep, that's the notion. The core schema test suite previously didn't include anything to exercise validators, so that adds it.

bearcage:
@borshop merge

bashopatricia:
create jira issue

Basho-JIRA:
Per Brian, don't need to touch this work, 2.0 and develop-2.2 branches have most up to date code.  [~bsparrow435] Can you close this if no longer needed?  Thanks 

***[posted via JIRA by Patricia Brewer]***

PR #654, opened 2014-11-19
Remove afunix from EXOMETER_PACKAGES.

uwiger:
Afunix is not used at this time, and may cause build issues on some
platforms.


jburwell:
Checked the buildbot get dependencies log and afunix was not imported.

jburwell:
:+1: 837bce2

jburwell:
@borshop merge

PR #657, opened 2014-11-25
Minimal riak_cli integration

jadeallenx:
I believe this is the bare minimum to get riak_cli into core and ready for future command implementation.


andrewjstone:
:+1: ec14c9f

PR #658, opened 2014-12-04
Centralize CLI registration

jadeallenx:
Provide a mechanism to centralize riak_cli command and usage registration.

Modules should implement their own `register_cli/0` functions and then
add the module name(s) to the list defined in the `?CLI_MODULES` 
macro.

The `riak_core_cli_registry:register_cli/0` function is called from
`riak_core_app:start/2` around line 104.


andrewjstone:
:+1:

PR #659, opened 2014-12-05
Implement cluster CLI commands

jadeallenx:
Implement the following commands using riak_cli:
- cluster status (improved member-status)
- partition-count
- partitions
- partition id=<id>
- partition index=<index>
- members

TODO: After merge, some code clean up and boundary checking.

Depends on basho/riak_ee#295

### DIFF
--- a/src/riak_core_cluster_cli.erl
+++ b/src/riak_core_cluster_cli.erl
@@ -30,15 +30,20 @@
 
 -module(riak_core_cluster_cli).
 
+-behaviour(riak_cli_handler).
+
 -export([
-    register_all_commands/0,
-    register_all_usage/0,
+    register_cli/0,
     status/2,
     partition_count/2,
     partitions/2,
     partition/2,
     members/2
 ]).
+
+register_cli() ->
+    register_all_usage(),
+    register_all_commands().
 
 register_all_commands() ->
     lists:foreach(fun(Args) -> apply(riak_cli, register_command, Args) end,
@@ -99,12 +104,14 @@ status([], []) ->
     Rows = [ format_status(Node, Status, Ring, RingStatus) ||
       {Node, Status} <- AllStatus ],
 
-    Table = [riak_cli_status:table(Rows)],
+    Table = riak_cli_status:table(Rows),
 
-    io:format("---- Cluster Status ----~n~n", []),
-    io:format("Ring ready: ~p~n~n", [element(2, RingStatus)]),
-    riak_cli:print(Table),
-    io:format("Key: (C) = Claimant; availability marked with '!' is unexpected~n~n", []).
+    T0 = riak_cli_status:text("---- Cluster Status ----"),
+    T1 = riak_cli_status:text(io_lib:format("Ring ready: ~p~n", [element(2, RingStatus)])),
+    T2 = riak_cli_status:text(
+           "Key: (C) = Claimant; availability marked with '!' is unexpected"),
+    Out = riak_cli_status:alert([T0,T1,Table,T2]),
+    riak_cli:print([Out]).
 
 format_status(Node, Status, Ring, RingStatus) ->
     {Claimant, _RingReady, Down, MarkedDown, Changes} = RingStatus,
@@ -160,7 +167,10 @@ partition_count([], [{node, Node}]) ->
     riak_cli:print(Table);
 partition_count([], []) ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
-    io:format("Cluster-wide partition-count: ~p~n", [riak_core_ring:num_partitions(Ring)]).
+    T = riak_cli_status:text(
+         io_lib:format("Cluster-wide partition-count: ~p", 
+                       [riak_core_ring:num_partitions(Ring)])),
+    riak_cli:print([T]).
 
 %%%
 %% cluster partitions
@@ -196,12 +206,13 @@ partitions_output(Node) ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     RingSize = riak_core_ring:num_partitions(Ring),
     {Primary, Secondary, Stopped} = riak_core_status:partitions(Node, Ring),
-    io:format("Partitions owned by ~p:~n", [Node]),
+    T0 = riak_cli_status:text(io_lib:format("Partitions owned by ~p:", [Node])),
     Rows = generate_rows(RingSize, primary, Primary) 
            ++ generate_rows(RingSize, secondary, Secondary)
            ++ generate_rows(RingSize, stopped, Stopped),
-    Table = [riak_cli_status:table(Rows)],
-    riak_cli:print(Table).
+    Table = riak_cli_status:table(Rows),
+    Alert = riak_cli_status:alert([T0, Table]),
+    riak_cli:print([Alert]).
 
 generate_rows(_RingSize, Type, []) ->
     [[{type, Type}, {index, "--"}, {id, "--"}]];
@@ -256,14 +267,18 @@ id_out(InputType, Number) ->
 %% than the ringsize - 1 partition id
 id_out1(index, Index, Ring, RingSize) ->
     Owner = riak_core_ring:index_owner(Ring, Index),
-    io:format("Partition index: ~p -> id: ~p~n\t(owner: ~p)~n", 
-              [Index, hash_to_partition_id(Index, RingSize), Owner]);
+    T = riak_cli_status:text(
+          io_lib:format("Partition index: ~p -> id: ~p~n\t(owner: ~p)", 
+              [Index, hash_to_partition_id(Index, RingSize), Owner])),
+    riak_cli:print([T]);
 
 id_out1(id, Id, Ring, RingSize) when Id < RingSize ->
     Idx = partition_id_to_hash(Id, RingSize),
     Owner = riak_core_ring:index_owner(Ring, Idx),
-    io:format("Partition id: ~p -> index: ~p~n\t(owner: ~p)~n", 
-              [Id, partition_id_to_hash(Id, RingSize), Owner]);
+    T = riak_cli_status:text(
+          io_lib:format("Partition id: ~p -> index: ~p~n\t(owner: ~p)~n", 
+              [Id, partition_id_to_hash(Id, RingSize), Owner])),
+    riak_cli:print([T]);
 id_out1(id, Id, _Ring, _RingSize) ->
     {error, ["Id ", Id, " is invalid."]}.
 
@@ -300,7 +315,8 @@ members([], []) ->
     ).
 
 member_output(L) ->
-    lists:foreach(fun({N, _S}) -> io:format("~p~n", [N]) end, L).
+    O = [ riak_cli_status:text(atom_to_list(N)) || {N, _S} <- L ],
+    riak_cli:print(O).
 
 get_status() ->
     {ok, Ring} = riak_core_ring_manager:get_my_ring(),
--- a/src/riak_core_cluster_cli.erl
+++ b/src/riak_core_cluster_cli.erl
@@ -118,10 +118,10 @@ is_claimant(Node, _Other) ->
 
 node_availability(Node, Down, MarkedDown) ->
     case {lists:member(Node, Down), lists:member(Node, MarkedDown)} of
-        {false, false} -> "  up   "; % put common case at the top
-          {true, true} -> " down  ";
-         {true, false} -> " down! ";
-         {false, true} -> "  up!  "
+         {false, false} -> "  up   "; 
+         {true,  true } -> " down  ";
+         {true,  false} -> " down! ";
+         {false, true } -> "  up!  "
     end.
 
 %%%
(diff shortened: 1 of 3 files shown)


src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,334 @@
...
+      {Node, Status} <- AllStatus ],
+
+    Table = riak_cli_status:table(Rows),
+
+    T0 = riak_cli_status:text("---- Cluster Status ----"),
+    T1 = riak_cli_status:text(io_lib:format("Ring ready: ~p~n", [element(2, RingStatus)])),
+    T2 = riak_cli_status:text(
+           "Key: (C) = Claimant; availability marked with '!' is unexpected"),
+    Out = riak_cli_status:alert([T0,T1,Table,T2]),
+    riak_cli:print([Out]).
```
andrewjstone: I've been thinking about this a bit. I didn't really intend for implementers to constantly call `riak_cli:print/1`. I intended for it to be called automatically and use the proper writer module given global --csv flags etc..., but of course I didn't implement it that way. I spent a lot of time focusing on config the last few weeks and if you look in riak_cli.erl you'll see that the config ops (show, describe, set) wrap the output in a print statement, and the callbacks themselves return errors tuples or status types. Returning errors works fine for config because it's all internal and riak_cli_error knows how to parse the riak_cli specific errors and generate alerts from them. However, since that requires modifying riak_cli itself, outside users of riak_cli should only ever return status types in callbacks. 

This was a mistake in implementation that I will fix on Monday. We can then modify each PR to do the right thing and only return a status(). In your case it will just be removing the riak_cli:print/1 calls. Hopefully the same elsewhere.

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,334 @@
...
+
+id_out1(id, Id, Ring, RingSize) when Id < RingSize ->
+    Idx = partition_id_to_hash(Id, RingSize),
+    Owner = riak_core_ring:index_owner(Ring, Idx),
+    T = riak_cli_status:text(
+          io_lib:format("Partition id: ~p -> index: ~p~n\t(owner: ~p)~n", 
+              [Id, partition_id_to_hash(Id, RingSize), Owner])),
+    riak_cli:print([T]);
+id_out1(id, Id, _Ring, _RingSize) ->
+    {error, ["Id ", Id, " is invalid."]}.
```
andrewjstone: This doesn't mean anything to riak_cli. I'm not sure what exactly you are trying to do here. When we fix things up it should return an alert status.

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,334 @@
...
+
+member_output(L) ->
+    O = [ riak_cli_status:text(atom_to_list(N)) || {N, _S} <- L ],
+    riak_cli:print(O).
+
+get_status() ->
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    lists:keysort(2, riak_core_ring:all_member_status(Ring)).
+
+%%% FIXME! -> REMOVE AFTER MERGE
```
andrewjstone, line 287: Is this John's code that needs to be merged into the new integration branch?

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,334 @@
...
+                   partitions_usage(), partition_usage(),
+                   members_usage()
+                  ]).
+
+%%%
+%% Cluster status
+%%%
+
+status_register() ->
+    [
```
andrewjstone: Just my personal style opinion, but I don't like the brackets on separate lines in erlang. I prefer something like this, but It's your decision.

``` erlang
       [["riak-admin", "cluster", "status"], % Cmd
        [],                                  % KeySpecs
        [],                                  % FlagSpecs
        fun status/2].                       % Implementation callback
```

src/riak_core_cluster_cli.erl
jadeallenx: OK, gotcha. Now that basho/riak_cli#14 has been merged, I will fix it up.

src/riak_core_cluster_cli.erl
jadeallenx: OK, will fix.

src/riak_core_cluster_cli.erl
jadeallenx, line 287: Yes. It is, but I couldn't find a clean commit to cherry pick so I just cut'n'pasted; my plan was to rework these calls once we have a branch with all the things.

src/riak_core_cluster_cli.erl
jadeallenx: OK. I don't have a strong preference, but I found it avoided merged conflicts sometimes. I will fix it up.

jadeallenx:
@andrewjstone I've pushed a new commit responding to your code review.

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,289 @@
...
+format_status(Node, Status, Ring, RingStatus) ->
+    {Claimant, _RingReady, Down, MarkedDown, Changes} = RingStatus,
+    [{node, is_claimant(Node, Claimant)},
+     {status, Status},
+     {avail, node_availability(Node, Down, MarkedDown)},
+     {ring, claim_percent(Ring, Node)},
+     {pending, future_claim_percentage(Changes, Ring, Node)}
+    ].
+
+is_claimant(Node, Node) ->
```
andrewjstone, line 114: This is totally a bikeshed comment and won't hold up a +1 of this PR. I was just thinking that we should align this better. Either put the (c) after the Node name, since it's left aligned, or  put 3 spaces before the non-claimant nodes so it is right - aligned.

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,289 @@
...
+    Rows = [ format_status(Node, Status, Ring, RingStatus) ||
+      {Node, Status} <- AllStatus ],
+
+    Table = riak_cli_status:table(Rows),
+
+    T0 = riak_cli_status:text("---- Cluster Status ----"),
+    T1 = riak_cli_status:text(io_lib:format("Ring ready: ~p~n", [element(2, RingStatus)])),
+    T2 = riak_cli_status:text(
+           "Key: (C) = Claimant; availability marked with '!' is unexpected"),
+    [riak_cli_status:alert([T0,T1,Table,T2])].
```
andrewjstone: Why did you wrap these elements in alerts?

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,289 @@
...
+partitions_output(Node) ->
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    RingSize = riak_core_ring:num_partitions(Ring),
+    {Primary, Secondary, Stopped} = riak_core_status:partitions(Node, Ring),
+    T0 = riak_cli_status:text(io_lib:format("Partitions owned by ~p:", [Node])),
+    Rows = generate_rows(RingSize, primary, Primary) 
+           ++ generate_rows(RingSize, secondary, Secondary)
+           ++ generate_rows(RingSize, stopped, Stopped),
+    Table = riak_cli_status:table(Rows),
+    [riak_cli_status:alert([T0, Table])].
```
andrewjstone: Why use an alert here?

src/riak_core_cluster_cli.erl
jadeallenx: I think it's just my general misunderstanding of how status types ought to be used, but easily fixed.

src/riak_core_cluster_cli.erl
jadeallenx: Will fix.

jadeallenx:
Ping @andrewjstone

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,288 @@
...
+
+is_claimant(Node, Node) ->
+    " (C) " ++ atom_to_list(Node) ++ " ";
+is_claimant(Node, _Other) ->
+    "     " ++ atom_to_list(Node) ++ " ".
+
+node_availability(Node, Down, MarkedDown) ->
+    case {lists:member(Node, Down), lists:member(Node, MarkedDown)} of
+        {false, false} -> "  up   "; % put common case at the top
+          {true, true} -> " down  ";
```
andrewjstone: Can you line this clause up please?

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,288 @@
...
+%%%
+%% cluster partition-count
+%%%
+
+partition_count_register() ->
+    [
+     ["riak-admin", "cluster", "partition-count"], % Cmd
+     [],                                           % KeySpecs
+     [{node, [{shortname, "n"}, {longname, "node"}, 
+              {typecast, fun list_to_atom/1}]}],   % FlagSpecs
```
andrewjstone: This should change to use `riak_cli_typecast:to_node/1`. Otherwise you end up with the following: 

```
Andrews-MacBook-Pro:riak_ee ajs$ dev/dev1/bin/riak-admin cluster partition-count --node=dev2
+----+----------+-----+
|node|partitions| pct |
+----+----------+-----+
|dev2|    0     |  0.0|
+----+----------+-----+
```

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,288 @@
...
+
+%%%
+%% cluster partitions
+%%%
+
+partitions_register() ->
+    [["riak-admin", "cluster", "partitions"],      % Cmd
+     [],                                           % KeySpecs
+     [{node, [{shortname, "n"}, {longname, "node"},
+              {typecast, fun list_to_atom/1}]}],   % FlagSpecs
```
andrewjstone: Use: `riak_cli_typecast:to_node/1`. Yep, I realize this is all my fault.

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,288 @@
...
+            "riak-admin cluster partition index=22835963083295358096932575511191922182123945984\n\n",
+            "Returns the id or index for the specified index or id.\n"],
+    [["riak-admin", "cluster", "partition"], Text].
+
+partition([{index, Index}], []) when Index >= 0 ->
+    id_out(index, Index);
+partition([{id, Id}], []) when Id >= 0 ->
+    id_out(id, Id);
+partition([{Op, Value}], []) ->
+    [riak_cli_status:text(["ERROR: The given value ", integer_to_list(Value),
```
andrewjstone: So, I clearly need to write better docs as I've learned the last couple days. Errors should actually be wrapped in alerts. That way, eventually, we can bold or make them red, etc... We also can build separate error csv files.

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,288 @@
...
+          io_lib:format("Partition index: ~p -> id: ~p~n\t(owner: ~p)", 
+              [Index, hash_to_partition_id(Index, RingSize), Owner]));
+id_out1(id, Id, Ring, RingSize) when Id < RingSize ->
+    Idx = partition_id_to_hash(Id, RingSize),
+    Owner = riak_core_ring:index_owner(Ring, Idx),
+    riak_cli_status:text(
+          io_lib:format("Partition id: ~p -> index: ~p~n\t(owner: ~p)~n", 
+              [Id, partition_id_to_hash(Id, RingSize), Owner]));
+id_out1(id, Id, _Ring, _RingSize) ->
+    riak_cli_status:text(["ERROR: Id ", integer_to_list(Id), " is invalid."]).
```
andrewjstone: Same as above, wrap this in an alert please.

src/riak_core_cluster_cli.erl
```diff
@@ -0,0 +1,288 @@
...
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    RingSize = riak_core_ring:num_partitions(Ring),
+    [id_out1(InputType, Number, Ring, RingSize)].
+
+%% XXX FIXME: Validation of id - test for boundary condition and that it's lower
+%% than the ringsize - 1 partition id
+id_out1(index, Index, Ring, RingSize) ->
+    Owner = riak_core_ring:index_owner(Ring, Index),
+    riak_cli_status:text(
+          io_lib:format("Partition index: ~p -> id: ~p~n\t(owner: ~p)",
```
andrewjstone: I don't really like the `\t` here. I prefer things left-aligned. Even if you put the id on the next line with the owner it would work left-aligned. I will open this bikeshed up to @macintux @aberghage.

Currently output looks like this:

```
Andrews-MacBook-Pro:riak_ee ajs$ dev/dev1/bin/riak-admin cluster partition index=867766597165223607683437869425293042920709947392
Partition index: 867766597165223607683437869425293042920709947392 -> id: 38
    (owner: 'dev2@127.0.0.1')
```

andrewjstone: We should do the same thing with partition id=XX also.

jadeallenx:
@andrewjstone think 3e28706 is a winner :)

andrewjstone:
I'm seeing the following behavior, but I believe it should be fixed in riak_cli. In this case `["riak-admin", "cluster", "partition"]` has both a command and usage point. If riak_cli knew that the command required more than 1 argument it could automatically show usage if no arguments were given. This would require each user to change their specs to inform riak_cli of the number of required arguments. Alternatively, we could always show the usage if a client returns `{error, usage}` or `{error, usage, Status}`. That's probably a more robust solution for cases like these.

```
Andrews-MacBook-Pro:riak_ee ajs$ dev/dev1/bin/riak-admin cluster partition
RPC to 'dev1@127.0.0.1' failed: {'EXIT',
                                 {function_clause,
                                  [{riak_core_cluster_cli,partition,
                                    [[],[]],
                                    [{file,"src/riak_core_cluster_cli.erl"},
                                     {line,218}]},
                                   {riak_cli,run,1,
                                    [{file,"src/riak_cli.erl"},{line,70}]},
Andrews-MacBook-Pro:riak_ee ajs$ dev/dev1/bin/riak-admin cluster partition -h
Usage: riak-admin cluster partition id=0
riak-admin cluster partition index=22835963083295358096932575511191922182123945984

Returns the id or index for the specified index or id.
```

andrewjstone:
:+1:  3e28706

Great work @mrallen1

andrewjstone:
@borshop merge

PR #660, opened 2014-12-05
Feature/riak cli handoff enable

nickelization:
This adds CLI commands to enable/disable handoffs:

riak-admin <enable | disable> <inbound | outbound | both> [--node <NODE>] [--all]

If handoffs are disabled, any currently active handoffs in the specified direction will be terminated.

If the riak-admin "set" command is used to disable handoffs in a given direction via the newly exposed handoff.disable_inbound or handoff.disable_outbound settings, corresponding handoffs will also be automatically terminated.

This PR also includes some additional unit testing in riak_core_handoff_manager.

### DIFF
--- a/src/riak_core_handoff_cli.erl
+++ b/src/riak_core_handoff_cli.erl
@@ -1,0 +1,129 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2014 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+-module(riak_core_handoff_cli).
+
+-behavior(riak_cli_handler).
+
+-export([register_cli/0]).
+
+-spec register_cli() -> ok.
+register_cli() ->
+    register_cli_usage(),
+    register_cli_cfg(),
+    register_cli_cmds().
+
+register_cli_cmds() ->
+    CmdList = [handoff_cmd_spec(EnOrDis, Dir) ||
+               EnOrDis <- [enable, disable],
+               Dir     <- [inbound, outbound, both]],
+    lists:foreach(fun(Args) -> apply(riak_cli, register_command, Args) end,
+                  CmdList).
+
+register_cli_cfg() ->
+    lists:foreach(fun(K) ->
+                          riak_cli:register_config(K, fun handoff_cfg_change_callback/3)
+                  end, [["handoff", "disable_inbound"], ["handoff", "disable_outbound"]]).
+
+register_cli_usage() ->
+    riak_cli:register_usage(["riak-admin", "handoff"], handoff_usage()),
+    riak_cli:register_usage(["riak-admin", "handoff", "enable"], handoff_enable_disable_usage()),
+    riak_cli:register_usage(["riak-admin", "handoff", "disable"], handoff_enable_disable_usage()).
+
+handoff_usage() ->
+    ["riak-admin handoff <subcommand> [args]\n\n",
+     "Currently implemented handoff commands are:\n",
+     "  enable   Enable handoffs for the specified node(s)\n",
+     "  disable  Disable handoffs for the specified node(s)\n"
+    ].
+
+handoff_enable_disable_usage() ->
+    ["riak-admin handoff <enable | disable> <inbound | outbound | both> ",
+     "[[--node | -n] <Node>] [--all]\n\n",
+     "  Enable or disable handoffs on the specified node(s).\n",
+     "  If handoffs are disabled in a direction, any currently\n",
+     "  running handoffs in that direction will be terminated.\n\n"
+     "Options\n",
+     "  -n <Node>, --node <Node>\n",
+     "      Modify the setting on the specified node (default: local node only)\n",
+     "  -a, --all\n",
+     "      Modify the setting on every node in the cluster\n"
+    ].
+
+handoff_cmd_spec(EnOrDis, Direction) ->
+    Cmd = ["riak-admin", "handoff", atom_to_list(EnOrDis), atom_to_list(Direction)],
+    Callback = fun([], Flags) ->
+                       handoff_change_enabled_setting(EnOrDis, Direction, Flags)
+               end,
+    [
+     Cmd,
+     [], % KeySpecs
+     [{all, [{shortname, "a"},
+             {longname, "all"}]},
+      {node, [{shortname, "n"},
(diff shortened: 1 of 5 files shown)


src/riak_core_handoff_cli.erl
```diff
@@ -0,0 +1,98 @@
...
+                                  60000),
+
+    case Down of
+        [] ->
+            ok;
+        _ ->
+            io:format("handoff ~s failed on nodes: ~p~n", [EnOrDis, Down])
+    end;
+handoff_change_enabled_setting(EnOrDis, Direction, [{node, NodeStr}]) ->
+    %% FIXME switch to using crash-proof riak_cli utility function?
```
andrewjstone: Yep, you should use [riak_cli_typecast:to_node/1](https://github.com/basho/riak_cli/blob/develop/src/riak_cli_typecast.erl#L8)

andrewjstone:
@nm Let's try to keep the lines to 80 characters.

src/riak_core_handoff_cli.erl
```diff
@@ -0,0 +1,98 @@
...
+
+handoff_usage() ->
+    ["usage: riak-admin handoff <subcommand> [args]\n\n",
+     "Currently implemented handoff commands are:\n",
+     "  enable   Enable handoffs for the specified node(s)\n",
+     "  disable  Disable handoffs for the specified node(s)\n"
+    ].
+
+handoff_enable_disable_usage() ->
+    ["usage: riak-admin handoff <enable | disable> <inbound | outbound | both> [[--node | -n] <Node>] [--all]\n\n",
```
andrewjstone: Remove the "usage:" at the beginning. it's already added by riak_cli. Although, it may be better to remove it from riak-cli in the future. not sure.

src/riak_core_handoff_cli.erl
```diff
@@ -0,0 +1,98 @@
...
+                          riak_cli:register_config(K, fun handoff_cfg_change_callback/3)
+                  end, [["handoff", "disable_inbound"], ["handoff", "disable_outbound"]]).
+
+register_cli_usage() ->
+    riak_cli:register_usage(["riak-admin", "handoff"], handoff_usage()),
+    riak_cli:register_usage(["riak-admin", "handoff", "enable"], handoff_enable_disable_usage()),
+    riak_cli:register_usage(["riak-admin", "handoff", "disable"], handoff_enable_disable_usage()).
+
+handoff_usage() ->
+    ["usage: riak-admin handoff <subcommand> [args]\n\n",
```
andrewjstone: remove "usage:"

src/riak_core_handoff_cli.erl
```diff
@@ -0,0 +1,98 @@
...
+              {longname, "node"}]}], % FlagSpecs
+     Callback
+    ].
+
+handoff_change_enabled_setting(_EnOrDis, _Direction, Flags) when length(Flags) > 1 ->
+    io:format("Can't specify both --all and --node flags~n", []);
+handoff_change_enabled_setting(EnOrDis, Direction, [{all, _}]) ->
+    Nodes = riak_cli_nodes:nodes(),
+    {_, Down} = rpc:multicall(Nodes,
+                                  riak_core_handoff_manager,
```
andrewjstone: Can you line these lines up with Nodes please?

src/riak_core_handoff_cli.erl
```diff
@@ -0,0 +1,98 @@
...
+     "Options\n",
+     "  -n <Node>, --node <Node>\n",
+     "      Modify the setting on the specified node (default: local node only)\n",
+     "  -a, --all\n",
+     "      Modify the setting on every node in the cluster\n"
+    ].
+
+handoff_cmd_spec(EnOrDis, Direction) ->
+    Cmd = ["riak-admin", "handoff", atom_to_list(EnOrDis), atom_to_list(Direction)],
+    Callback = fun([], Flags) ->
```
andrewjstone, line 72: This callback must return a `riak_cli_status:status()`. Use a `[riak_cli_status:text/1]` status instead of printing directly with `io:format`.

Yes, I realize this changed after you submitted this PR. Moving targets are hard.  :dash:

andrewjstone, line 72: Eventually we'll do the same with config callbacks, but now it won't affect this PR ;)

src/riak_core_handoff_manager.erl
```diff
@@ -639,4 +688,53 @@ simple_handoff () ->
...
+                               {'DOWN', Ref, process, Pid, max_concurrency} ->
+                                   ok;
+                               Other ->
+                                   {error, unexpected_message, Other}
+                           after
+                               1000 ->
+                                   {error, timeout_waiting_for_down_msg}
+                           end
+                   end,
+
```
andrewjstone, line 717: I'm hesitant to introduce this test into riak_core. We shouldn't allow any spurious failures if we know they exist. It took a long time to get rid of the existing ones.

What you can do instead is use a wait-for pattern. We do this a lot in riak_tests. You'd do something like this: https://github.com/basho/riak_test/blob/master/src/rt.erl#L624

Then call it with a function that returns a boolean like so:

``` erlang
Fun = fun() ->  0 =:= length(status()) end,
?assertEqual(ok, wait_until(Fun, 30, 1000)).
```

andrewjstone:
Seeing a couple dialyzer warnings:

```
riak_core_cli_registry.erl:31: Invalid type specification for function riak_core_cli_registry:register_node_finder/0. The success typing is () -> 'true'
riak_core_cluster_cli.erl:208: Call to missing or unexported function riak_core_status:partitions/2
```

jadeallenx:
@andrewjstone Those dialyzer warnings are because I think @nickelization is missing a commit to export that function, but it's in my feature branch.

src/riak_core_handoff_cli.erl
```diff
@@ -0,0 +1,107 @@
...
+    end;
+handoff_change_enabled_setting(EnOrDis, Direction, [{node, NodeStr}]) ->
+    Node = riak_cli_typecast:to_node(NodeStr),
+    Result = rpc:call(Node, riak_core_handoff_manager, handoff_change_enabled_setting,
+             [EnOrDis, Direction], 60000),
+    case Result of
+        {badrpc, Reason} ->
+            Output = io_lib:format("Failed to update handoff settings on node ~p. Reason: ~p",
+                                   [Node, Reason]),
+            [riak_cli_status:text(Output)];
```
andrewjstone: This should really be wrapped in an alert for future formatting purposes.

src/riak_core_handoff_cli.erl
```diff
@@ -0,0 +1,108 @@
...
+    case Down of
+        [] ->
+            [riak_cli_status:text("All nodes successfully updated")];
+        _ ->
+            Output = io_lib:format("Handoff ~s failed on nodes: ~p", [EnOrDis, Down]),
+            [riak_cli_status:alert([riak_cli_status:text(Output)])]
+    end;
+handoff_change_enabled_setting(EnOrDis, Direction, [{node, NodeStr}]) ->
+    Node = riak_cli_typecast:to_node(NodeStr),
+    Result = rpc:call(Node, riak_core_handoff_manager, handoff_change_enabled_setting,
```
andrewjstone: I missed this one earlier. You should use `riak_cli_nodes:safe_rpc/4` instead.

src/riak_core_handoff_cli.erl
```diff
@@ -0,0 +1,108 @@
+-module(riak_core_handoff_cli).
```
andrewjstone, line 21: Should add a copyright header in here.

andrewjstone:
:+1: 06701ea
Nice work @nickelization

PR #661, opened 2014-12-08
Feature/revised/riak cli handoff status 1239

JeetKunDoug:
Implement "riak-admin handoff summary" and "riak-admin handoff details [--node=<node>] [--all]"

### DIFF
--- a/src/riak_core_handoff_status.erl
+++ b/src/riak_core_handoff_status.erl
@@ -1,0 +1,468 @@
+%% -------------------------------------------------------------------
+%%
+%% Copyright (c) 2014 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+-module(riak_core_handoff_status).
+
+-include("riak_core_handoff.hrl").
+
+-export([handoff_summary/2,
+         handoff_details/2,
+         register_cli/0,
+         collect_node_transfer_data/0]).
+
+-define(ORDERED_TRANSFERS_FOR_DISPLAY,
+        [{ownership, "Ownership"}, {resize, "Resize"}, {hinted, "Hinted"}, {repair, "Repair"}]).
+
+%%%%%%
+%% riak_cli registration
+
+register_cli() ->
+    register_all_commands(),
+    register_all_usage().
+
+register_all_commands() ->
+    lists:foreach(fun(Args) -> apply(riak_cli, register_command, Args) end,
+                  [register_summary_command(), register_details_command()]).
+
+register_summary_command() ->
+    [
+     ["riak-admin", "handoff", "summary"], %% Cmd
+     [],                                   %% KeySpecs
+     [],                                   %% Flags
+     fun handoff_summary/2
+    ].
+
+register_details_command() ->
+    [
+     ["riak-admin", "handoff", "details"],                %% Cmd
+     [],                                                  %% KeySpecs
+     [
+      {node, [{shortname, "n"}, {longname, "node"},       %% Flags
+              {typecast, fun riak_cli_typecast:to_node/1}]},
+      {all, [{shortname, "a"}, {longname, "all"}]}
+     ],
+     fun handoff_details/2
+    ].
+
+register_all_usage() ->
+    lists:foreach(fun(Args) -> apply(riak_cli, register_usage, Args) end,
+                  [register_summary_usage(),
+                   register_details_usage(),
+                   register_handoff_usage()]).
+
+register_handoff_usage() ->
+    [
+     ["riak-admin", "handoff"],
+     [
+      "riak-admin handoff <sub-command>\n\n",
+      "  View handoff related status\n\n",
+      "  Sub-commands:\n",
+      "    summary    Show cluster-wide handoff summary\n",
+      "    details    Show details of all active transfers (per-node or cluster wide)."
+     ]
+    ].
+
+
+register_summary_usage() ->
+    [
+     ["riak-admin", "handoff", "summary"],
+     [
+      "riak-admin handoff summary\n\n",
+      "Displays a summarized view of handoffs occuring on the \n",
+      "current node (no flags), another node (using --node), or \n",
+      "the entire cluster (using --all)\n"
+     ]
+    ].
+
+register_details_usage() ->
+    [
+     ["riak-admin", "handoff", "details"],
+     [
+      "riak-admin handoff details [--node node] [--all]\n\n",
+      "Displays a summarized view of handoffs occuring on the \n",
+      "current node (no flags), another node (using --node), or \n",
+      "the entire cluster (using --all)\n"
+     ]
+    ].
+
+%% end riak_cli registration
+%%%%%%
+
+%%%%%%
+%% riak_cli callbacks
+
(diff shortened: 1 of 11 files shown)


src/riak_core_handoff_status.erl
```diff
@@ -0,0 +1,460 @@
...
+correlate_active_transfers(Transfers) ->
+    {Inbound, Outbound} = lists:partition(fun(T) -> transfer_info(T, direction) =:= inbound end, Transfers),
+    InboundWithType = populate_inbound_types(Inbound, Outbound),
+    Outbound ++ InboundWithType.
+
+
+populate_inbound_types(Inbound, Outbound) ->
+    [case find_matching_outbound(Transfer, Outbound) of
+         [] -> Transfer;
+         [Match | _] ->
```
nickelization: Throwing away extra list entries seems potentially suspect to me. If find_matching_outbound should only ever return one item, then we should only match on one item. If not, can we safely ignore extra entries? I'm wondering whether we might potentially end up showing the wrong inbound transfer type here. What if there are multiple outbound transfers that have the same target_node/target_partition/mod, but different types? (Not sure what mod means though, so maybe that rules out differing types?)

JeetKunDoug: I think I did that because I had once seen a literal duplicate in active transfers somewhere, but perhaps I was hallucinating, as I wasn't crazy about it either. I'll try to remove it and do some testing to see what's up, but you really should never have a duplicate active transfer from what I can tell.

src/riak_core_handoff_status.erl
```diff
@@ -0,0 +1,451 @@
...
+
+    KnownStats = count_known_transfers(Known),
+    ActiveStats = count_active_transfers(Active),
+
+    %% Each stats structure is an orddict with
+    %% "Node:TransferType:Direction" and "TransferType" keys, values
+    %% are integers. Also ActiveStats will have a `nodes' key with a
+    %% ordset of nodes.
+
+    Header = riak_cli_status:text("Each cell indicates active transfers and, in parenthesis, the number of all known transfers. The 'Total' column is the sum of the active transfers."),
```
nickelization: Shorten this line, perhaps?

src/riak_core_vnode_manager.erl
```diff
@@ -555,7 +580,7 @@ idx2vnode(Idx, Mod, _State=#state{idxtab=T}) ->
 delmon(MonRef, _State=#state{idxtab=T}) ->
     case ets:lookup(T, MonRef) of
         [#monrec{key= {Index, Mod} = Key}] ->
-	    _ = unregister_vnode_stats(Mod, Index),
+            _ = unregister_vnode_stats(Mod, Index),
```
nickelization, line 582: Why the "_ = "?

JeetKunDoug, line 582: No idea - that was actually in the code long before we touched it, but emacs reformatted some things and it ended up moving that line.
Generally, I'm guessing it's to be explicit about "I know there's a return value, but I don't care" but I also don't know if that's idiomatic Erlang (or Riak Erlang, or what).

src/riak_core_handoff_status.erl
```diff
@@ -0,0 +1,453 @@
...
+
+-spec handoff_details([tuple()], [tuple()]) -> riak_cli_status:status().
+handoff_details([], []) ->
+    build_handoff_details(node());
+handoff_details([], [{node, Node}]) ->
+    build_handoff_details(Node);
+handoff_details([], [{all, _Value}]) ->
+    build_handoff_details(all);
+handoff_details([], _) ->
+    [riak_cli_status:alert([riak_cli_status:text("Cannot use both --all and --node flags at the same time.")])].
```
jadeallenx, line 47: Question for @andrewjstone should there be a way in the flag specs to say that flags x and y are mutually exclusive? Future enhancement for riak_cli perhaps?

src/riak_core_handoff_status.erl
```diff
@@ -0,0 +1,453 @@
...
+            end
+    end.
+
+row_details(T, RingSize) ->
+    SourceNode = transfer_info(T, src_node),
+    TargetNode = transfer_info(T, target_node),
+    SourceIdx = transfer_info(T, src_partition),
+    TargetIdx = transfer_info(T, target_partition),
+    Source = io_lib:format("~ts:~B", [SourceNode, riak_core_ring_util:hash_to_partition_id(SourceIdx, RingSize)]),
+    Target = io_lib:format("~ts:~B", [TargetNode, riak_core_ring_util:hash_to_partition_id(TargetIdx, RingSize)]),
```
jadeallenx, line 103: Would be more readable if you put riak_core_ring_util:hash_to_partition_id/2 behind a local function, IMO.

jadeallenx:
Great work @JeetKunDoug Thanks.

src/riak_core_handoff_status.erl
andrewjstone, line 47: Yep @mrallen1 , I just opened an issue for this: https://github.com/basho/riak_cli/issues/22

nickelization:
:+1:

src/riak_core_handoff_status.erl
```diff
@@ -0,0 +1,468 @@
...
+    case lists:keyfind(Key, 1, Status) of
+        {Key, Value} -> Value;
+        _ -> " "
+    end.
+
+transfer_stat(Transfer, Key) ->
+    case transfer_info(Transfer, stats) of
+        no_stats -> unknown;
+        Stats ->
+            {Key, Value} = lists:keyfind(Key, 1, Stats),
```
andrewjstone, line 125: Will this ever return false?

src/riak_core_handoff_status.erl
```diff
@@ -0,0 +1,468 @@
...
+                             ({_Node, {{_Module, _Index},
+                                       {resize, _Dir, _Data}}}) ->
+                                  false;
+                             (_) ->
+                                  true
+                          end,
+                          Known),
+    coalesce_known_transfers(Known1, RingTransfers).
+
+%% Transform a single tuple with a list of modules (from ring.next) to
```
andrewjstone: Comment isn't really accurate anymore. better to just put a type spec.

andrewjstone: or just change the comment I suppose.

src/riak_core_handoff_status.erl
JeetKunDoug, line 125: If it does, it should blow up - the list of stats is hard-coded in riak_core_handoff_manager:calc_stats and is always the same.

src/riak_core_handoff_status.erl
```diff
@@ -0,0 +1,468 @@
...
+     fun handoff_details/2
+    ].
+
+register_all_usage() ->
+    lists:foreach(fun(Args) -> apply(riak_cli, register_usage, Args) end,
+                  [register_summary_usage(),
+                   register_details_usage(),
+                   register_handoff_usage()]).
+
+register_handoff_usage() ->
```
andrewjstone: So when we merge this will have to merge with riak_core_handoff_cli which does the same thing

andrewjstone: One of them will win now depending upon order of registration.

src/riak_core_handoff_status.erl
```diff
@@ -0,0 +1,468 @@
...
+      "current node (no flags), another node (using --node), or \n",
+      "the entire cluster (using --all)\n"
+     ]
+    ].
+
+register_details_usage() ->
+    [
+     ["riak-admin", "handoff", "details"],
+     [
+      "riak-admin handoff details [--node node] [--all]\n\n",
```
andrewjstone: This is the same as the summary description. Not sure if intended.

src/riak_core_handoff_status.erl
JeetKunDoug: Doing this now, as I've rebased on top of integration, which has riak_core_handoff_cli.

src/riak_core_handoff_status.erl
JeetKunDoug: Nope - correcting.

src/riak_core_handoff_status.erl
JeetKunDoug: rewrote per our quick pairing session.

andrewjstone:
:+1:  8fc16f9

JeetKunDoug:
@borshop merge

PR #664, opened 2014-12-10
Handle down nodes deterministically

macintux:
For a cluster of > 3 nodes, if a node was administratively marked down, it was effectively impossible to commit a `cluster leave` action because the `random` library was used to determine how to cope with the down node, but the claimant's seed value wasn't used inside `handle_down_nodes/2`.

Thanks to @nickelization for the assistance in chasing this down.

One perhaps irrelevant observation: we're now using the same seed twice for `random:uniform_s/2` invocations (also used in `riak_core_gossip:attempt_simple_transfer/7`).

@jtuple, per our discussion last night, here's the fix. Interested in opinions on what type of testing we should add for this.

### DIFF
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -1308,7 +1308,7 @@ update_ring(CNode, CState, Replacing, Seed, Log, false) ->
                [length(riak_core_ring:pending_changes(CState4))]}),
 
     %% Remove transfers to/from down nodes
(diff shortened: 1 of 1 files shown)


macintux:
Noteworthy: `riak-admin leave -f` on the node in question is an effective workaround for the problem.

Basho-JIRA:
This has not yet been reviewed and merged 

**_[posted via JIRA by John Daily]**_

PR #667, opened 2014-12-12
Track minimal set of ring transition info in cluster metadata

macintux:
Today, we have no way of reporting on the completion percentage of a set of ring ownership transfers because the ring `next` field is winnowed down as each transfer completes.

This PR will update cluster metadata each time a ring ownership/resize transition is initiated to reflect the total number of transfers required and the new ring version. It is effectively experimental and deserves some tests.


macintux:
This PR is past its due by date, and I don't want to give riak_core food poisoning. Closing it without merging.

andrewjstone:
@macintux Are we ever going to add this code, or is it unnecessary? Trying to wrangle the corresponding jira issue.

macintux:
Sorry, was a busy day.

I'd like to take another shot at the problem if we're going to think about moving the ring to CMD (probably a good topic for next week). I wasn't all that happy with this code and there didn't seem to be a burning need for the changes, which is why I closed it out originally.

andrewjstone:
Cool. Sounds good. Thanks.

PR #669, opened 2014-12-12
Add transfer_limit config callback to handoff_cli

andrewjstone:
This ensures that when the limit is reduced processes exceeding the
limit get shut down. Uses already existing functionality in
riak_core_handoff_manager for the shutdown.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -16,6 +16,5 @@
   {pbkdf2, ".*", {git, "git://github.com/basho/erlang-pbkdf2.git", {tag, "2.0.0"}}},
   {eleveldb, ".*", {git, "git://github.com/basho/eleveldb.git", {branch, "2.0"}}},
   {exometer_core, ".*", {git, "git://github.com/basho/exometer_core.git", {tag, "1.0.0-basho1"}}},
-  {getopt, ".*", {git, "git://github.com/jcomellas/getopt.git", {tag, "626698975e63866156159661d100785d65eab6f9"}}},
   {riak_cli, ".*", {git, "git@github.com:basho/riak_cli.git", {branch, "master"}}} % XXX FIXME - should point to a real tag once released
 ]}.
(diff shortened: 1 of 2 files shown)


andrewjstone:
cc @macintux

macintux:
Force pushed a rebase onto an also-rebased `integration/riak-admin-handoff-team` branch

macintux:
The 'help enable' output to be underspecified compared to 'enable help'. Specifically, the `[args]` are unknown until you try something, which seems risky; as a sysadmin, I'd really rather not try a command to see what happens.

```
Thunder:bin John$ ./riak-admin handoff help enable
Usage: riak-admin handoff <subcommand> [args]

Currently implemented handoff commands are:
  enable   Enable handoffs for the specified node(s)
  disable  Disable handoffs for the specified node(s)

Thunder:bin John$ ./riak-admin handoff enable help
Usage: riak-admin handoff <enable | disable> <inbound | outbound | both> [[--node | -n] <Node>] [--all]

  Enable or disable handoffs on the specified node(s).
  If handoffs are disabled in a direction, any currently
  running handoffs in that direction will be terminated.

Options
  -n <Node>, --node <Node>
      Modify the setting on the specified node (default: local node only)
  -a, --all
      Modify the setting on every node in the cluster
```

(This is outside the scope of the current PR)

Update: actually, I don't think it is. Notice that also the "enable inbound" syntax is not illustrated via "help enable".

macintux:
Would really like to see more context-sensitive output. What exactly did I do?

```
Thunder:bin John$ ./riak-admin handoff disable inbound
Handoff setting successfully updated
Thunder:bin John$ ./riak-admin handoff enable inbound
Handoff setting successfully updated
```

E.g., is that for this node? Cluster-wide? Did I just enable or disable?

macintux:
```
$ ./riak-admin handoff disable outbound
Handoff setting successfully updated
```

Yet, joining another node to this one works and all transfers occur. Hrm.

macintux:
Ok, now, focusing on THIS particular PR, not a big fan of this behavior:

```
Thunder:dev1 John$ ./bin/riak-admin transfer_limit dev3@127.0.0.1
Invalid limit: dev3@127.0.0.1
Thunder:dev1 John$ ./bin/riak-admin transfer_limit help
Invalid limit: help
```

andrewjstone:
@macintux weird. There should have to be a `set`, `show`, or `describe` in there for that to work at all. I just reproduced locally as well. Will fix.

andrewjstone:
@macintux I just confirmed that this behavior is a result of transfer_limit being a valid 'legacy' command.  We can't change it without breaking backwards compatibility. 
https://github.com/basho/riak_ee/blob/develop/rel/files/riak-admin#L831

PR #676, opened 2014-12-16
Add clique_status:usage() return for case when we get neither Index or ID

JeetKunDoug:
Requires https://github.com/basho/clique/pull/25. Addresses https://github.com/basho/riak_core/pull/659#issuecomment-66381007


jadeallenx:
:+1:

andrewjstone:
rebar.config actually needs to change also so that it includes a later version of clique

andrewjstone:
@JeetKunDoug I hope you don't mind, but I made the change to rebar.config and pushed it. It all works so +1.

andrewjstone:
:+1:  8424513

JeetKunDoug:
@andrewjstone thanks for wrapping this up - I will never mind someone finishing up my work for me :)

PR #678, opened 2014-12-17
Integration/riak admin handoff team

andrewjstone:
Should be merged together with corresponding riak and riak_ee PRs for 2.0.4 release.

### DIFF
--- a/docs/ring-resizing.md
+++ b/docs/ring-resizing.md
@@ -243,11 +243,15 @@ isn't implicitly enlarged during the transition, it is not used during expansion
 One other minor change made to the vnode interface, regardless of
 whether or not resizing is being used, is the first argument
 of `handoff_starting/2` is now passed as `{HOType, {Idx, Node}}`.
-`HOType` is one of `ownership`, `hinted`, `resize`, or `repair`.
+`HOType` is one of `ownership`, `hinted`, or `resize`.
 The `Idx` being transferred to is always the index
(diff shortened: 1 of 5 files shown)


priv/riak_core.schema
```diff
@@ -95,6 +95,24 @@
   hidden
 ]}.
 
+%% @doc Disables outbound handoff transfers for this node. If you
+%% turn this setting on at runtime with riak-admin, it will kill any
+%% outbound handoffs currently running.
+{mapping, "handoff.disable_outbound", "riak_core.disable_outbound_handoff", [
```
seancribbs: Pretty sure @angrycub will comment on this, but in other areas we tended not to have config keys that include the state of the thing in the name, especially when the type is a flag.  How about "handoff.outbound" -> on/off? It's easy to invert the flag, for example:

``` erlang
{mapping, "handoff.outbound", "riak_core.disable_outbound_handoff", [
  {datatype, {enum, [{on, false}, {off, true}]}},
  {default, on},
  hidden
]}.
```

angrycub: Agreed.  Double negatives are a very sad thing.

andrewjstone: I agree as well. @nickelization I think we should use @seancribbs idea.

andrewjstone:
:+1:  06a20ad

andrewjstone:
@borshop merge

PR #680, opened 2014-12-17
Make riak_core_status:active_partitions safe if a node is down.

JeetKunDoug:
Previously, would crash. Now returns empty ordset instead so riak_core_cluster_cli:partitions works and shows stopped partitions correctly.


jadeallenx:
:+1:

JeetKunDoug:
:+1: 6450799

jadeallenx:
@borshop merge

PR #682, opened 2014-12-17
add riak-admin handoff config support

andrewjstone:


macintux:
There are other handoff values; do we want to constrain the `config` command to only show those that can be adjusted, or should we represent all of them?

``` erlang
%% @doc handoff.port is the TCP port that Riak uses for
%% intra-cluster data handoff.
{mapping, "handoff.port", "riak_core.handoff_port", [
  {default, {{handoff_port}} },
  {datatype, integer},
  hidden
]}.

%% @doc To encrypt riak_core intra-cluster data handoff traffic,
%% uncomment the following line and edit its path to an appropriate
%% certfile and keyfile.  (This example uses a single file with both
%% items concatenated together.)
{mapping, "handoff.ssl.certfile", "riak_core.handoff_ssl_options.certfile", [
%%  {commented, "/tmp/erlserver.pem"},
  {datatype, file},
  hidden
]}.

%% @doc if you need a seperate keyfile for handoff
{mapping, "handoff.ssl.keyfile", "riak_core.handoff_ssl_options.keyfile", [
  {datatype, file},
  hidden
]}.
```

andrewjstone:
@macintux You are correct. I forgot about those. I think we should include them. They are visible and settable. I will update this PR.

It leads to another point though, as this list grows they are no longer going to fit in a single row of a table.
It's almost like we should list each one on it's own line and separate groups of variables by node.

andrewjstone:
As discussed with @macintux We are omitting displaying the ssl related files due to other issues.

andrewjstone:
@macintux Ready again for review.

andrewjstone:
@macintux This [PR](https://github.com/basho/clique/pull/29) fixes the issue with out of order config variables in tables.

macintux:
@borshop merge

PR #685, opened 2014-12-18
Improve handoff enable/disable config naming

nickelization:
This avoids the confusing double-negative situation where you have to turn off the disabling of handoff in order to enable it. Now you just say e.g. "set handoff.inbound=on" to enable inbound handoff, and "set handoff.inbound=off" to disable it.

### DIFF
--- a/priv/riak_core.schema
+++ b/priv/riak_core.schema
@@ -95,21 +95,21 @@
   hidden
 ]}.
 
-%% @doc Disables outbound handoff transfers for this node. If you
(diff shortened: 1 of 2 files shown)


macintux:
Reviewing

andrewjstone:
@borshop retry

andrewjstone:
:+1: 2e2feee

andrewjstone:
Rebased to deal with the whitelisting merge.

PR #691, opened 2014-12-29
Allow schema loading with an environment variable

andrewjstone:
Allow tests to use alternate schema paths via an environment variable in
riak_core.

Bump Clique to 0.2.4 so that schema loading is done explicitly by the
caller and not automatically during application startup.


andrewjstone:
bors is f-d. closing.

PR #696, opened 2015-01-22
Bump clique to 0.2.5

macintux:
`rebar.config` update

### DIFF
--- a/src/riak_core_handoff_cli.erl
+++ b/src/riak_core_handoff_cli.erl
@@ -146,11 +146,11 @@ details_usage() ->
 handoff_config(_Args, Flags) when length(Flags) > 1 ->
     [clique_status:text("Can't specify both --all and --node flags")];
 handoff_config(_Args, []) ->
-    clique_config:show(config_vars());
-handoff_config(_Args, [{all, _}]) ->
-    clique_config:show(config_vars() ++ ["-a"]);
+    clique_config:show(config_vars(), []);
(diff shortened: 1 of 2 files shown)


macintux:
@gcymbalski ping

nickelization:
:+1: a1cca68e855c3285e0358c4a77464d2a2af36a4d

src/riak_core_handoff_cli.erl
```diff
@@ -146,11 +146,11 @@ details_usage() ->
 handoff_config(_Args, Flags) when length(Flags) > 1 ->
     [clique_status:text("Can't specify both --all and --node flags")];
 handoff_config(_Args, []) ->
-    clique_config:show(config_vars());
+    clique_config:show(config_vars(), []);
 handoff_config(_Args, [{all, _}]) ->
-    clique_config:show(config_vars() ++ ["-a"]);
+    clique_config:show(config_vars(), [{all, undefined}]);
```
jadeallenx: I wonder if it would be a good idea to pass through the value for `[{all, Arg}]` in the function head to clique_config:show/2 in case the value is ever anything other than `undefined`

nickelization: Ah, yes, fair point. Right now we don't support values for the --all flag, but it is theoretically possible we might in the future. I'll rebase in that change in just a moment...

jadeallenx:
:+1: 46317de

jadeallenx:
@borshop merge

Issue #619, opened 2014-08-04
CIDR badmatch error


macintux:
If the admin forgets the trailing slash/mask, things get ugly. Need to handle this gracefully.

```
$ ./riak-admin security add-source jdaily 127.0.0.1 trust
RPC to 'riak@127.0.0.1' failed: {'EXIT',
                                 {{badmatch,["127.0.0.1"]},
                                  [{riak_core_console,parse_cidr,1,
                                    [{file,"src/riak_core_console.erl"},
                                     {line,1141}]},
                                   {riak_core_console,add_source,1,
                                    [{file,"src/riak_core_console.erl"},
                                     {line,961}]},
                                   {rpc,'-handle_call_call/6-fun-0-',5,
                                    [{file,"rpc.erl"},{line,205}]}]}}
```

macintux:
Two options: report parsing errors politely, or treat an IP address without a netmask as /32.

Basho-JIRA:
Filed PR: https://github.com/basho/riak_core/pull/703 

**_[posted via JIRA by John Daily]**_

PR #700, opened 2015-02-10
Parse invalid CIDR values without throwing exceptions

macintux:
Addresses #619


macintux:
Note: I chose not to create tests because I'd like to let cuttlefish do this for us instead (see also basho/cuttlefish#182).

macintux:
And...shockingly (not) tests run from the Riak console worked but in deployment it...doesn't.

macintux:
Closing, will open a new PR when this works.

PR #712, opened 2015-02-25
export security context for use in type-specs in riak-kv wm modules

zeeshanlakhani:
Example of a module that contains a type that would need it: https://github.com/basho/riak_kv/pull/1083/files#diff-96bb962935952335397285e8c1c70683R59.


bearcage:
+1 47d4f9b

zeeshanlakhani:
@borshop merge

PR #716, opened 2015-03-04
Add riak_core_ring_manager:is_stable_ring/0 public API [JIRA: RIAK-1598]

darach:
The status of a ring as stable (no topology change in a promotion period) or not is of general utility.
This patch proposes to refactor the predicate check into a public API so that it can be made more
generally useful.

### DIFF
--- a/src/riak_core_ring_manager.erl
+++ b/src/riak_core_ring_manager.erl
@@ -639,7 +639,7 @@ is_stable_ring(#state{ring_changed_time=Then}) ->
     DeltaUS = erlang:max(0, timer:now_diff(os:timestamp(), Then)),
     DeltaMS = DeltaUS div 1000,
     IsStable = DeltaMS >= ?PROMOTE_TIMEOUT,
-    {IsStable,DeltaMS}.
+    {IsStable, DeltaMS}.
 
 %% ===================================================================
 %% Unit tests
--- a/src/riak_core_ring_manager.erl
+++ b/src/riak_core_ring_manager.erl
@@ -84,7 +84,8 @@
          ring_trans/2,
          run_fixups/3,
          set_cluster_name/1,
-         stop/0]).
+         stop/0,
+         is_stable_ring/0]).
 
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
         terminate/2, code_change/3]).
@@ -211,6 +212,9 @@ ring_trans(Fun, Args) ->
 
 set_cluster_name(Name) ->
     gen_server:call(?MODULE, {set_cluster_name, Name}, infinity).
+
+is_stable_ring() ->
+    gen_server:call(?MODULE, is_stable_ring, infinity).
 
 %% @doc Exposed for support/debug purposes. Forces the node to change its
 %%      ring in a manner that will trigger reconciliation on gossip.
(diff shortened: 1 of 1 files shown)


src/riak_core_ring_manager.erl
```diff
@@ -406,7 +410,10 @@ handle_call({ring_trans, Fun, Args}, _From, State=#state{raw_ring=Ring}) ->
 handle_call({set_cluster_name, Name}, _From, State=#state{raw_ring=Ring}) ->
     NewRing = riak_core_ring:set_cluster_name(Ring, Name),
     State2 = prune_write_notify_ring(NewRing, State),
-    {reply, ok, State2}.
+    {reply, ok, State2};
+handle_call(is_stable_ring, _From, State) ->
```
andrewjstone, line 414: Nitpick: Please use spaces after commas in tuples here and below for consistency with surrounding code in this file.

src/riak_core_ring_manager.erl
```diff
@@ -630,6 +635,12 @@ prune_write_ring(Ring, State) ->
     State2 = set_ring(Ring, State),
     State2.
 
+is_stable_ring(_State=#state{ring_changed_time=Then}) ->
```
andrewjstone: No need for _State here

src/riak_core_ring_manager.erl
```diff
@@ -721,5 +732,14 @@ do_write_ringfile_test() ->
     ?assertMatch({error,_}, do_write_ringfile(GenR(ring_perms), ?TEST_RINGFILE)),
     ok = file:change_mode(?TEST_RINGDIR, 8#00755).
 
--endif.
+is_stable_ring_test() ->
+    % This test is efferently coupled to the PROMOTE_TIMEOUT macro value
```
andrewjstone: Did you really mean efferently here?

src/riak_core_ring_manager.erl
```diff
@@ -721,5 +732,14 @@ do_write_ringfile_test() ->
...
+    {A,B,C} = Now = os:timestamp(),
+    Within = {A,B-45,C},
```
andrewjstone: This test will break if `?PROMOTE_TIMEOUT` is changed. Considering it's unlikely for the unit to be changed from ms, since that would require changing other code, I'd change this test so that it calculates the values in terms of `?PROMOTE_TIMEOUT`. Something like:

``` erlang
TimeoutSecs = ?PROMOTE_TIMEOUT div 1000,
Within = {A, B - (TimeoutSecs div 2), C},
Outside = {A, B - (TimeoutSecs + 1), C}
```

Technically this still has a chance of failing every 277 days if B is < ?TIMEOUT_SECS  or ?TIMEOUT_SECS/2, but I'm not really worried about that.

andrewjstone:
Hi @darach,
Thanks for the PR. If you make the minor changes requested we'll go ahead and get it merged in.

src/riak_core_ring_manager.erl
darach: Yes. Its a 'note to self' that changes to PROMOTE_TIMEOUT will change the boundaries of the chosen values in the test. From the point of view of a user this is surprising unless you read the code... You know, the comment works better without reference to efference at all anyway, so I'll remove...

darach:
Thank you for calling out the :space_invader:'s I didn't notice breaking with the surround style.
Your version of the test is better and clearly documents the dependency on ?PROMOTE_TIMEOUT, so adopted. Let me know if there's anything else I can change to more closely reflect the prevailing style and conventions.

andrewjstone:
:+1: 3666df2

Thanks for making those changes @darach

andrewjstone:
@borshop merge

darach:
Awesome! :two_hearts:

mbbroberg:
^ Love seeing these moments :+1:

PR #718, opened 2015-03-13
A touch more cleanup related to [RIAK-1506]

tburghart:
These are tweaks that were added in the course of analyzing how CNs were being evaluated, and are worth keeping.
Log the ACL when SSL peer name doesn't match it.
Add optimized invalid_cn_pair/2 patterns to avoid string case conversion.
Comments!

### DIFF
--- a/src/riak_core_ssl_util.erl
+++ b/src/riak_core_ssl_util.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_core: Core Riak Application
 %%
-%% Copyright (c) 2007-2013 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2015 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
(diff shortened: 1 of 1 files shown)


src/riak_core_ssl_util.erl
```diff
@@ -198,15 +209,16 @@ verify_ssl(Cert, valid_peer, {App, MyCommonName}) ->
...
             {fail, duplicate_common_name};
         _ ->
-            case validate_common_name(CommonName,
-                    app_helper:get_env(App, peer_common_name_acl, "*")) of
+            ACL = app_helper:get_env(App, peer_common_name_acl, "*"),
+            case validate_common_name(CommonName, ACL) of
                 {true, Filter} ->
-                    lager:info("SSL connection from ~s granted by ACL ~s",
+                    lager:info("SSL connection from ~s granted by ACL \"~s\"",
                         [CommonName, Filter]),
```
macintux: At a glance, looks like this should be `ACL`, not `Filter` (but haven't looked at `validate_common_name` yet).

macintux:
@borshop merge

PR #719, opened 2015-03-13
Add handle_tick callback to vnode

loucash:
In my riak_core based project I need a mechanism to periodically revisit state of my vnode.

**My first approach** was to send a tick using `erlang:send_after/3`, to handle a message in `MyVnode:handle_info/2` and to resend a tick.
The problem with this approach was that, because [`MyTickInterval < 2 * inactivity_timeout`](https://github.com/basho/riak_core/blob/develop/src/riak_core_vnode.erl#L245), vnode never had a chance to send [`inactive`](https://github.com/basho/riak_core/blob/develop/src/riak_core_vnode.erl#L463) event to `riak_core_vnode_manager`, thus manager never had a chance to [trigger handoff](https://github.com/basho/riak_core/blob/develop/src/riak_core_vnode_manager.erl#L468). Because handoff was blocked, one behaviour that I observed was node hanging during cluster leave `riak_core:leave()`.

**My second approach** was that I will add another callback to vnode `handoff_ready/1`, so vnode could cancel a tick and pretend to be "inactive". The best place I could find to do that was [`riak_core_vnode_manager:update_handoff/4`](https://github.com/basho/riak_core/blob/develop/src/riak_core_vnode_manager.erl#L710). Unfortunately this function is called periodically when manager receives [`management_tick`](https://github.com/basho/riak_core/blob/develop/src/riak_core_vnode_manager.erl#L433). So I went back to square one, as calling `handoff_ready` callback was causing cancellation of inactivity timer in vnode.

**My third approach** (presented in this PR):
- `gen_fsm` builtin timeout mechanism is replaced by manual calls of `riak_core_vnode:start_inactivity_timer/1,2` and `riak_core_vnode:stop_inactivity_timer/1`;
- `MyVnode:init/2` can return new option, which is `{tick, TickInterval::non_neg_integer()}`
- when `tick` option is present, then new callback: `MyVnode:handle_tick/1` also needs to be present. This callback is called every `TickInterval` ms.
- `MyVnode:handle_tick/1` callback can return three different values:
  - `{ok, NewState::term()}` - updates modstate
  - `{ok, NewState::term(), NewTickInterval::non_neg_integer()}` - updates modstate and `tick_interval`
  - `{cancel, NewState::term()}` - cancels tick

Because I do not have an access to quickcheck I could not run `test/core_vnode_eqc.erl` tests. I think that a minimum regression test would be to turn on tick in [`test/mock_vnode.erl`](https://gist.github.com/loucash/610d71f28550271bb4f0) and expect all cases to pass. Could you confirm that (so I could attach `test/mock_vnode.erl` changes)?

**Further improvements**:
- API to reinstate `tick` (after it was cancelled or it wasn't defined at `MyVnode:init/1' options)
- API to manually send `tick` message (if someone wants to ping vnode from outside)
- adjust `tick` message? `$tick`?

### DIFF
--- a/src/riak_core_vnode.erl
+++ b/src/riak_core_vnode.erl
@@ -63,10 +63,11 @@
         (R == normal orelse R == shutdown orelse
                                             (is_tuple(R) andalso element(1,R) == shutdown))).
 
--export_type([vnode_opt/0, pool_opt/0]).
+-export_type([vnode_opt/0, pool_opt/0, tick_opt/0]).
 
--type vnode_opt() :: pool_opt().
+-type vnode_opt() :: pool_opt() | tick_opt().
 -type pool_opt() :: {pool, WorkerModule::module(), PoolSize::pos_integer(), WorkerArgs::[term()]}.
+-type tick_opt() :: {tick, Interval::non_neg_integer()}.
 
 -callback init([partition()]) ->
     {ok, ModState::term()} |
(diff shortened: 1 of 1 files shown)


darach:
:+1: This would simplify some periodic vnode tasks in the riak_core based project i'm leading too.

PR #720, opened 2015-03-18
Be more tolerant of lager dependency version

edudev:
Currently riak_core enforces lager 2.0.3 and projects needing lager 2.1.0 have to go through workarounds.
The main benefit of the newer version is the support of maps, which riak_core currently doesn't use much, so there is little need in changing the source tag version.


seancribbs:
:-1: We are making a concerted effort to use only tagged versions of dependencies. I would prefer to simply bump the dependency to 2.1.1.

PR #724, opened 2015-03-25
Cache the regnames created in the vnode proxy in an ets table

seancribbs:
Identified as a hotspot in the write-once path for riak_kv. Even in the 
previously optimized version, calling reg_name often enough to identify vnode
proxy processes can consume more time per-call than an ets lookup on a public
table. This amortizes the cost by caching the computed atom in ets, making the
penalty on miss higher than the original computation but names will be
computed fewer times overall.


seancribbs:
@jonmeredith @fadushin Please see if this helps!

PR #725, opened 2015-03-30
Set dependency on eleveldb 2.1.0.0 tag.

fadushin:
This is to pull in the latest tagged version of eleveldb, which has the async_put operations for the write_once path.

Note that I left the rebar VsnRegex ".*".


fadushin:
@borshop merge

PR #730, opened 2015-04-30
Try very hard to consult a file before giving up

russelldb:
If a file is being read from a prior version and fails, set the
encoding to old default and try again before failing

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -62,12 +62,17 @@
          make_fold_req/4,
          make_newest_fold_req/1,
          proxy_spawn/1,
-         proxy/2
+         proxy/2,
+         consult/1
         ]).
 
 -include("riak_core_vnode.hrl").
 
 -ifdef(TEST).
+-ifdef(EQC).
+-export([prop_any_bin_consult/0]).
+-include_lib("eqc/include/eqc.hrl").
+-endif.
 -include_lib("eunit/include/eunit.hrl").
 -export([counter_loop/1,incr_counter/1,decr_counter/1]).
 -endif.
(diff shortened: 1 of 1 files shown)


src/riak_core_util.erl
```diff
@@ -150,6 +155,55 @@ replace_file(FN, Data) ->
             Err
     end.
 
+%% @doc Similar to {@link file:consult/1} but will consult files that
+%% are not valid unicode. Specifially files written with
```
seancribbs, line 159: "Specifically"

src/riak_core_util.erl
```diff
@@ -876,5 +930,16 @@ proxy_spawn_test() ->
         ok
     end.
 
+-ifdef(EQC).
+
+prop_any_bin_consult() ->
```
seancribbs, line 935: Can you wrap this in an eunit so it gets run in the normal suite?

src/riak_core_util.erl
```diff
@@ -150,6 +155,55 @@ replace_file(FN, Data) ->
...
+%% @doc Similar to {@link file:consult/1} but will consult files that
+%% are not valid unicode. Specifially files written with
+%% `io_lib:format("~p,", io_list())' in r16 before `utf8' became the
+%% default encoding.
+-spec consult(file:filename()) -> {ok, [term()]} | {error, Err :: term()}.
+consult(FName) ->
+    case file:consult(FName) of
+        {ok, Terms} ->
+            {ok, Terms};
+        {error, _Reason} ->
```
tburghart, line 167: I think the pattern should be:

``` erlang
{error, {_L, _M, _T}} ->
    % error interpreting terms in the file
    maybe_latin_consult(FName); % more on that below
OpenErr ->
    % error opening the file
    OpenErr
```

src/riak_core_util.erl
```diff
@@ -150,6 +155,55 @@ replace_file(FN, Data) ->
...
+            latin_consult(FName)
```
tburghart, line 168: Shouldn't bother with this unless there's reason to believe it may work. file:consult will have used an encoding marker in the file if there was one, so we can assume there's not one there, or it's wrong :(, or the file's bad, or **maybe** we're tripping over a default encoding mismatch, which could potentially go either way.

Maybe something more like:

``` erlang
consult(FName) ->
    case file:consult(FName) of
        {ok, Terms} ->
            {ok, Terms};
        {error, {_L, _M, _T}} ->
            % error interpreting terms in the file
            alternate_consult(FName);
        OpenErr ->
            % error opening the file
            OpenErr
    end.

alternate_consult(FName) ->
    case file:open(FName, [read]) of
        {ok, Fd} ->
            R = consult_stream(Fd),
            file:close(Fd),
            R;
        Error ->
            Error
    end.

consult_stream(Fd) ->
    % riak_core on the feature/otp-17 branch already has otp_compat
    % available. alternately, we can match the pattern returned from
    % erlang:system_info(otp_release), but that seems less intuitive.
    AltEnc = case otp_compat:otp_version() > 16 of
        true ->
            latin1;
        _ ->
            utf8
    end,
    case epp:set_encoding(Fd, AltEnc) of
        none ->
            consult_stream(Fd, 1, []);
         Enc ->    
            % encoding is explicitly set in the file, we cant override it
            {error, Enc}
    end.
```

src/riak_core_util.erl
russelldb, line 167: I don't follow at all. Sorry.

src/riak_core_util.erl
russelldb, line 168: Are you sure? Why do you think there'd be an encoding marker?  I mean, this works.

In general there is a reason to believe it will work on vnode_status files written in r16 but read in 17. If you think that is too specific I can just move the code to riak_kv_vnode_status_mgr. I can assure you the code is here for a reason.

src/riak_core_util.erl
russelldb, line 935: I think I will probably remove it, the test is in riak_kv_vnode_status_mgr. Maybe the code should be too, actually, since it addresses that specific problem.

src/riak_core_util.erl
russelldb, line 167: I mean, _why_ do you think that, when what is there works as is?

seancribbs, line 167: I think he means that you could get a POSIX error like `enoent` or `eaccess`, which should be distinguished from a decoding error.

russelldb, line 167: That makes sense, good catch.

tburghart, line 167: correct

src/riak_core_util.erl
russelldb, line 168: @tburghart that impl looks way better than mine. Want to push it to the branch rather than have me copy/paste it and push?  And also then I can click on lines and ask "why this?" etc, like why do the version check?

tburghart, line 168: I'm assuming this might be used with files we haven't thought of yet, so an encoding marker **_could**_ be present in the file. If it is, `epp:set_encoding` won't override it, so we shouldn't bother trying.

The version check is there to make sure the retry uses the opposite encoding from what `file:consult` used, so we're not just performing the same operation the hard way. If run on R16, there's no reason to retry reading it in `latin1` encoding, because `file:consult` already tried that..

russelldb, line 168: Yeah. You know what, I really don't like code that may or may not do X in future. I think I will just move this all to riak_kv_vnode_status_mgr to handle that one specific problem.

tburghart, line 168: I'll add it to the repo ...

russelldb:
Killing this. It should be for the r17 branch. Stupidly I conflated the issue of corrupt vnode status files killing nodes and the r17 issue. Someone else can deal with r17, or I'll do it later. The KV PR just needs a a tweak to work fine on r16, and when the r17 issue comes up, @tburghart's code on riak_core can be added.

PR #735, opened 2015-05-05
provide behaviour info for riak_core_vnode_worker with -callbacks

hmmr:
For some reason, dialyzer fails to use exported behaviour_info/1 and
goes on to emit warnings in riak_kv.  But it works fine with callback
specs expressly provided.


cmeiklejohn:
The reason is that the callback format changed between R15 and R16 from `behavior_info` to `-callback`.

hmmr:
Fair enough. A compiler warning would be appropriate, imho, once the old form becomes useless for dialyzer.

Anyway, this fix serves to remove one line in riak_kv/dialyzer.ignored-warnings, which is why it's here.

hmmr:
Will retarget to 2.1; closing.

PR #738, opened 2015-05-18
The default cli schema is located in riak_core/priv

andreashasse:
When not configuring schema_dirs the riak_core app can not start. The code that is suppose to handle the case seems to point to the wrong location. There is a riak_core.schema in riak_core/priv. I assumed that this is the one that should be used as a default.

### DIFF
--- a/src/riak_core_cli_registry.erl
+++ b/src/riak_core_cli_registry.erl
@@ -50,5 +50,5 @@ load_schema() ->
         {ok, Directories} ->
             ok = clique_config:load_schema(Directories);
         _ ->
-            ok = clique_config:load_schema([code:lib_dir()])
+            ok = clique_config:load_schema([code:priv_dir(riak_core)])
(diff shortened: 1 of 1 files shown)


andrewjstone:
When building a riak release we copy all schema files to the lib dir, not just riak_core's. We then load them all at once. https://github.com/basho/riak/blob/develop/rel/reltool.config#L97-L120

Each app can have it's own schema that you may want to use, so the directory shouldn't be limited to just the riak_core. This was an easy way to handle this in riak.

There is perhaps an argument to be made that copying all the schemas to the libdir is the wrong choice, and they should be either stored somewhere else or registered at run-time like other clique entities. However, since the path is configurable as an environment variable it doesn't seem to be that big of a deal.

Closing, as we at least don't want this change.

andreashasse:
Ah, thanks for the reference.

PR #746, opened 2015-06-20
A module to plug into riak_net_rla as metadata-backed storage backend

hmmr:
This is part of [NET-4](https://bashoeng.atlassian.net/browse/NET-4).
- A new module, `riak_core_metadata_rla_backend`, is added, to be used as a cluster metadata -backed storage backend for Resource Location Authority (RLA, existing in `riak_net`). It makes metadata get/put/delete operations available for RLA-specific needs (register/deregister/lookup/list).
- Introduce a dependency on riak_net, where the behaviour used in that module is defined.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -13,6 +13,7 @@
   {lager, "2.0.3", {git, "git://github.com/basho/lager.git", {tag, "2.0.3"}}},
   {poolboy, ".*", {git, "git://github.com/basho/poolboy.git", {tag, "0.8.1p3"}}},
   {basho_stats, ".*", {git, "git://github.com/basho/basho_stats.git", {tag, "1.0.3"}}},
+  {riak_net, ".*", {git, "git://github.com/basho/riak_net.git", {branch, "feature/az/resource-location-authority"}}},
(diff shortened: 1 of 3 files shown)


sumerman:
Introducing dependency on a closed source library into an open source one is a bad idea.

hmmr:
@sumerman That's temporary, pending construction works in riak_net. riak_net will be made public before this PR gets reviewed and merged.

sumerman:
@hmmr It is also a dependency most users probably do not want to have. We already have leveldb because promised hashtree-backend refactoring never happened.

hmmr:
@sumerman I cannot speculate on leveldb issue, but there's little to be afraid of wrt encumbering riak_core with non-essential dependencies. Firstly, because riak_net is just half a dozen files with no heavy dependencies on its own, and a single, self-contained Erlang app; second, it's indeed meant to make life easier for the various intra-cluster connectivity tasks; and third, again, this dependency won't happen until riak_net becomes public and production-ready.

hmmr:
Closing this PR for now, to avoid possible confusion.

PR #747, opened 2015-06-23
Update riak_core to use clique 0.3.0

nickelization:
Clique 0.3.0 introduces an extra argument that's passed to the command callbacks. The existing code in riak_core that uses clique doesn't need this extra argument, so we can just ignore it, but we still need to add an extra _ argument at the beginning of each command callback.


paegun:
+1 looks great, moved up to clique 0.3.0. Thank you for clarifying the unused first arg is _BaseCmd.

paegun:
+1 even better with the unused arg named :+1:

paegun:
+1 ddf4c9c8a1bd6f921ebef2766ae6fae83414fa0f

nickelization:
@borshop merge

PR #751, opened 2015-07-09
Revert removal of legacy gossip as it is a breaking API change.

JeetKunDoug:
- Found when trying to update MDC to riak_core/2.1


jonmeredith:
+1 16b0020

jonmeredith:
@borshop merge

JeetKunDoug:
This was necessary because MDC uses functionality to downgrade/upgrade rings when transmitting data, and we'd be breaking mixed cluster-of-clusters MDC of we removed it.
2 Options for fixing this
1) Keep downgrade/upgrade
2) Fix MDC to do things better (but, of course, we still need to get a few versions away from the change before we'll be able to kill this functionality).

PR #752, opened 2015-07-10
Bump clique to version 0.3.2

nickelization:
The latest version of clique includes several bugfixes over 0.3.1.


nickelization:
To help make review easier, here are the changes between clique 0.3.1 and 0.3.2: https://github.com/basho/clique/compare/0.3.1...0.3.2

0.3.2 fixes some regressions that broke the set/show/describe commands, and also cleans up the code for generating error output if a command is not found and has no associated usage message.

The second bit isn't really relevant to riak_core, since any command that riak-admin can pass in will be a supported command with usage info already registered. But we definitely don't want to break riak-admin set/show/describe, so this is a fairly important PR for those fixes alone.

nickelization:
Closing this PR so that I can recreate it against the 2.1 branch.

PR #759, opened 2015-08-04
Fix/az/assorted dialyzer issues

hmmr:
A trivial clean-up work to squelch some of the dialyzer warnings. The first two commits replace the old-style `behaviour_info/1` with a series of `-callback`s, and a third one adds `cluster_info` as a dependency in `rebar.config`.

In the newly added mention of `cluster_info` in `rebar.config`, I pinned it to tag 2.0.3, which as of this writing, is the latest tag it has.

### DIFF
--- a/src/riak_core_coverage_fsm.erl
+++ b/src/riak_core_coverage_fsm.erl
@@ -3,7 +3,7 @@
 %% riak_core_coverage_fsm: Distribute work to a covering set of VNodes.
 %%
 %%
-%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2015 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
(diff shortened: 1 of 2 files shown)


macintux:
Do you know the impact of making `cluster_info` a dependency? It looks like it's intended to be optional. /cc @slfritchie

hmmr:
Hmm.. One reason for this dependency to not be in `rebar.config`, would be the reciprocal dependency of `cluster_info` on `riak_core`, but it doesn't appear to be the case. What could it be, then?

Going to look for clues more closely, meanwhile.

macintux:
My concern is that if it adds overhead and/or bug/attack vectors, adding it to all instances of Riak is a bad idea. Today it looks like an optional add-on for troubleshooting, and there’s no obvious way to disable it if it’s present in the environment.

If we’re going to add it to rebar.config, we should also add a flag that’s false by default for registering it.

-John

hmmr:
Aha, I see it now: https://github.com/basho/riak_core/blob/develop/src/riak_core_app.erl#L63. I removed the last commit, pushed again.

macintux:
I don't see any reduction of dialyzer messages (except the cluster_info stuff which was still present when I grabbed this branch). What messages do the callback changes suppress?

hmmr:
They are in riak_kv, actually: https://github.com/basho/riak_kv/blob/develop/dialyzer.ignore-warnings#L52 and L54 (or https://github.com/basho/riak_kv/blob/2.1/dialyzer.ignore-warnings#L20 and L21).

PR #764, opened 2015-08-11
Do not allow node to join cluster when still launching

macintux:
@mrallen1 and @jonmeredith identified a race condition when nodes are being joined quickly during automated testing; nodes which are still busy starting applications can lead to divergent rings and a node can transition from being a member of a ring back to joining status with resulting deadlock.

The race condition only manifests itself during staged joins; this change checks the joining node's `init` status and blocks the join if the node is starting (or stopping).

Corresponding `riak_test` PR to follow.

### DIFF
--- a/src/riak_core.erl
+++ b/src/riak_core.erl
@@ -123,15 +123,25 @@ standard_join(Node, Rejoin, Auto) when is_atom(Node) ->
             {error, not_reachable}
     end.
 
+init_complete({started, _}) ->
+    true;
+init_complete(_) ->
+    false.
+
 standard_join(Node, Ring, Rejoin, Auto) ->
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
I can't see what the failures on the builder are, but this appears to be straightforward enough.

macintux:
The failure is the lock problem that's been biting `riak_core` builds. Borshop retry doesn't work until someone plusses it IIRC.

cmeiklejohn:
:+1: 3edf54a

macintux:
@borshop merge

jonmeredith:
RIAK-2064 just for lovely tracking purposes.

PR #765, opened 2015-08-11
Address default bucket type defaults in the absence of riak.conf

macintux:
See basho/riak#727 (RIAK-1662) (RIAK-1784) for context. Splits out custom bucket type defaults so that the default bucket type, in the presence of a legacy `app.config`, does not pick up `dvv` or `allow_mult=true`.

@russelldb created a test, PR to follow shortly.


jadeallenx:
:+1: 9b205cf

basho/riak_test#782 passes with this PR applied against 2.0.6

macintux:
@borshop merge

PR #766, opened 2015-08-11
Pin eleveldb version to 2.0.7 for riak_core 2.0 branch

hazen:
Previously the `eleveldb` library was linked to the 2.0 branch which meant that unintended features were being pulled into `riak_core` as dependencies.  Going forward we can pin to a fixed release and move to the next tag as needed.


jadeallenx:
:+1: 061c05c

hazen:
@borshop merge

PR #772, opened 2015-08-20
pin eleveldb dependency on branch prototype/timeseries-range-scan

hmmr:


andytill:
+1 e80d9932a3af12f6a0dd03e0586812a1bbe557dc

PR #778, opened 2015-09-08
Event notifications when bucket types are stored by the riak_core_metadata_manager [JIRA: RIAK-2174]

andytill:
A gen_event event manager metadata notifications, currently serving bucket type store events which are triggered for new and updated bucket type metadata.

### DIFF
--- a/src/riak_core_metadata_evt.erl
+++ b/src/riak_core_metadata_evt.erl
@@ -1,0 +1,53 @@
+%% -------------------------------------------------------------------
+%% 
+%% An API for a gen_event event manager for bucket type events.
+%%
+%% Copyright (c) 2015 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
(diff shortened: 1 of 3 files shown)


hmmr:
+1

jonmeredith:
Hold up on merging.  Not all events are bucket types, this is supposed to be a general mechanism and now we're coupling implementation specifics in here.

If we want to break this out / start using the plumbtree library we have to keep this modular.

andytill:
@jonmeredith Are you saying the event should be the `{FullPrefix, Key}` argument that comes into store, and the listener should filter using the `FullPrefix` to only handle bucket types?

cmeiklejohn:
I believe the way this should work is that other applications should _register_ events to be handled based on a key/prefix in a generic way, without putting any notion of bucket types here.

hmmr:
Alternatively, interested parties in riak_kv can just poll certain prefixes in metadata. This will keep changes completely outside riak_core.

plumtree is an option, too (data_platform_ee uses that).

cmeiklejohn:
Are you using the standard plumtree or the patches I've applied? Membership state in the existing implementation has concurrency and data loss bugs unless patched...

hmmr:
That's a question @nickelization or @cuyler can answer.

DSomogyi:
create jira issue

Basho-JIRA:
I have talked to Jon and suggested an API like the following:

{code:erlang}
riak_core_metadata_evt:swap_handler({core,bucket_types}, my_handler_mod, Args).
{code}

The handler will subscribe to events of a particular type to make it more generic for types other than bucket_types. I need to rework the code in the metadata_callbacks branch so that gen_event managers are created when a new type is subscribed to. The notifier will need to lookup the event manager for the type and call notify on that if it exists.  If it doesn't exist then it is a no-op. 

**_[posted via JIRA by Andy Till]**_

andytill:
@hmmr @cmeiklejohn @jonmeredith I have created the new PR, #779 for this change with the suggested changes to the API and the DDL checks at bucket activation.

Basho-JIRA:
#779 has been merged a while ago. I suppose the jira ticket can be closed, too. 

**_[posted via JIRA by Andrei Zavada]**_

PR #782, opened 2015-09-24
Bump Clique to version 0.3.5

nickelization:
Version 0.3.4 adds the ability to return different exit status codes from commands, which should be useful in the future as we convert more riak-admin commands to use Clique.


paegun:
develop is already at 0.3.5, so closing.

Issue #784, opened 2015-10-01
erlang:now is deprecated in Erlang 18 [JIRA: RIAK-2211]


TBK145:
Since erlang:now is deprecated in Erlang 18, it prevents the compilation of riak_core on Erlang 18 (http://www.erlang.org/news/88).

This should probably be replaced with erlang:timestamp (http://www.erlang.org/doc/man/erlang.html#timestamp-0).

matehat:
Filed a pull request for that: #799

PR #785, opened 2015-10-02
Erlang 18 compatible.

TBK145:
Riak_core now compiles on Erlang 18, because all occurrences of (the now deprecated) erlang:now() are replaced by erlang:timestamp().


TBK145:
Issue here: https://github.com/basho/riak_core/issues/784.

mbbroberg:
Awesome work @TBK145! I know our community members, at Basho and externally, would love to see this merge. I'll share this in this week's Riak Recap too.

TBK145:
Thanks!
There's one little catch though. For riak_core to completely compile on Erlang 18, a few of it's dependencies need to be updated too. I already created pull requests for those.

slfritchie:
Hi, please be aware that review will probably require some extra time -- there is probably one or more invocations of now() that implicitly require strictly increasing values, something that os:timestamp() doesn't do.

nickelization:
+1 to Scott's comments, removing the monotonicity guarantee is definitely scary.

Another big issue in the short term is the lack of an erlang:timestamp() function in R16, which is currently the only Erlang version that Riak the database supports.

I already commented about this on the associated riak_ensemble PR since I noticed that one first, but I'll just repeat here: until we can drop support support for older Erlang versions, we might want to just enable the nowarn_deprecated_function compile flag. That way things will still compile on R18, even if warnings_as_errors is enabled; and although erlang:now is deprecated, it's still present and usable in R18. At some point in the future we can look at removing all the erlang:now calls, but in the short term I don't know if that's going to be doable.

cmeiklejohn:
Correct, @slfritchie.  

From examining the Lasp fork of Plumtree, our internal integration with eleveldb, and our fork of Riak Core, it's not sufficient to just replace calls with `os:timestamp` given that most of the calls to `now()` are used to generate a unique actor.

Again, I will refer to use of the `time_compat` library for providing R16 - 18 support, as we've done with Lasp.

jadeallenx:
Agree with @cmeiklejohn that ultimately using `time_compat` seems to be right way forward, but in the short term `-compile(nowarn_deprecated_function)` would work for R18, but most likely not any future releases. 

We pretty much **cannot** substitute `now()` -> `os:timestamp()` for most calls to `now()`.

cmeiklejohn:
If the goal is to get proper 18 support, it's just a matter of finding someone with the available resources to perform the work.  

I'm :-1: on `-compile(nowarn_deprecated_function)` given that ultimately someone will just forget that the flag was added and never actually fix the library.

jadeallenx:
Yeah it's not a great choice @cmeiklejohn.

TBK145:
If a lot of the uses are to have a unique actor, should it be replaced by `erlang:unique_integer()` in Erlang 18?

cmeiklejohn:
It should probably use `erlang:unique_integer([positive])`.

TBK145:
Great, I will change it using time_compat tomorrow.

TBK145:
This time I used the time_compat library, however, a lot of functions were using the Timestamp format, so in those cases I used time_compat:timestamp(). Let me know if this should be done different.

mvaled:
Is this have been moved elsewhere? It seems staled from a long time now.

bsparrow435:
We are currently in process of bringing all our Riak dependencies up to OTP18/19 compatibility. This work is  scheduled to complete shortly after the new year and I'll have some updates to share then. While the changes to the library are small to bring it up to compatibility, many of the dependencies have a more complicated road to be brought up to date. Thank you very much for your contribution and we'll merge this PR once we've finioshed bringing the deps up to date as well.

hachreak:
Hi @bsparrow435, 
how the works for OTP18/19 compatibility are going? There are any news? :smile: 
Could we help somehow?
Thanks.

bsparrow435:
Hello! Ted gave a great talk at Erlang Factory on the status of this work: https://www.youtube.com/watch?v=TcUTsYjEon4

I'll add him to this conversation so he can update on status of all the repos. @tburghart 

Thanks!

hachreak:
Hi @bsparrow435,
thanks to quick reply me! :smile: 
I'll see for sure the video, thanks!
Some minutes ago I found this fork that looks make a kind of "workaround" to the problem: https://github.com/project-fifo/riak_core

bsparrow435:
Yep, if you're looking to build a separate app on Riak core, heinz has done fantastic work on project fifo. Our focus had been on getting the entire Riak database project updated at the same time to coincide with a release. Thanks for contributing and if you have any questions please let us know.

PR #790, opened 2015-10-16
Make DDL distribution more robust

macintux:
- Do not allow DDL to be updated post-activation
- Include a DDL comparison as part of the claimant-driven activation process so that if messages are delayed the claimant doesn't accidentally validate a node who has compiled an earlier version of the DDL but has yet to receive the later version


hmmr:
Looks good to me, but more competent eyes would be appropriate.

JeetKunDoug:
:+1: 42b0dda

macintux:
@borshop merge

PR #792, opened 2015-10-23
Bugfix/hashtree delete reopen

zeeshanlakhani:
- Update hashtree.erl and the associated eqc test to account for deletes/reopenings across possible race conditions, crashes. 
- Correctly handle full vs incremental/partial segment update & rebuild.
- Force rehashing of tree to do a full rebuild.
- Handle leveldb iteration `invalid_iterator` errors correctly when running into '*' segments.

Related: https://github.com/basho/yokozuna/pull/578 + https://github.com/basho/riak_kv/pull/1242.

@jonmeredith please review.

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -1,6 +1,6 @@
 %% -------------------------------------------------------------------
 %%
-%% Copyright (c) 2012 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2012-2015 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
@@ -100,7 +100,6 @@
 %% ``local_compare/2'' and ``do_remote/1'' for examples (-ifdef(TEST) only).
 
 -module(hashtree).
-
 -export([new/0,
          new/2,
          new/3,
@@ -125,12 +124,20 @@
          segments/1,
          width/1,
          mem_levels/1,
-         path/1]).
+         path/1,
+         next_rebuild/1,
+         set_next_rebuild/2,
+         mark_open_empty/2,
+         mark_open_and_check/2,
+         mark_clean_close/2]).
 -export([compare2/4]).
 -export([multi_select_segment/3, safe_decode/1]).
+
+-define(ALL_SEGMENTS, ['*', '*']).
+-define(BIN_TO_INT(B), list_to_integer(binary_to_list(B))).
+
 -ifdef(TEST).
--compile([export_all]).
--export([local_compare/2]).
+-export([fake_close/1, local_compare/2]).
 -export([run_local/0,
          run_local/1,
          run_concurrent_build/0,
@@ -164,6 +171,7 @@
 -type proplist() :: proplists:proplist().
 -type orddict() :: orddict:orddict().
 -type index() :: non_neg_integer().
+-type index_n() :: {index(), pos_integer()}.
 
 -type keydiff() :: {missing | remote_missing | different, binary()}.
 
@@ -175,19 +183,23 @@
 
 -type select_fun(T) :: fun((orddict()) -> T).
 
--record(state, {id             :: tree_id_bin(),
-                index          :: index(),
-                levels         :: pos_integer(),
-                segments       :: pos_integer(),
-                width          :: pos_integer(),
-                mem_levels     :: integer(),
-                tree           :: dict(),
-                ref            :: term(),
-                path           :: string(),
-                itr            :: term(),
-                write_buffer   :: [{put, binary(), binary()} | {delete, binary()}],
+-type next_rebuild() :: full | incremental.
+
+-record(state, {id                 :: tree_id_bin(),
+                index              :: index(),
+                levels             :: pos_integer(),
+                segments           :: pos_integer(),
+                width              :: pos_integer(),
+                mem_levels         :: integer(),
+                tree               :: dict(),
+                ref                :: term(),
+                path               :: string(),
+                itr                :: term(),
+                next_rebuild       :: next_rebuild(),
+                write_buffer       :: [{put, binary(), binary()} |
+                                       {delete, binary()}],
                 write_buffer_count :: integer(),
-                dirty_segments :: array()
+                dirty_segments     :: array()
                }).
 
 -record(itr_state, {itr                :: term(),
@@ -244,6 +256,7 @@ new({Index,TreeId}, LinkedStore, Options) ->
                    mem_levels=MemLevels,
                    %% dirty_segments=gb_sets:new(),
                    dirty_segments=bitarray_new(NumSegments),
+                   next_rebuild=full,
                    write_buffer=[],
                    write_buffer_count=0,
                    tree=dict:new()},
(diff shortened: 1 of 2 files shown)


src/hashtree.erl
```diff
@@ -355,38 +369,68 @@ update_tree(State) ->
...
+    State2 = maybe_clear_buckets(Segments, State),
     State3 = update_tree(Segments, State2),
-    %% State3#state{dirty_segments=gb_sets:new()}.
-    State3#state{dirty_segments=bitarray_new(NumSegments)}.
+    %% State2#state{dirty_segments=gb_sets:new()}
+    State3#state{dirty_segments=bitarray_new(NumSegments),
+                 next_rebuild=incremental}.
+
+maybe_clear_buckets(?ALL_SEGMENTS, State) ->
+    Segments = lists:seq(0, State#state.segments div State#state.width),
```
jonmeredith: Possible off-by-one here, think it should be `State#state.segments div State#state.width - 1`

src/hashtree.erl
```diff
@@ -355,38 +369,68 @@ update_tree(State) ->
...
+maybe_clear_buckets(_Segments, State) ->
+    State.
+
+
+clear_bucket(MinLevel, Level, _Segments, State) when Level =< MinLevel ->
+    State;
+clear_bucket(MinLevel, Level, Segments, State = #state{width = Width}) ->
+    %% ignoring del_bucket state return
+    State2 = lists:foldl(fun(Segment, State1) ->
+                                 del_bucket(Level, Segment, State1)
```
jonmeredith: del_memory_bucket isn't actually used any more, could simplify and remove del_bucket/del_memory_bucket and just call `del_disk_bucket` in the clean. Can't think of any cases where we would need to delete memory buckets individually.

src/hashtree.erl
```diff
@@ -413,14 +457,21 @@ width(#state{width=W}) ->
...
 %%       trees per file. This is intentional. If we want per tree metadata
 %%       this will need to be added as a separate thing.
--spec write_meta(binary(), binary(), hashtree()) -> hashtree().
+-spec write_meta(binary(), binary()|term(), hashtree()) -> hashtree().
 write_meta(Key, Value, State) when is_binary(Key) and is_binary(Value) ->
     HKey = encode_meta(Key),
     ok = eleveldb:put(State#state.ref, HKey, Value, []),
-    State.
+    State;
+write_meta(Key, Value0, State) when is_binary(Key) ->
```
jonmeredith, line 523: Thinking about it, the open/close metadata does need to be per-tree as we could crash during a partial shutdown and there's no way to know when the underlying leveldb will close.

Derp, realizing that's why Joe broke out the flush/update stuff so it could all be called before closing the database.  If you share the leveldb then as soon as the first close is called, you have no opportunity to flush/update_tree on the others.

That probably means reworking the way the opened/closed counters should work.

We didn't spot this because the unit test does not share trees.

src/hashtree.erl
```diff
@@ -590,23 +680,42 @@ get_env(Key, Default) ->
...
-                            Hashes1 = get_bucket(Level, Bucket, StateAcc),
-                            Hashes2 = orddict:from_list(NewHashes),
-                            Hashes3 = orddict:merge(fun(_, _, New) -> New end,
-                                                    Hashes1,
-                                                    Hashes2),
-                            StateAcc2 = set_bucket(Level, Bucket, Hashes3, StateAcc),
-                            NewBucket = {Bucket, hash(Hashes3)},
-                            {StateAcc2, [NewBucket | BucketsAcc]}
-                    end, {State, []}, Groups),
+update_levels(Level, Groups, State, full) ->
```
jonmeredith: I think these two cases could be collapsed to

```
update_levels(Level, Groups, State, How) ->
     {_, _, NewState, NewBuckets} = rebuild_fold(Level, Groups, State, How),
     Groups2 = group(NewBuckets, State#state.width),
     update_levels(Level - 1, Groups2, NewState, How).
```

src/hashtree.erl
```diff
@@ -637,18 +748,24 @@ group(L, Width) ->
 
 -spec get_memory_bucket(integer(), integer(), hashtree()) -> any().
 get_memory_bucket(Level, Bucket, #state{tree=Tree}) ->
-    case dict:find({Level, Bucket}, Tree) of
+    R = case dict:find({Level, Bucket}, Tree) of
```
jonmeredith: debugging change, should cleanup to be the original code.

src/hashtree.erl
```diff
@@ -759,8 +881,28 @@ iterator_move(Itr, Seek) ->
 
 -spec iterate({'error','invalid_iterator'} | {'ok',binary(),binary()},
               #itr_state{}) -> #itr_state{}.
-iterate({error, invalid_iterator}, IS=#itr_state{}) ->
+iterate({error, invalid_iterator}, IS=#itr_state{current_segment='*'}) ->
```
jonmeredith, line 899: Do we need this case as well as the `remaining_segments=['*']` below?

src/hashtree.erl
```diff
@@ -759,8 +881,28 @@ iterator_move(Itr, Seek) ->
...
+    case Segments of
+        [] ->
+            IS;
+        [NextSeg | Remaining] ->
+            Seek = encode(Id, NextSeg, <<>>),
+            IS2 = IS#itr_state{current_segment=NextSeg,
+                               remaining_segments=Remaining,
+                               segment_acc=[],
+                               final_acc=[{CurSeg, F(Acc)} | FinalAcc]},
+            iterate(iterator_move(Itr, Seek), IS2)
```
jonmeredith, line 919: Should check with the backends team if it's safe to move a 'bad' iterator, or if we should really open a new one.

src/hashtree.erl
```diff
@@ -1155,6 +1297,10 @@ peval(L) ->
 %%% EUnit
 %%%===================================================================
 
+-spec ref(hashtree()) -> reference().
+ref(#state{ref=Ref}) ->
```
jonmeredith: Would be better to call the `fake_close` from the EQC test instead of exposing this.

src/hashtree.erl
zeeshanlakhani: +1. figured.

src/hashtree.erl
zeeshanlakhani: +1.

src/hashtree.erl
zeeshanlakhani: ah yes, derp.

src/hashtree.erl
zeeshanlakhani, line 899: will test.

src/hashtree.erl
zeeshanlakhani, line 919: @erikleitch @matthewvon thoughts to @jonmeredith's note/this_code?

src/hashtree.erl
zeeshanlakhani: +1.

src/hashtree.erl
zeeshanlakhani: @jonmeredith would something akin to what i had before ... {<tree_id>, [{opened, <val>}, {closed, <val>}]} make sense again?

src/hashtree.erl
matthewvon, line 919: Iterators are by design available to move.  The only caveat is that they are not intended to be held open for extended periods of time (hello AAE, do you hear me, bastard).  So immediate reuse is good.  Holding open for potential later use is really bad.

src/hashtree.erl
zeeshanlakhani, line 523: nvm previous comment, i get your point @jonmeredith. Do we want to roll some sort of ets+dets localized reader/writer thing (storing per tree-id) and not tying the counter to level? Or is that fraught w/ much peril? :).

src/hashtree.erl
zeeshanlakhani, line 899: Still needed. Must be due to last iterate state?

Without it:

```
{'EXIT',{badarg,[{hashtree,encode,3,[{file,"src/hashtree.erl"},{line,786}]},
                 {hashtree,iterate,2,[{file,"src/hashtree.erl"},{line,877}]},
                 {hashtree,multi_select_segment,3,
                           [{file,"src/hashtree.erl"},{line,838}]},
                 {hashtree,update_tree,2,
                           [{file,"src/hashtree.erl"},{line,412}]},
                 {hashtree,update_perform,1,
                           [{file,"src/hashtree.erl"},{line,382}]},
                 {hashtree_eqc,'-prop_correct/0-fun-3-',2,
                               [{file,"test/hashtree_eqc.erl"},{line,270}]}]}}
[{init,{state,false,undefined,[],[]}},
 {set,{var,1},{call,hashtree_eqc,start,[{16,128,1}]}},
 {set,{var,2},
      {call,hashtree_eqc,write,
            [t2,
             [{<<"0">>,
               <<102,49,159,215,241,62,142,112,131,27,67,123,48,166,207,
                 87,176,231,179,129>>}]]}}]

hashtree_eqc:start({16, 128, 1}) -> ok
hashtree_eqc:write(t2,
    [{<<48>>,
      <<102, 49, 159, 215, 241, 62, 142, 112, 131, 27, 67, 123, 48,
        166, 207, 87, 176, 231, 179, 129>>}]) ->
  ok
false
```

src/hashtree.erl
jonmeredith, line 523: I'd say fraught with peril.  I think we've been trying to fix the library in ways it wasn't quite intended.  hashtree is used as a library by kv_index_hashtree/yz_index-hashtree, and it's those two containers that know if it's shutdown correctly.

We should probably fix by providing some functions for manipulating the open/closed counts and setting the type of rebuild.

`mark_open_and_check(TreeId, State)` - compare opened/closed, set rebuild type, increment opened, 
`mark_clean_close(TreeId, State)` - set closed to opened.

zeeshanlakhani, line 523: Done.

src/hashtree.erl
```diff
@@ -759,8 +878,28 @@ iterator_move(Itr, Seek) ->
...
+iterate({error, invalid_iterator}, IS=#itr_state{current_segment='*'}) ->
     IS;
+iterate({error, invalid_iterator}, IS=#itr_state{itr=Itr,
+                                                 id=Id,
+                                                 current_segment=CurSeg,
+                                                 remaining_segments=Segments,
+                                                 acc_fun=F,
+                                                 segment_acc=Acc,
+                                                 final_acc=FinalAcc}) ->
+    case Segments of
```
jonmeredith, line 908: We should add a comment to explain how you could get here. Suggestion:

```
%%  Likely missing dirty segment encountered - e.g. delete marked segment dirty, but removed last entries for it
```

zeeshanlakhani, line 908: +1

src/hashtree.erl
```diff
@@ -355,41 +368,104 @@ update_tree(State) ->
...
+%%
+%%
+%%      If it was a clean shutdown, set `next_rebuild' to be an incremental one.
+%%      Otherwise, if it was a dirty shutdown, set `next_rebuild', instead,
+%%      to be a full one.
+-spec mark_open_and_check(index_n()|binary(), hashtree()) -> hashtree().
+mark_open_and_check(TreeId, State) when is_binary(TreeId) ->
+    MetaTerm = read_meta_term(TreeId, [], State),
+    OpenedCnt = proplists:get_value(opened, MetaTerm, 0),
+    ClosedCnt = proplists:get_value(closed, MetaTerm, 0),
```
jonmeredith: Thinking about upgrading, it would be safer to default closed to -1 so that it forced a full rebuild. Existing AAE trees may have had their dirty segments lost on restart.

Anything creating a new known-empty hashtree should explicitly set the open metadata to a known 0,0 before calling mark_open_and_check - maybe mark_open_empty?

Otherwise trees in potentially unknown state are marked as good until the next rebuild is triggered in 7 days.

zeeshanlakhani: +1... had something like this earlier, will make it right.

src/hashtree.erl
```diff
@@ -355,41 +368,104 @@ update_tree(State) ->
...
-    State2 = snapshot(State),
-    rehash_perform(State2).
+    State2 = flush_buffer(State),
+    State3 = snapshot(State2),
+    rehash_perform(State3).
 
 -spec rehash_perform(hashtree()) -> hashtree().
 rehash_perform(State) ->
-    Hashes = orddict:from_list(hashes(State, ['*', '*'])),
+    Hashes = orddict:from_list(hashes(State, ?ALL_SEGMENTS)),
```
zeeshanlakhani: test

test/hashtree_eqc.erl
```diff
@@ -43,142 +59,206 @@ sha(Bin) ->
...
+
+    Insert(t1, [lists:keyfind(K, 1, Only2) ||  K <- Missing, lists:keyfind(K, 1,
+                Only2) /= false]),
+    Insert(t2, [lists:keyfind(K, 1, Only1) ||  K <- RemoteMissing, lists:keyfind(K, 1,
+                Only1) /= false]),
+    Insert(t2, [lists:keyfind(K, 1, Only1) ||  K <- Different, lists:keyfind(K, 1,
+                Only1) /= false]),
+    ok.
+
+%% write_differing(Tree1, Tree2, {Key, Hash1}, Hash2) ->
```
jonmeredith: Could just remove this - it doesn't work or have any value in the new model.

jonmeredith:
+1  a590d5e

jonmeredith:
+1 f55a61d
(after rebase)

zeeshanlakhani:
@borshop merge

PR #793, opened 2015-11-06
Push more `riak-admin` commands through `clique`

sanmiguel:
There are still commands that do not go via `clique` but I want to checkpoint this now before I go further.

NB: This will require a corresponding PR in riak to update the `riak-admin` script.

Commands now available via clique:
- `riak-admin member-status`
- `riak-admin security *`

`riak-admin member-status` is largely a duplicate of `riak-admin cluster status` with the addition of node-counts by status ('Joining', 'Valid' etc).

I've gone a little further than in refactoring than originally intended for `riak-admin security` commands: I found it annoying to have to add usage in one place, command spec in another and remember to update the base usage string for each sub-command. To tackle this, command spec and usage are now in one (or more, see `riak_core_security_cli:grant_*`) function. Usage is split into 2 parts: structured usage; and description. This way, we simply gather all structured usage lines to build the overall base usage. No repetition, no having to cross-check in completely different parts of the file.

I'm still not 100% sold that this is a sufficiently objective improvement but subjectively it feels better IMHO.

Aside from this, because we've now separated the code to print the output from the code to structure it, the tests can now actually verify what would get displayed rather than simply assert that no exceptions are thrown.

@cuyler @nickelization : Gents, I welcome any and all comments, suggestions and criticisms either here or in person next week!


nickelization:
Hey @sanmiguel, sorry for the long lag time on reviewing. I will start taking a look this evening, but realistically probably won't finish reviewing until sometime tomorrow.

sanmiguel:
Hey @nickelization no worries! Let me know if you have any questions while you're going through it.

nickelization:
So, I thought of a potentially tricky point that we might need to address. We may want to keep the output format of certain commands precisely the same, to avoid breaking any tools customers may have written to parse riak-admin command output. I believe this is why, for instance, we have a new "riak-admin cluster status" command that overlaps a lot with the much older "riak-admin member-status": I think the original intent was to deprecate member-status in favor of cluster status, but to leave both commands in so that anyone relying on member-status would not be forced to update their infrastructure.

I think it was either @mrallen1 or @macintux that implemented the cluster status command so maybe one of them could confirm or deny my vague memories of this.

src/riak_core_ring_cli.erl
```diff
@@ -0,0 +1,90 @@
...
+-export([
+         register_cli/0,
+         member_status/3
+        ]).
+
+-spec register_cli() -> ok.
+register_cli() ->
+    register_cli_usage(),
+    register_cli_cmds(),
+    ok.
```
nickelization, line 13: Minor detail, but the `ok` here is redundant now that `register_cli_cmds` is implemented using `lists:foreach`.

jadeallenx:
Unfortunately we can't really change the output of the older commands without breaking some customers automation tools. It occurs to me that _perhaps_ - I haven't thought it through closely - we ought to have some kind of `deprecated/2` output function that takes as its parameters when effective and what command is replacing it.  This warning header could be disabled given an appropriate command line flag `--deprecation-warnings off` or something? That _probably_ wouldn't break people's tools and humans who read it would be aware of the changes.

We could use that to warn people that these older commands have been superseded by others and eventually disable or reroute the command to the new command.

Honestly @andrewjstone would know best of all of us how carefully we ought to proceed if he feels like weighing in. :)

PR #798, opened 2015-12-17
Support TS coverage requests

macintux:
WIP. @hmmr will be carrying this set of PRs forward.

This is a significant reworking of how coverage requests are handled, including a change to the `riak_core_coverage_fsm` behaviour to allow different Riak core apps to plug in different coverage plan modules (thus eliminating the `colocated` tuple).

### DIFF
--- a/src/riak_core_coverage_fsm.erl
+++ b/src/riak_core_coverage_fsm.erl
@@ -3,7 +3,7 @@
 %% riak_core_coverage_fsm: Distribute work to a covering set of VNodes.
 %%
 %%
-%% Copyright (c) 2007-2011 Basho Technologies, Inc.  All Rights Reserved.
+%% Copyright (c) 2007-2015 Basho Technologies, Inc.  All Rights Reserved.
 %%
 %% This file is provided to you under the Apache License,
 %% Version 2.0 (the "License"); you may not use this file
--- a/src/riak_core_coverage_fsm.erl
+++ b/src/riak_core_coverage_fsm.erl
@@ -223,15 +223,17 @@ find_plan(Mod, Target, NVal, PVC, ReqId, Service, Request) ->
 -spec interpret_plan(vnode_coverage()) ->
                             {list({index(), node()}),
                              list({index(), list(index())|tuple()})}.
-interpret_plan(#vnode_coverage{vnode_identifier=TargetHash,
-                               subpartition={Mask, BSL}}) ->
-    {[{TargetHash, node()}], [{TargetHash, {Mask, BSL}}]};
-interpret_plan(#vnode_coverage{vnode_identifier=TargetHash,
-                               partition_filters=[]}) ->
+interpret_plan(#vnode_coverage{vnode_identifier  = TargetHash,
+                               partition_filters = [],
+                               subpartition = undefined}) ->
     {[{TargetHash, node()}], []};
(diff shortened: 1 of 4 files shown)


src/riak_core_coverage_fsm.erl
```diff
@@ -206,20 +212,27 @@ maybe_start_timeout_timer(Timeout) ->
...
+find_plan(_Mod, #vnode_coverage{}=Plan, _NVal, _PVC, _ReqId, _Service,
+          _Request) ->
+    interpret_plan(Plan);
+find_plan(Mod, Target, NVal, PVC, ReqId, Service, Request) ->
+    Mod:create_plan(Target, NVal, PVC, ReqId, Service, Request).
+
+%% @private
+%% Take a `vnode_coverage' record and interpret it as a mini coverage plan
+-spec interpret_plan(vnode_coverage()) ->
+                            {list({index(), node()}),
```
lehoff, line 224: Not something to hold back the PR, but I think we should define an alias for both `{index(), node()}` and `{index(), list(index())|tuple()}` and put them in a module (not in hrl file!).

src/riak_core_coverage_plan.erl
```diff
@@ -177,13 +180,22 @@ replace_traditional_chunk(VnodeIdx, Node, Filters, NVal,
...
+            MinPar =< RingSize ->
+                RingSize;
+            true ->
+                next_power_of_two(MinPar)
+        end,
+    create_subpartition_plan(VNodeTarget, NVal, Count, PVC, ReqId, Service);
 create_subpartition_plan(VNodeTarget, NVal, Count, PVC, ReqId, Service) ->
-    %% IMPORTANT: `Count' is assumed to be a power of 2 as determined
-    %% by `riak_client'. Anything else will behave badly.
+    %% IMPORTANT: `Count' is assumed to be a power of 2. Anything else
```
lehoff, line 197: I would do a check on the power of 2 and crash here instead of carrying on with something we now will behave badly.

src/riak_core_coverage_plan.erl
```diff
@@ -759,6 +764,17 @@ covers(KeySpace, CoversKeys) ->
 data_bits(PartitionCount) ->
     160 - round(math:log(PartitionCount) / math:log(2)).
 
+%% @private
+
+%% Crimes against computerkind. Finds the next value of two greater
+%% than the requested value using string manipulation.
+next_power_of_two(X) ->
+    Next = 1 bsl length(hd(io_lib:format("~.2b", [X]))),
```
lehoff: We cannot have crimes against computerkind ;-)

``` erlang
next_power_of_two(N) ->
    round(math:pow(2,round(math:log2(N-1) + 0.5))).
```

That should calculate it precisely.

macintux: log2 isn’t present in R16.

I could do some magic with log, but it would probably be slower than the current implementation.

-John

lehoff: log2(N) = log(N)/log(2)

```
log2(X) -> math:log(X)/math:log(2).
```

I cannot imagine that being slower than having fun with `io_lib` and `length`.

lehoff:
+1

hmmr:
@borshop merge

PR #799, opened 2015-12-23
Erlang 18 patch

matehat:
This removes all use of `erlang:now/0`, in two ways depending on where it was used:
1. `erlang:unique_integer/0` when it was used to immediately create a unique hash, since the unicity property is garanteed with that simple function call.
2. `riak_core_util:unique_seed/0` (new) which generates a 3-tuple of positive integers (garanteed to be unique using `erlang:unique_integer/0`) when it was expecting that structure.

### DIFF
--- a/src/chashbin.erl
+++ b/src/chashbin.erl
@@ -37,7 +37,7 @@
 
 -record(chashbin, {size   :: pos_integer(),
                    owners :: owners_bin(),
-                   nodes  :: tuple(node())}).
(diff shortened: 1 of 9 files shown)


DeadZen:
Is this a duplicate or improvement of #785 (RIAK-2226) ?

matehat:
@DeadZen duplicate indeed

DeadZen:
Great, thanks for the clarification @matehat

PR #805, opened 2016-01-25
register a ddl_req_version capability, needed for #ddl_v2{}

hmmr:
RTS-865

Goes with https://github.com/basho/riak_kv/pull/1332.

The new version of the ddl record has an additional field, containing table/bucket type properties.

### DIFF
--- a/src/riak_core_app.erl
+++ b/src/riak_core_app.erl
@@ -2,7 +2,7 @@
 %%
 %% riak_core: Core Riak Application
 %%
-%% Copyright (c) 2007-2010 Basho Technologies, Inc.  All Rights Reserved.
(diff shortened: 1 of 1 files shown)


jonmeredith:
Does riak_core need to know about riak kv/ql DDL records? I think this should be registered in riak_kv_app instead - unless there are any uses of the DDL structures at the riak_core layer.

hmmr:
@jonmeredith true, riak_core has no reference to `#ddl_v1{}`. On the other hand, wouldn't it _need_ to know about which version of the ddl record it is when an old node gets a new version of DDL pushed to it as part of ddl propagation?

But ultimately, nothing in riak_core introspects the record itself, and any differences in capabilities are only relevant within and for riak_kv, so you are correct.

PR #806, opened 2016-01-26
relax check for DDL compiled status, don't compare DDL objects

hmmr:
RTS-865

Needed for basho/riak_kv#1332, requires https://github.com/basho/riak_ql/pull/99.

The compiled DDL structures will be different between `#ddl_v1{}` and `#ddl_v2{}` due to the presence of additional 'properties' field in the latter record. With respect to representing DDL schemata, both versions are identical, however, so we can avoid passing the DDL object for comparison (eventually in `riak_kv_metadata_store_listener:handle_info(is_type_compiled)`).

### DIFF
--- a/src/riak_core_metadata_evt_sup.erl
+++ b/src/riak_core_metadata_evt_sup.erl
@@ -79,7 +79,13 @@ is_type_compiled(BucketType, DDL) when is_binary(BucketType) ->
     case ets:lookup(?TABLE, ?BUCKET_TYPE_PREFIX) of
         [{?BUCKET_TYPE_PREFIX, Pid}] ->
             Handlers = gen_event:which_handlers(Pid),
-            Req = {is_type_compiled, [BucketType, DDL]},
+            %% riak_ts-1.1 had in
+            %% riak_kv_metadata_store_listener:is_type_compiled a
+            %% comparison of DDL objects, so downgrade our DDL for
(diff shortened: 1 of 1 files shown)


src/riak_core_metadata_evt_sup.erl
```diff
@@ -79,7 +79,13 @@ is_type_compiled(BucketType, DDL) when is_binary(BucketType) ->
     case ets:lookup(?TABLE, ?BUCKET_TYPE_PREFIX) of
         [{?BUCKET_TYPE_PREFIX, Pid}] ->
             Handlers = gen_event:which_handlers(Pid),
-            Req = {is_type_compiled, [BucketType, DDL]},
+            %% riak_ts-1.1 had in
+            %% riak_kv_metadata_store_listener:is_type_compiled a
+            %% comparison of DDL objects, so downgrade our DDL for
+            %% that (unnecessary) check to succeed:
```
andytill, line 85: It is not unnecessary, the DDL has to be compiled to a module on all nodes or pretty much all TS operations are going to fail with `undef` errors.

hmmr, line 85: But, we still don't need to compare DDL objects, in order to just determine if it is compiled, do we?

hmmr:
Will need to be resubmitted to riak_ts-develop-1.3.

PR #809, opened 2016-02-04
merge 2.0 to develop_2.2 for eleveldb update

JeetKunDoug:


jadeallenx:
Echoing my thoughts from HipChat, I would really appreciate some kind of explanation why the old version is 2.1.10 and its now going to be 2.0.12 - seems like in 6 months I'll be confused by the regression on semver.

jadeallenx:
:+1: 2d4c553

JeetKunDoug:
@borshop merge

PR #827, opened 2016-03-30
Prevent handoff.ip to be local [JIRA: RIAK-3060]

Licenser:
Horrible things happen when the handoff IP is set to localhost, and with horrible things I mean that monsters are going to eat all the data when nodes are being removed or added o.O

### DIFF
--- a/include/riak_core_handoff.hrl
+++ b/include/riak_core_handoff.hrl
@@ -10,6 +10,8 @@
 -define(PT_MSG_SYNC, 3).
 -define(PT_MSG_CONFIGURE, 4).
 -define(PT_MSG_BATCH, 5).
+-define(PT_MSG_VERIFY_NODE, 6).
+-define(PT_MSG_UNKNOWN, 255).
 
 -record(ho_stats,
         {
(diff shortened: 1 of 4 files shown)


nickelization:
Hi Heinz, thanks for the pull request, this is definitely something we've run into before :). See http://docs.basho.com/riak/latest/community/product-advisories/210-dataloss/

However, we can't completely forbid the handoff IP from ever being set to localhost, since we often run multiple nodes on the same machine during development and testing.

It would be great if you wanted to submit a patch that is more selective, which guards against handing off to ourselves, but does so without prohibiting local multi-node setups. I don't know if we can accept this patch as it stands, though.

Licenser:
I can see that this makes  testing a bit harder, however you can just as well use an machine local non `127.0.0.1` ip during tests. My personal take is that users not deleting their data outweighs not having to set an ip during tests.

I agree that a proper tests of 'not handoff to yourself' is the something that would improve the security, but even with that in place, forbidding users to set it to a most likely wrong value would seem like a good start.

bashopatricia:
create jira issue

PR #828, opened 2016-04-01
AAE delete edge case

jonmeredith:
## Context

This pull request resolves issues discovered while debugging yz_aae_test.
The test would fail when AAE did not repair the same set of keys.

Those particular keys were deletes that removed the last entry in the segments at the end of a hashtree (but when there was another treeid following it in the same leveldb database).

This PR fixes the test to detect the problem and also fixes hashtree itself.

One other problem detected with the updated EQC model is that the open/close counters only work correctly if the rebuild mode is set to full between snapshot creation and and update perform.  The *_index_hashtree modules will need to implement that as separate PRs.
## Reproduction

If you check code out to the first commit in the PR, the test has been modified, but the fix not applied yet.  It often takes longer than the eunit test time, so you can provoke from the console using

```
eqc:quickcheck(eqc:testing_time(30*60,hashtree_eqc:prop_correct())).
```

Make sure the .eunit is in your path to pick up hashtree with the right exports.
## Example Failure

```
hashtree_eqc:start({1048576, 1024, 0}, [{1, 0}, {1, 1}], mark_empty, mark_empty,
    []) ->
  ok
hashtree_eqc:write(t2,
    [{<<50, 51>>,
      <<150, 155, 129, 115, 36, 118, 147, 22, 134, 106, 101, 83, 105,
        135, 166, 182, 63, 51, 224, 94>>},
     {<<48>>,
      <<150, 155, 129, 115, 36, 118, 147, 22, 134, 106, 101, 83, 105,
        135, 166, 182, 63, 51, 224, 94>>}]) ->
  ok
hashtree_eqc:delete(t1, <<50, 51>>) -> ok
hashtree_eqc:delete(t2, <<49, 53>>) -> ok
hashtree_eqc:update_snapshot(t1, s1) -> ok
hashtree_eqc:update_snapshot(t2, s2) -> ok
hashtree_eqc:update_perform(t2) -> ok
hashtree_eqc:update_perform(t1) -> ok
hashtree_eqc:local_compare() -> [{missing, <<48>>}]

Reason:
  Post-condition failed:
  [{missing, <<48>>}, {missing, <<50, 51>>}] /= [{missing, <<48>>}]
```

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -410,6 +410,25 @@ update_tree(Segments, State=#state{next_rebuild=NextRebuild, width=Width,
                                    levels=Levels}) ->
     LastLevel = Levels,
     Hashes = orddict:from_list(hashes(State, Segments)),
+    %% Paranoia to make sure all of the hash entries are updated as expected
+    lager:debug("segments ~p -> hashes ~p\n", [Segments, Hashes]),
+    case {Segments, length(Segments), length(Hashes)} of
+	{?ALL_SEGMENTS, _, _} ->
+	    NextRebuild = full, % Paranoia to crash if ever here on incremental
+	    ok;
+	{_, Len, Len} ->
+	    ok;
+	{_, _SegLen, _HashLen} ->
+	    %% At this point the hashes are no longer sufficient to update
+	    %% the upper trees.  Alternative is to crash here, but that would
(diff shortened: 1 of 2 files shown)


jonmeredith:
Closed in favor of #830.

PR #830, opened 2016-04-05
AAE canonical trees [JIRA: RIAK-2644]

jonmeredith:
N.B. Successor to #828, requires basho/eleveldb#188 merged for dialyzer to pass.
## Context

This pull request resolves issues discovered while debugging yz_aae_test.
The test would fail when AAE did not repair the same set of keys.
## Hashtree Fixes
- If keys in the last segments of the tree were deleted (but when there was another treeid following it in the same leveldb database), then the accumulator function was not called with an empty list of hashes for the remaining final segments.
- Always calls prefetch_stop after using iterators to make sure their output is ordered.
- Adds a paranoia check to `update_tree` to do a full rebuild if the list of dirty hashes is not the right length, and log.
- Replaces the previous bucket clearing code which issued a delete for all buckets regardless whether they were populated.  Now scans the on disk bucket keys to do the removal.
- Removes empty segments/buckets from higher levels to keep the tree canonical.
- Avoids empty call to leveldb:write if the flush buffer is empty.
## Model Fixes:
- Open/close counters only work correctly if the rebuild mode is set to full between snapshot creation and and update perform.  The *_index_hashtree modules will need to implement that as separate PRs.
- Now checks top level hashes if the compare set is empty to make sure the trees are canonical.
## Reproduction

If you check code out to the first commit in the PR, the test has been modified, but the fix not applied yet.  It often takes longer than the eunit test time, so you can provoke from the console using

```
eqc:quickcheck(eqc:testing_time(30*60,hashtree_eqc:prop_correct())).
```

Make sure the .eunit is in your path to pick up hashtree with the right exports.
## Example Failure

```
hashtree_eqc:start({1048576, 1024, 0}, [{1, 0}, {1, 1}], mark_empty, mark_empty,
    []) ->
  ok
hashtree_eqc:write(t2,
    [{<<50, 51>>,
      <<150, 155, 129, 115, 36, 118, 147, 22, 134, 106, 101, 83, 105,
        135, 166, 182, 63, 51, 224, 94>>},
     {<<48>>,
      <<150, 155, 129, 115, 36, 118, 147, 22, 134, 106, 101, 83, 105,
        135, 166, 182, 63, 51, 224, 94>>}]) ->
  ok
hashtree_eqc:delete(t1, <<50, 51>>) -> ok
hashtree_eqc:delete(t2, <<49, 53>>) -> ok
hashtree_eqc:update_snapshot(t1, s1) -> ok
hashtree_eqc:update_snapshot(t2, s2) -> ok
hashtree_eqc:update_perform(t2) -> ok
hashtree_eqc:update_perform(t1) -> ok
hashtree_eqc:local_compare() -> [{missing, <<48>>}]

Reason:
  Post-condition failed:
  [{missing, <<48>>}, {missing, <<50, 51>>}] /= [{missing, <<48>>}]
```

### DIFF
--- a/src/hashtree.erl
+++ b/src/hashtree.erl
@@ -410,6 +410,25 @@ update_tree(Segments, State=#state{next_rebuild=NextRebuild, width=Width,
                                    levels=Levels}) ->
     LastLevel = Levels,
     Hashes = orddict:from_list(hashes(State, Segments)),
+    %% Paranoia to make sure all of the hash entries are updated as expected
+    lager:debug("segments ~p -> hashes ~p\n", [Segments, Hashes]),
+    case {Segments, length(Segments), length(Hashes)} of
+	{?ALL_SEGMENTS, _, _} ->
+	    NextRebuild = full, % Paranoia to crash if ever here on incremental
+	    ok;
+	{_, Len, Len} ->
+	    ok;
+	{_, _SegLen, _HashLen} ->
+	    %% At this point the hashes are no longer sufficient to update
+	    %% the upper trees.  Alternative is to crash here, but that would
+	    %% lose updates and is the action taken on repair anyway.
+	    %% Save the customer some pain by doing that now and log.
+	    %% Enable lager debug tracing with lager:trace_file(hashtree, "/tmp/ht.trace"
+	    %% to get the detailed segment information.
+	    lager:warning("Incremental AAE hash was unable to find all required data, "
+			  "forcing full rebuild of ~p", [State#state.path]),
+	    update_perform(State#state{next_rebuild = full})
+    end,
     Groups = group(Hashes, Width),
     update_levels(LastLevel, Groups, State, NextRebuild).
 
@@ -707,6 +726,7 @@ update_levels(0, _, State, _) ->
     State;
 update_levels(Level, Groups, State, Type) ->
     {_, _, NewState, NewBuckets} = rebuild_fold(Level, Groups, State, Type),
+    lager:debug("level ~p hashes ~w\n", [Level, NewBuckets]),
     Groups2 = group(NewBuckets, State#state.width),
     update_levels(Level - 1, Groups2, NewState, Type).
 
@@ -868,10 +888,18 @@ multi_select_segment(#state{id=Id, itr=Itr}, Segments, F) ->
                    encode(Id, First, <<>>)
            end,
     IS2 = iterate(iterator_move(Itr, Seek), IS1),
-    #itr_state{current_segment=LastSegment,
+    #itr_state{remaining_segments = LeftOver,
+               current_segment=LastSegment,
                segment_acc=LastAcc,
                final_acc=FA} = IS2,
-    Result = [{LastSegment, F(LastAcc)} | FA],
+
+    %% iterate completes without processing the last entries in the state.  Compute
+    %% the final visited segment, and add calls to the F([]) for all of the segments
+    %% that do not exist at the end of the file (due to deleting the last entry in the
+    %% segment).
+    Result = [{LeftSeg, F([])} || LeftSeg <- lists:reverse(LeftOver),
+				  LeftSeg =/= '*'] ++
+	[{LastSegment, F(LastAcc)} | FA],
     case Result of
         [{'*', _}] ->
             %% Handle wildcard select when all segments are empty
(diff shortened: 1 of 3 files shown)


JeetKunDoug:
@jonmeredith any way I can con you into completing the sentence "If keys in the last segments of the tree were deleted (but when there was another treeid following it in the same leveldb database), then the accumulator function being called on" in the PR description - seems to be missing something.

src/hashtree.erl
```diff
@@ -835,13 +915,18 @@ decode(Bin) ->
...
+    {TreeId, Level, Bucket}.
+
 -spec encode_meta(binary()) -> meta_bin().
 encode_meta(Key) ->
     <<$m,Key/binary>>.
 
 -spec hashes(hashtree(), list('*'|integer())) -> [{integer(), binary()}].
 hashes(State, Segments) ->
-    multi_select_segment(State, Segments, fun hash/1).
+     multi_select_segment(State, Segments, fun hash/1).
```
zeeshanlakhani: extra space? :D.

test/hashtree_eqc.erl
```diff
@@ -68,109 +114,265 @@ object() ->
...
-        [{1, {call, ?MODULE, reconcile, []}} || Started] ++
-        []
+      %% Update snapshots/trees. If memory is enabled must test with update_tree
+      %% If not, can use the method used by kv/yz_index_hashtree and separate
+      %% the two steps, dumping the result from update_perform.
+        [{10*SF, {call, ?MODULE, update_tree,  [t1, s1]}} || Snap1 == undefined] ++
+	[{10*SF, {call, ?MODULE, update_snapshot,  [t1, s1]}} || Snap1 == undefined, MemLevels == 0] ++
+	[{10*SF, {call, ?MODULE, update_perform,   [t1]}} || Snap1 == created, MemLevels == 0] ++
+	[{10*SF, {call, ?MODULE, set_next_rebuild, [t1]}} || Snap1 == updated] ++
+        [{10*SF, {call, ?MODULE, update_tree,  [t2, s2]}} || Snap1 == undefined] ++
```
nickelization: Should this check be `|| Snap2 == undefined`? I haven't actually worked through all the implications of whether this makes sense, but my gut feeling is that this might be a typo, just based on it not matching up with the pattern of how all the other calls in this list are structured.

jonmeredith: Well spotted - a copy pasta. Will fix.

src/hashtree.erl
```diff
@@ -966,13 +994,19 @@ iterate({ok, K, V}, IS=#itr_state{itr=Itr,
...
+
+            %% Do not update segment/final_acc
+            IS2 = IS#itr_state{current_segment=NextSeg,
+                               segment_acc=[],
+                               remaining_segments=Remaining,
                                final_acc=[{Segment, F(Acc)} | FinalAcc],
-                               prefetch=false},
-            iterate(iterator_move(Itr, prefetch_stop), IS2);
+                               prefetch=true}, % will be after second move
+            _ = iterator_move(Itr, prefetch_stop), % ignore the pre-fetch,
```
nickelization, line 1069: I think I understand everything in this PR except for this one bit here. In the old code, we recurse on the result of prefetch_stop in case it happens to drop us off at the beginning of the next interesting segment. Why can't we continue to do that? In the common case where we don't get to an interesting segment, I would think we'd just fall through to the prefetch=false clause below and then seek to the next segment from there.

The comment acknowledges we're risking a reseek, but doesn't explain why we need to take that risk. Maybe it's just not that big a risk though, and it's a cheap operation regardless, so the code is just simpler this way? But either way, I'd like to know what motivated the change.

jonmeredith, line 1069: I think the original code duplicated entry in the final_acc.  You could revert it to the original code and rerun EQC, if you get a failure, enable the debug logging and it should be obvious.

src/hashtree.erl
jonmeredith: Thank you for your contribution.  We'll be in touch.

src/hashtree.erl
```diff
@@ -958,21 +1058,19 @@ iterate({ok, K, V}, IS=#itr_state{itr=Itr,
...
                                prefetch=true},
             iterate(iterator_move(Itr, prefetch), IS2);
-        {Id, NextSeg, [NextSeg|Remaining], _} ->
-            %% A previous prefetch_stop left us at the start of the
-            %% next interesting segment.
+        {Id, _, [NextSeg | Remaining], true} ->
+            %% Pointing at uninteresting segment, but need to halt the
+            %% prefetch to ensure the iterator can be reused
+
+            %% Do not update segment/final_acc
```
nickelization: I think this comment should probably be removed, since now we do in fact update `segment` and `final_acc` here.

nickelization:
👍 c8efc3e

After much reviewing, testing, and playing with this code, I'm satisfied that I can't find any issues with it. Also, hopefully no issues with +1ing my own commit - everything else looks good, and since that last fix is just a trivial comment change (which I ran by Doug already), I'm bending the rules slightly and going ahead with it.

nickelization:
Bumping the eleveldb version has fixed the eunit and dialyzer errors we were hitting before. Merging...

nickelization:
@borshop merge

jadeallenx:
:+1: a83023e

nickelization:
@borshop merge

bashopatricia:
create jira issue

Basho-JIRA:
[~pbrewer] What was done here and why? Need information for documentation for e.g. 

**_[posted via JIRA by Seema Jethani]**_

Basho-JIRA:
[~nmarino] Can you provide some info on what was done for this?  Looks like you worked on it per GH.  [~seema][~lrother] This one appears it may have missed the release notes pull list for 2.0.7.  I found some stories yesterday that had the correct label of 2.0.7 but had "2.2" instead of "2.2.0" which means the story probably did not make the release notes starter list.  Lauren, if it helps, I can make a differences list from 2.0.7 to 2.2.0 as we get closer to release.   

**_[posted via JIRA by Patricia Brewer]**_

Basho-JIRA:
[~pbrewer], nothing else needs to be done here. The GitHub PR has already been reviewed and merged, but for some reason when you commented with "create jira issue" on the PR, the JIRA issue was created as "needs review". I'll go ahead and mark this as "done".

It looks like it actually was included in the release notes for 2.0.7. Search for #830 (RIAK-2526) (RIAK-2644): http://docs.basho.com/riak/kv/2.0.7/release-notes/ The summary given was:

"Several bugs were found with hash trees that, in rare cases, could cause AAE to fail to repair missing data."

This is a very brief, high-level summary of the issue, but I think that's about as good as it's going to get from an end-user perspective. The PR includes a few different fixes for very subtle edge cases in AAE, and they are hard to describe succinctly. I think it's good that we linked the PR in the release notes though, so that if someone does have the desire and know-how to understand the details, they can still click through and read about it in more detail. 

**_[posted via JIRA by Nick Marino]**_

Basho-JIRA:
Oh, and this has also been merged forward into to develop-2.2, so no worries there either. 

**_[posted via JIRA by Nick Marino]**_

PR #832, opened 2016-04-18
Modified riak_core_bucket:all_n to include n_vals from active bucket types

fadushin:
This PR modified the riak_core_bucket all_n function to include n_vals from active bucket types, not just buckets stored in the ring.  This function is used in riak_kv, and in some cases (e.g., riak_kv_ensemble), n_vals from bucket types are added.  However, in the calculation of the set of AAE trees, n_vals from bucket types is conspicuously missing, when in fact it is required.  This PR consolidates the logic of active n_val calculations in one place.

This PR is a pre-requision for `riak_kv` PR https://github.com/basho/riak_kv/pull/1399

### DIFF
--- a/src/riak_core_bucket.erl
+++ b/src/riak_core_bucket.erl
@@ -42,6 +42,11 @@
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
+-type riak_core_ring() :: riak_core_ring:riak_core_ring().
+-type bucket_type()  :: riak_core_bucket_type:bucket_type().
+-type bucket_props() :: [{term(), term()}].
+-type nval_set() :: ordsets:ordset(pos_integer()).
+
 -define(METADATA_PREFIX, {core, buckets}).
 
 %% @doc Add a list of defaults to global list of defaults for new
@@ -62,7 +67,7 @@ set_bucket({<<"default">>, Name}, BucketProps) ->
 set_bucket({Type, _Name}=Bucket, BucketProps0) ->
     case riak_core_bucket_type:get(Type) of
         undefined -> 
-            lager:error("Attempt to set properties of non-existant bucket type ~p", [Bucket]),
+            lager:error("Attempt to set properties of non-existent bucket type ~p", [Bucket]),
             {error, no_type};
         _ -> set_bucket(fun set_bucket_in_metadata/2, Bucket, BucketProps0)
     end;
@@ -125,7 +130,7 @@ get_bucket(Name) ->
     Meta = riak_core_ring_manager:get_bucket_meta(Name),
     get_bucket_props(Name, Meta).
 
-%% @spec get_bucket(Name, Ring::riak_core_ring:riak_core_ring()) ->
+%% @spec get_bucket(Name, Ring::riak_core_ring()) ->
 %%          BucketProps :: riak_core_bucketprops()
 %% @private
 get_bucket({<<"default">>, Name}, Ring) ->
@@ -166,7 +171,7 @@ reset_bucket(Bucket) ->
 
 %% @doc Get bucket properties `Props' for all the buckets in the given
 %%      `Ring' and stored in metadata
--spec get_buckets(riak_core_ring:riak_core_ring()) ->
+-spec get_buckets(riak_core_ring()) ->
                          Props::list().
 get_buckets(Ring) ->
     RingNames = riak_core_ring:get_buckets(Ring),
@@ -182,7 +187,7 @@ get_buckets(Ring) ->
     RingBuckets ++ MetadataBuckets.
 
 %% @doc returns a proplist containing all buckets and their respective N values
--spec bucket_nval_map(riak_core_ring:riak_core_ring()) -> [{binary(),integer()}].
+-spec bucket_nval_map(riak_core_ring()) -> [{binary(),integer()}].
 bucket_nval_map(Ring) ->
     [{riak_core_bucket:name(B), riak_core_bucket:n_val(B)} ||
         B <- riak_core_bucket:get_buckets(Ring)].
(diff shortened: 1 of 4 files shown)


src/riak_core_bucket.erl
```diff
@@ -192,14 +197,61 @@ bucket_nval_map(Ring) ->
...
 
+%% @private
+-spec bucket_type_nvals() -> nval_set().
+bucket_type_nvals() ->
+    riak_core_bucket_type:fold(fun bucket_type_prop_nval/2, ordsets:new()).
+
+%% @private
+-spec bucket_type_prop_nval({bucket_type(), bucket_props()}, nval_set()) ->
+    nval_set().
+bucket_type_prop_nval({_BType, BProps}, Accum) ->
```
JeetKunDoug: Would like to add `_fold` or `_folder` to `bucket_type_prop_nval` and `bucket_prop_nval` to make it clear they are used in a fold and not for more general use.

src/riak_core_bucket.erl
```diff
@@ -192,14 +197,61 @@ bucket_nval_map(Ring) ->
...
-                       end, [DefaultN], BucketNs),
+    BucketNVals = bucket_nvals(Ring),
+    BucketTypeNVals = bucket_type_nvals(),
+    ordsets:from_list(
+        ordsets:union(BucketNVals, BucketTypeNVals)
+    ).
+
+%% @private
+-spec bucket_nvals(riak_core_ring()) -> nval_set().
+bucket_nvals(Ring) ->
```
fadushin, line 210: Should this definition of bucket_nvals just take the above definition of all_n (or something close to it), to minimize differences?

src/riak_core_bucket.erl
```diff
@@ -42,6 +42,11 @@
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
+-type riak_core_ring() :: riak_core_ring:riak_core_ring().
+-type bucket_type()  :: binary().
```
nickelization: Instead of `binary()`, how about `riak_core_bucket_type:bucket_type()` (like what you did for `riak_core_ring()`).

(Would be nice if we could also do the same thing for `bucket_props()` but it looks like that type is currently defined in a riak_kv module, so I guess we'll have to define it separately since it's not a dependency.)

src/riak_core_bucket.erl
```diff
@@ -192,14 +197,61 @@ bucket_nval_map(Ring) ->
...
+-spec all_n(riak_core_ring()) -> [pos_integer(),...].
 all_n(Ring) ->
-    BucketNs = bucket_nval_map(Ring),
-    DefaultN = default_object_nval(),
-    AllN = lists:foldl(fun({_, N}, Acc) ->
-                               ordsets:add_element(N, Acc)
-                       end, [DefaultN], BucketNs),
+    BucketNVals = bucket_nvals(Ring),
+    BucketTypeNVals = bucket_type_nvals(),
+    ordsets:from_list(
```
nickelization: I think the `ordsets:from_list` call is redundant: that function takes an unordered list and returns an ordset (which is implemented as an ordered list). But `ordsets:union` should already be returning an ordset.

I suspect you might've meant to use the `ordsets:to_list` function, but that's also kind of not necessary since ordsets already have a defined implementation, which uses a list. Not really sure why the `to_list` API function exists, unless maybe it's just to make the type specs work out nicer or whatnot.

src/riak_core_bucket.erl
```diff
@@ -192,14 +197,61 @@ bucket_nval_map(Ring) ->
...
+            bucket_prop_nval(BProps, Accum);
+        _ ->
+            Accum
+    end.
+
+%% @private
+-spec bucket_prop_nval(bucket_props(), nval_set()) ->
+    nval_set().
+bucket_prop_nval(BProps, Accum) ->
+    case proplists:get_value(n_val, BProps) of
```
nickelization: Not sure if it's worth bothering to change, since this may not be a performance-critical path, but the `lists:keyfind` function is substantially faster than `proplists:get_value`. The functions in the `proplists` module have additional features that make them worthwhile sometimes, but for this case `lists:keyfind` has pretty much the same semantics anyway.

src/riak_core_bucket.erl
fadushin: Doug and I agreed that keeping the types aligned with ordsets was better than assuming the ordset(T) was defined at [T], but you're right, the return should be to_list.

JeetKunDoug: I do think to_list is, today at least, there to keep Dialyzer happy. However, it's certainly possible that someday Ericsson may change the internal representation, so I'd fix this to be to_list, which is essentially a no-op:

From https://github.com/erlang/otp/blob/maint/lib/stdlib/src/ordsets.erl#L70:

``` erlang
to_list(S) -> S.
```

src/riak_core_bucket.erl
```diff
@@ -192,14 +197,61 @@ bucket_nval_map(Ring) ->
 default_object_nval() ->
     riak_core_bucket:n_val(riak_core_bucket_props:defaults()).
 
+%% @private
+%% @doc For testing/debugging
+-spec all_n() -> [pos_integer(),...].
+all_n() ->
```
fadushin: I just realized I am opposed to this gratuitous function.  I will just call riak_core_ring_manager:get_my_ring() in the EQC test code.

nickelization:
👍 2904453

src/riak_core_bucket.erl
```diff
@@ -192,19 +197,58 @@ bucket_nval_map(Ring) ->
...
+    case lists:keyfind(n_val, 1, BProps) of
+        false ->
+            Accum;
+        {n_val, NVal} ->
+            ordsets:add_element(NVal, Accum)
+    end.
+
 name(BProps) ->
-    proplists:get_value(name, BProps).
+    {name, Name} = lists:keyfind(name, 1, BProps),
```
JeetKunDoug: Fred pointed out that these lists:keyfind replacements will fail if somehow the key (name or n_val) isn't in the list - with proplists:get_value it would just return `undefined` but not lists:keyfind will return `false` and the pattern match here won't work. There was also a concern that bucket props may actually _be_ proplists, including the strange cases of bare atoms in the list, but this appears to not be the case, as riak_core_bucket_props uses all kinds of lists:\* functions that will fail if a bare atom is in the list.

JeetKunDoug:
:+1: 8ecc355

fadushin:
@borshop merge

PR #834, opened 2016-04-26
Bugfix/listkeys filters

erikleitch:
**Context**

stream_list_keys has always returned inconsistent results when used with TS data.  I've tracked this down to the coverage filter functions that are applied during the fold over keys on a vnode.  These filters check whether a key should be included, by checking if its hash falls within a specified hash range for a vnode.  These filters are applied to the storage keys in leveldb.  For TS, however, these filters need to be applied not to the storage key (which is the local key), but instead to the partition key, which can be derived from the local key by an appropriate transformation.

This PR addresses this inconsistency by changing the filter functions to use a key conversion function on the storage key prior to hashing. It is one of three PRs: riak_kv (https://github.com/basho/riak_kv/pull/1404), riak_core (https://github.com/basho/riak_core/pull/834) and riak_pipe (https://github.com/basho/riak_pipe/pull/105).  For reference, I include my diagram of the listkeys code path here:

 RiakTS Listkeys Path

**Changes in this repo**
- **riak_core_coverage_fsm.erl**<br>
  - record 'state', field 'coverage_plan_mod' has been changed  to 'coverage_plan_fn'.  The previous version of this code used coverage_plan_mod to return the name of a module whose static function create_plan/6 function would be used to construct the coverage plan.  This PR uses coverage_plan_fn to return a dynamically constructed function that creates the coverage plan, using the key conversion function specific to a TS table.
  - riak_kv_keys_fsm:init(), called in riak_core_coverage_fsm:init(), now
    returns coverage_plan_fn instead of a coverage_plan_mod
  - find_plan changed to use CoveragePlanFn directly rather than
    Mod:create_plan

### DIFF
--- a/src/riak_core_coverage_fsm.erl
+++ b/src/riak_core_coverage_fsm.erl
@@ -123,7 +123,7 @@ behaviour_info(_) ->
                 vnode_master :: atom(),
                 plan_fun :: function(),
                 process_fun :: function(),
-                coverage_plan_mod :: module()
+                coverage_plan_fn :: function()
                }).
 
 %% ===================================================================
@@ -165,7 +165,7 @@ init([Mod,
       From={_, ReqId, _},
       RequestArgs]) ->
     {Request, VNodeSelector, NVal, PrimaryVNodeCoverage,
-     NodeCheckService, VNodeMaster, Timeout, PlannerMod, ModState} =
+     NodeCheckService, VNodeMaster, Timeout, CoveragePlanFn, ModState} =
         Mod:init(From, RequestArgs),
     maybe_start_timeout_timer(Timeout),
     PlanFun = plan_callback(Mod),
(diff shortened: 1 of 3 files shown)


gordonguthrie:
This is a tentative "I'm not sure what I know" comment

This is an API breaker in riak_core

Our three backends, leveldb, bitcask and memory all expose themselves to riak_core via riak_kv, so the API change _should_ work for all of them (or do we need to check that?)

How do we announce API changes in riak_core for open source devs that use it for non-KV apps?

gordonguthrie:
More tentative comments - is this change going to work with Yokazuna?

https://github.com/basho/yokozuna/blob/develop/src/yz_cover.erl#L159

hmmr:
+1 babeef2

hmmr:
@borshop merge

PR #836, opened 2016-05-02
Fix for get_value

fadushin:
There was a lingering pattern match against a specific atom in the get_value function which was accidentally merged.


jvoegele:
+1 f55a8fb

fadushin:
@borshop merge

PR #837, opened 2016-05-05
Export bucket properties type for use in other modules/apps

JeetKunDoug:
We have (badly) duplicated bucket property type specs across Riak - these were needed in Yokozuna so clean up/export them here. Later work should revisit other places where we use bucket properties but have not specified them well.

Prerequisite for basho/yokozuna#638.

### DIFF
--- a/src/riak_core_bucket.erl
+++ b/src/riak_core_bucket.erl
@@ -21,31 +21,34 @@
 %% -------------------------------------------------------------------
 
 %% @doc Functions for manipulating bucket properties.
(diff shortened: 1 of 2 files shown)


jadeallenx:
Love this. ❤️

fadushin:
@JeetKunDoug Squash 3 commits into one?  Cleanliness next to godliness, and all that jazz?

fadushin:
+1 9f3cfbf

fadushin:
@borshop merge

PR #841, opened 2016-06-02
Correct management of AAE iterators [JIRA: RIAK-3059]

matthewvon:
This branch changes AAE's management of leveldb iterators.  Previously, the AAE code would implicitly close a previous iterator only when needing a fresh iterator.  The act of leaving the previous iterator open causes leveldb to not purge retired files held by the previous iterator.  This leads to disk space bloat (and slightly slows down the operation of the fresh iterator since its retrievals can occur during large file delete operations).


bashopatricia:
create jira issue

PR #847, opened 2016-07-19
Add support for setting/querying a cluster-wide time zone

macintux:
riak/riak_ee PRs to come. Expect to change `rebar.config` once my repo is migrated to Basho's ownership and once the API has stabilized enough to lay down a tag. I think it's reasonable to review/approve these PRs now in the expectation of more to arrive before the next TS release, but that'll be a topic for discussion, I'm sure.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -18,5 +18,6 @@
   {pbkdf2, ".*", {git, "git://github.com/basho/erlang-pbkdf2.git", {tag, "2.0.0"}}},
   {eleveldb, ".*", {git, "git://github.com/basho/eleveldb.git", {tag, "ts-1.0.4"}}},
   {exometer_core, ".*", {git, "git://github.com/basho/exometer_core.git", {tag, "1.0.0-basho6"}}},
+  {jam, ".*", {git, "git://github.com/macintux/jam.git", {branch, "master"}}},
   {clique, "0.3.5", {git, "git://github.com/basho/clique.git", {tag, "0.3.5"}}}
 ]}.
--- a/rebar.config
+++ b/rebar.config
@@ -18,6 +18,6 @@
   {pbkdf2, ".*", {git, "git://github.com/basho/erlang-pbkdf2.git", {tag, "2.0.0"}}},
   {eleveldb, ".*", {git, "git://github.com/basho/eleveldb.git", {tag, "ts-1.0.4"}}},
   {exometer_core, ".*", {git, "git://github.com/basho/exometer_core.git", {tag, "1.0.0-basho6"}}},
-  {jam, ".*", {git, "git://github.com/macintux/jam.git", {branch, "master"}}},
+  {jam, ".*", {git, "git://github.com/basho/jam.git", {branch, "develop"}}},
   {clique, "0.3.5", {git, "git://github.com/basho/clique.git", {tag, "0.3.5"}}}
 ]}.
--- a/src/riak_core_console.erl
+++ b/src/riak_core_console.erl
@@ -31,7 +31,8 @@
          print_users/1, print_user/1, print_sources/1,
          print_groups/1, print_group/1, print_grants/1,
          security_enable/1, security_disable/1, security_status/1, ciphers/1,
-	 stat_show/1, stat_info/1, stat_enable/1, stat_disable/1, stat_reset/1]).
+         timezone/1,
+         stat_show/1, stat_info/1, stat_enable/1, stat_disable/1, stat_reset/1]).
 
 %% New CLI API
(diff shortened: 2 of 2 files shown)


src/riak_core_console.erl
```diff
@@ -752,6 +753,27 @@ commit_staged([]) ->
                       "plan and try to commit again~n")
     end.
 
+timezone([]) ->
+    TZ = riak_core_metadata:get({riak_core, timezone}, string,
+                                [{default, "not configured"}]),
+    io:format("~s~n", [TZ]);
```
hazen, line 759: Do we want these to be `lager:info` or `lager:debug` maybe?

hazen, line 759: Guessing we want to explicitly return the `TZ` instead of the string?

macintux, line 759: Nope, the console code uses `io:format` consistently.

src/riak_core_console.erl
```diff
@@ -752,6 +753,27 @@ commit_staged([]) ->
                       "plan and try to commit again~n")
     end.
 
+timezone([]) ->
```
hazen, line 756: A `spec` would make me happy so I would know what we are expecting.

macintux, line 756: Ok. We're not returning anything from any of these (at least, we're not returning anything anyone cares about). riak-admin calls into this module with the expectation that the module will print results for the user to read.

macintux:
Failed to notice dialyzer is unhappy, will look at that. Build log doesn't show anything useful.

src/riak_core_console.erl
hazen, line 759: Really? Seems like there might be a cleaner way, but OK.

rebar.config
```diff
@@ -18,5 +18,6 @@
   {pbkdf2, ".*", {git, "git://github.com/basho/erlang-pbkdf2.git", {tag, "2.0.0"}}},
   {eleveldb, ".*", {git, "git://github.com/basho/eleveldb.git", {tag, "ts-1.0.4"}}},
   {exometer_core, ".*", {git, "git://github.com/basho/exometer_core.git", {tag, "1.0.0-basho6"}}},
+  {jam, ".*", {git, "git://github.com/macintux/jam.git", {branch, "master"}}},
```
hmmr: Shouldn't jam be under basho?

src/riak_core_console.erl
```diff
@@ -752,6 +753,27 @@ commit_staged([]) ->
...
+    TZ = riak_core_metadata:get({riak_core, timezone}, string,
+                                [{default, "not configured"}]),
+    io:format("~s~n", [TZ]);
+timezone([TZString]) ->
+    TZ = jam:process(jam_iso8601:parse_tz(TZString)),
+    attempt_set_timezone(jam:is_valid(TZ), TZ, TZString).
```
hmmr, line 762: I would rather have `jam_iso8601:parse_tz/1` either return the standard `{ok, TZ} | {error, Reason}`, or return `TZ` and throw a `badarg` if it's fed garbage. This way, you would avoid the need to have `TZ` validated separately and pass the validation result as a separate arg to `attempt_set_timezone` (and incidentally, dialyzer would produce a clearer message then).

rebar.config
hazen: @macintux mentioned this in HipChat.  We'll try to get the repo copied over to Basho, but not in this PR.

src/riak_core_console.erl
hazen, line 762: That would crash the console, right? That might not be the behavior we want.  It would be nice from a function-perspective, however.

hazen:
The good news is that indeed dialyzer is working

src/riak_core_console.erl
macintux, line 762: Because `jam` is designed around pipeline processing, it's very convenient to return the result directly instead of via tuples. The few cases where I do return tuples (see the README) I'd like to find a cleaner way to handle.

I'm not saying no, but neither am I saying yes.

hazen:
:+1: b33c63a

macintux:
DO NOT merge this yet. `jam` is still undergoing some breaking changes.

hazen:
@macintux I wasn't sure it you wanted to get that in first.  I'm easy.

macintux:
This is now ready for final +1

hmmr:
+1 bc3b903

macintux:
@borshop merge

macintux:
Bors missed this, concurrent with some github instability. YOLOmerging

PR #848, opened 2016-07-21
Filter local capabilities when no longer in ring

andytill:
`riak_core_capability` tracks the supported modes of capabilities for each clustered node in the `#state.supported` record field.  When the ring changes, new capabilities are added or updated. Previous to this change, capabilities that had been removed in the ring were not removed from the local copy, affecting the final negotiated capability value.  This can happen in the following scenario:
1. In a three node cluster, all nodes are version 2 and declare capability `my_cap` supported modes as `[b,a]` and a default of `a`.  Because `b` is preferred (first in the list) and all nodes specify the same modes then `b` is the negotiated capability.
2. One of the nodes is downgraded to version 1, where `my_cap` is not registered at all and doesn't appear in the nodes capabilities, or in the metadata it puts into the ring that is received by the other nodes.

With this change, when capabilities are added for a node, they must also appear in the ring capabilities for that node or be removed.

### DIFF
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -373,7 +373,7 @@ add_node2(Node, Capabilities, State) ->
 %% is not declared. We must remove it from our local copy of its supported modes
 %% so that the default value can be used.
 filter_removed_caps(Node, Capabilities, State1) ->
-    {NodeSupported, _Removed} = lists:partition(
+    NodeSupported = lists:filter(
         fun({CapName, _}) ->
             lists:keymember(CapName, 1, Capabilities)
         end, get_supported(Node, State1)),
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -349,7 +349,11 @@ remove_members(Ring, State=#state{supported=Supported}) ->
 %% to determine capabilities through RPC to the node. If RPC fails, use
 %% default values. However, unresolved nodes will be marked as such and RPC
 %% re-attempted at the next server tick.
-add_node(Node, [], State=#state{unknown=Unknown}) ->
+add_node(Node, Capabilities, State1) ->
+    State2 = add_node2(Node, Capabilities, State1),
+    filter_removed_caps(Node, Capabilities, State2).
+
+add_node2(Node, [], State=#state{unknown=Unknown}) ->
     {Capabilities, Resolved} = query_capabilities(Node, State),
     Unknown2 = case Resolved of
                    true ->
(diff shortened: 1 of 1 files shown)


src/riak_core_capability.erl
```diff
@@ -359,9 +363,23 @@ add_node(Node, [], State=#state{unknown=Unknown}) ->
...
+add_node2(Node, Capabilities, State) ->
     add_node_capabilities(Node, Capabilities, State).
 
+%% Remove supported capabilities for a node where it no longer appears in the
+%% rings capabilities for the node.
+%%
+%% This can occur when another node downgrades to a version where the capability
+%% is not declared. We must remove it from our local copy of its supported modes
+%% so that the default value can be used.
+filter_removed_caps(Node, Capabilities, State1) ->
```
hazen, line 375: Could we get a unit test for this?

src/riak_core_capability.erl
```diff
@@ -349,7 +349,11 @@ remove_members(Ring, State=#state{supported=Supported}) ->
 %% to determine capabilities through RPC to the node. If RPC fails, use
 %% default values. However, unresolved nodes will be marked as such and RPC
 %% re-attempted at the next server tick.
-add_node(Node, [], State=#state{unknown=Unknown}) ->
+add_node(Node, Capabilities, State1) ->
```
hazen, line 352: Where does the incoming list of capabilities come from?

andytill, line 352: The ring metadata.

cmeiklejohn: tumblr_lziq98lmec1qzamtlo1_500

hazen:
@JeetKunDoug Please let us know what your thoughts are before we merge this

src/riak_core_capability.erl
```diff
@@ -359,9 +363,23 @@ add_node(Node, [], State=#state{unknown=Unknown}) ->
...
     add_node_capabilities(Node, Capabilities, State).
 
+%% Remove supported capabilities for a node where it no longer appears in the
+%% rings capabilities for the node.
+%%
+%% This can occur when another node downgrades to a version where the capability
+%% is not declared. We must remove it from our local copy of its supported modes
+%% so that the default value can be used.
+filter_removed_caps(Node, Capabilities, State1) ->
+    {NodeSupported, _Removed} = lists:partition(
```
JeetKunDoug: Since you're just throwing away _Removed, please use `lists:filter` instead of `lists:partition`.

andytill: A hangover from some logging I was doing, fixed in a0df72818dc3ebd8adb6997467197fefb69ff5ed.

JeetKunDoug:
Other than @javajolt 's comment about a unit test for `filter_removed_caps` I'm +1 for this change. Nice find, and nice work, @andytill.

src/riak_core_capability.erl
andytill, line 375: This is tested by https://github.com/basho/riak_test/pull/1112

gordonguthrie:
+1 good job

JeetKunDoug:
@andytill realize it's tested by a riak_test, but had someone bothered to write some unit tests for riak_core_capability in the first place, we may not have this bug. I realize you are close to release, so I'm not going to hound you for it, but if you get 20 minutes and can at least write some simple tests around this particular change, it would be greatly appreciated. If not, please follow up with me and get it scheduled.

gordonguthrie:
JIRA raised re @JeetKunDoug 's last comment on unit tests

https://bashoeng.atlassian.net/browse/RIAK-2731

gordonguthrie:
+1 a0df728

PR #849, opened 2016-07-22
Add new riak_core_throttle module for reusable throttling behavior

jvoegele:
https://bashoeng.atlassian.net/browse/RIAK-2626

As per RIAK-2626, we needed throttling of Yokozuna's AAE repairs to prevent overload due to the large number of key exchanges that will result from the new hashing algorithm. KV already had throttling behavior built into its AAE subsystem, but this did not apply to Yokozuna AAE. Therefore, we've extracted the KV throttling logic into the new `riak_core_throttle` module, where it can be used by both KV and Yokozuna (as well as any other subsystems that might need throttling behavior in the future).

The degree of throttling needed for any particular activity (such as Yokozuna AAE repair) is based on the current amount of load that the system is under. Since different subsystems have differing notions of "load", the `riak_core_throttle` module leaves the load calculation to the client code, but it allows for mapping of that load factor to  a specific time delay for that level of load. This is done by defining a "limits" mapping for an activity. See the `riak_core_throttle:set_limits/2` and `riak_core_throttle:set_throttle_by_load/2` functions for details.

Also note that throttling behavior can still be externally configured with Cuttlefish (including disabling throttling altogether) by virtue of the `riak_core_throttle:init/4` function.

### DIFF
--- a/src/riak_core_throttle.erl
+++ b/src/riak_core_throttle.erl
@@ -1,0 +1,358 @@
+%% -------------------------------------------------------------------
+%% Copyright (c) 2016 Basho Technologies, Inc.  All Rights Reserved.
+%%
+%% This file is provided to you under the Apache License,
+%% Version 2.0 (the "License"); you may not use this file
+%% except in compliance with the License.  You may obtain
+%% a copy of the License at
+%%
+%%   http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing,
+%% software distributed under the License is distributed on an
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+%%
+%% @doc Throttling for activities that can potentially overload resources.
+%% Provides support for "throttling" (i.e. slowing down) calling processes
+%% that, when under heavy load, could potentially overload other resources.
+%% Activities are identified by a unique key so that different throttle values
+%% can be used for different activities.
+%%
+%% The throttle for an activity can be set to a specific value with the
+%% `set_throttle/2' function, or it can be set to a value based on the current
+%% "load" for the activity by using the `set_limits/2' function to define a
+%% mapping from load factors to throttle values, and then periodically calling
+%% the `set_throttle_by_load/2' function with the current load factor. Note
+%% that "load factor" is an abstract concept, the purpose of which is simply
+%% to represent numerically the amount of load that an activity is under and
+%% map that level of load to a particular throttle level. Examples of metrics
+%% that could be used for load factors include the size of the mailbox for a
+%% particular process, or the number of messages per second being sent to an
+%% external service such as Solr.
+-module(riak_core_throttle).
+
+%% API
+-export([get_throttle/2,
+         set_throttle/3,
+         clear_throttle/2,
+         disable_throttle/2,
+         enable_throttle/2,
+         is_throttle_enabled/2,
+         init/4,
+         set_limits/3,
+         get_limits/2,
+         clear_limits/2,
+         create_limits_translator_fun/2,
+         set_throttle_by_load/3,
+         throttle/2]).
+
+-ifdef(TEST).
+-export([get_throttle_for_load/3]).
+-define(SLEEP(Time), Time).
+-else.
+-define(SLEEP(Time), timer:sleep(Time)).
+-endif.
+
+-type app_name() :: atom().
+-type activity_key() :: atom().
+-type throttle_time() :: non_neg_integer().
+-type load_factor() :: number() | atom() | tuple().
+-type limits() :: [{load_factor(), throttle_time()}].
+
+-export_type([activity_key/0, throttle_time/0, load_factor/0, limits/0]).
+
+-define(THROTTLE_KEY(Key), list_to_atom("throttle_" ++ atom_to_list(Key))).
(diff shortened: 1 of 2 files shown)


src/riak_core_throttle.erl
```diff
@@ -110,6 +156,38 @@ get_limits(Key) ->
...
 
+%% @doc Returns a fun that used in Cuttlefish translations to translate from
+%% configuration items of the form:
+%%  `ConfigPrefix'.throttle.tier1.`LoadFactorMeasure'
+%% to the list of tuples form expected by the `set_limits/2' function in this
+%% module. See riak_kv.schema and yokozuna.schema for example usages.
+create_limits_translator_fun(ConfigPrefix, LoadFactorMeasure) ->
+    fun(Conf) ->
+            %% Grab all of the possible names of tiers so we can ensure that
+            %% both mailbox_size and delay are included for each tier.
```
JeetKunDoug: Should change the comment to no longer include `mailbox_size` as that's not always the name... should be `LoadFactorMeasure` now IIRC.

src/riak_core_throttle.erl
```diff
@@ -0,0 +1,301 @@
...
+            TierNamesM = cuttlefish_variable:fuzzy_matches([ConfigPrefix, "throttle", "$tier", LoadFactorMeasure], Conf),
```
JeetKunDoug: Would perhaps rename `TierNamesM` to `LoadFactorTierNames` and `TierNameD` to `DelayTierNames` - both corrects the "M=Mailbox" issue as in the comment above, but also makes it easier to see what you're pulling out of the config at this point.

src/riak_core_throttle.erl
```diff
@@ -0,0 +1,301 @@
...
+            TierNamesD = cuttlefish_variable:fuzzy_matches([ConfigPrefix, "throttle", "$tier", "delay"], Conf),
+            TierNames = lists:usort(TierNamesM ++ TierNamesD),
+            Throttles = lists:sort(lists:foldl(
+                                     fun({"$tier", Tier}, Settings) ->
+                                             Mbox = cuttlefish:conf_get([ConfigPrefix, "throttle", Tier, LoadFactorMeasure], Conf),
```
JeetKunDoug: Again, rename MBox to LoadFactorValue or something along those lines

src/riak_core_throttle.erl
```diff
@@ -0,0 +1,301 @@
...
+-define(THROTTLE_KEY(Key), list_to_atom("throttle_" ++ atom_to_list(Key))).
+-define(THROTTLE_LIMITS_KEY(Key),
+        list_to_atom("throttle_limits_" ++ atom_to_list(Key))).
+-define(THROTTLE_KILL_KEY(Key),
+        list_to_atom("throttle_kill_" ++ atom_to_list(Key))).
+
+%% @doc Sets the throttle for the activity identified by `Key' to the specified
+%% `Time'.
+-spec set_throttle(activity_key(), throttle_time()) -> ok.
+set_throttle(Key, Time) when Time >= 0 ->
```
JeetKunDoug: As discussed, need to add Application Name to all of the set/get/etc. functions so different apps can coexist without fear of overlapping activity key names.

src/riak_core_throttle.erl
```diff
@@ -0,0 +1,301 @@
...
+-type throttle_time() :: non_neg_integer().
+-type load_factor() :: number() | atom() | tuple().
+-type limits() :: [{load_factor(), throttle_time()}].
+
+-export_type([activity_key/0, throttle_time/0, load_factor/0, limits/0]).
+
+-define(THROTTLE_KEY(Key), list_to_atom("throttle_" ++ atom_to_list(Key))).
+-define(THROTTLE_LIMITS_KEY(Key),
+        list_to_atom("throttle_limits_" ++ atom_to_list(Key))).
+-define(THROTTLE_KILL_KEY(Key),
```
JeetKunDoug: As discussed, we'll switch this from `THROTTLE_KILL_KEY` to `THROTTLE_ENABLED_KEY` and all the other things to not make this whole thing backwards, once everything else is working and tested.

src/riak_core_throttle.erl
```diff
@@ -0,0 +1,301 @@
...
+    not Killed.
+
+%% @doc Initializes the throttle for the activity identified by `Key' using
+%% configuration values found in the application environment for `AppName'.
+-spec init(Key::activity_key(),
+           AppName::atom(),
+           {LimitsKey::atom(), LimitsDefault::limits()},
+           {KillSwitchKey::atom(), KillSwitchDefault::boolean()}) ->
+    ok | {error, Reason :: term()}.
+init(Key,
```
JeetKunDoug: Switch order of Key and AppName as we're going to pass AppName everywhere else, and it has really become a 2-level hierarchy.

src/riak_core_throttle.erl
```diff
@@ -0,0 +1,356 @@
...
+        [],
+        TierNames)).
+
+%% @doc <p>Sets the throttle for the activity identified by `AppName' and `Key'
+%% to a value determined by consulting the limits for the activity. The
+%% throttle value is the value associated with the largest
+%% load factor <= `LoadFactor'.
+%% Normally the `LoadFactor' will be a number representing the current level
+%% of load for the activity, but it is also allowed to pass an atom or a tuple
+%% as the %% `LoadFactor'. This can be used when a numeric value cannot be
```
JeetKunDoug: formatting issue - remove %% from middle of comment.

jvoegele:
The refactor of `find_throttle_for_load_factor/2` looks good to me.

JeetKunDoug:
+1 999a73f

JeetKunDoug:
@borshop merge

PR #851, opened 2016-08-08
Support for async job switches

tburghart:
Adds support for async job enable/disable switches in other packages.
These changes don't affect the behavior of riak_core itself at all, they just centralize schema and runtime handling of the switches.

_Caveats:_
- RIAK-2744 added to address what's _not_ in this PR.

### DIFF
--- a/priv/riak_core.schema
+++ b/priv/riak_core.schema
@@ -246,7 +246,7 @@
             [{erlang:list_to_atom(Mod), erlang:list_to_atom(Op)} | Result];
         ({[_, _, _, _], false}, Result) ->
             Result;
-        ([_, _, _, _] = Setting, _) ->
+        ({[_, _, _, _], _} = Setting, _) ->
             cuttlefish:invalid(io_lib:format("~p", [Setting]));
         (_, Result) ->
             Result
--- a/priv/riak_core.schema
+++ b/priv/riak_core.schema
@@ -83,7 +83,7 @@
 {validator,
   "valid_ipaddr",
   "must be a valid IP address",
-  fun(AddrString) -> 
+  fun(AddrString) ->
     case inet_parse:address(AddrString) of
       {ok, _} -> true;
       {error, _} -> false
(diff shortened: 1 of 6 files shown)


src/riak_core_util.erl
```diff
@@ -704,6 +712,124 @@ proxy(Parent, Fun) ->
...
+%% WARNING:
+%% * This function is likely to be extended to accept a Job as well as a Class
+%%   when the Jobs API is live.
+job_class_disabled_message(binary, Class) ->
+    erlang:list_to_binary(job_class_disabled_message(text, Class));
+job_class_disabled_message(text, Class) ->
+    lists:flatten(io_lib:format("Operation '~p' is not enabled", [Class])).
+
+-spec report_job_request_disposition(
+    boolean(), term(), module(), atom(), pos_integer(), term())
```
JeetKunDoug: The meaning of parameters (especially the true/false in the function heads below) would be much easier to figure out if we used parameter names in this spec, rather than just types... maybe something like

``` erlang
-spec report_job_request_disposition(
    JobAccepted :: boolean(), 
    JobClass :: term(), 
    CallingModule :: module(), 
    CallingFunction :: atom(), 
    CallingLine :: pos_integer(), 
    Client :: term()), %% Still not exactly sure what "Client" is - ClientIdentifier? "HTTP/PB"?
) -> ok | {error, term()}.
```

src/riak_core_util.erl
```diff
@@ -704,6 +712,124 @@ proxy(Parent, Fun) ->
...
+job_class_enabled(Application, Operation) ->
+    {error, {badarg, {Application, Operation}}}.
+
+-spec enable_job_class(term()) -> ok | {error, term()}.
+%% @doc Internal API to enable the specified job class.
+%% WARNING:
+%% * This function may not remain in this form once the Jobs API is live!
+%% * Parameter types ARE NOT validated by the same rules as the public API!
+%% You are STRONGLY advised to use enable_job_class/2.
+enable_job_class(Class) ->
```
JeetKunDoug, line 752: I'm fairly certain both enable_job_class/1 and disable_job_class/1 contain race conditions (read, check, write with no way to do a "transaction" across the environment variable).

We may have to split the environment variable out into separate entries for each Application/Operation pair at this point? Do something similar to riak_core_throttle, which would allow you to just enable/disable an operation directly (which, admittedly, may still have multiple processes writing to it, but you wouldn't end up with one setting simply not being set at all because another overwrote the list after you did).

I'm not sure if we really _need_ a _list_ of job classes somewhere in your future design, but if we really do, we may have to punt and manage our own ETS table here to fix the overwrite race condition, where the table would have one row per application/operation pair.

tburghart, line 752: After discussion, we've decided to leave the behavior as-is for 2.2.

Issue RIAK-2744 has been entered to make sure this is addressed if or when we may drive these interfaces programmatically.

src/riak_core_util.erl
tburghart: Addressed in commit d14b78a

priv/riak_core.schema
```diff
@@ -200,3 +201,56 @@
     {datatype, {duration, ms}},
     hidden
 ]}.
+
```
JeetKunDoug, line 204: The schema elements should be tested in `riak_core_schema_tests.erl.` You may find some useful examples in the riak_kv multibackend schema tests.

src/riak_core_util.erl
```diff
@@ -704,6 +712,141 @@ proxy(Parent, Fun) ->
...
+        [{pid, erlang:self()}, {module, Mod}, {function, Func}, {line, Line}],
+        "Request '~p' accepted from ~p", [Class, Client]);
+report_job_request_disposition(false, Class, Mod, Func, Line, Client) ->
+    lager:log(warning,
+        [{pid, erlang:self()}, {module, Mod}, {function, Func}, {line, Line}],
+        "Request '~p' disabled from ~p", [Class, Client]).
+
 %% ===================================================================
 %% EUnit tests
 %% ===================================================================
```
JeetKunDoug, line 852: Would be good to have some simple unit tests around the enable/disable/enabled/disabled functions.

priv/riak_core.schema
tburghart, line 204: There are several tests that _could_ be added, but I elected not to because riak_core itself has no schema mappings, only the single cuttlefish translation, which is more effectively tested in other packages' unit tests and in riak_test system-level tests.

src/riak_core_util.erl
tburghart, line 852: As above, those operations are better tested it the system level in riak_test, where full application state is in place and the functions can be invoked via RPC as intended.

JeetKunDoug:
In the interest of time, and getting many other PRs merged, will +1 this for now w/ only systems tests. Need to revisit in 2.3 timeline and get things under eunit tests.

JeetKunDoug:
+1 d14b78a

tburghart:
@borshop merge

PR #852, opened 2016-08-09
Capability tests, Legacy Mapping Fix RIAK-2731

andytill:
Unit tests for `riak_core_capability.erl`.

This uncovered a bug where all capabilities would be filtered if they were not in the ring metadata, even if they came from querying the remote nodes OTP application environment.  The filtering has been moved to the case where the capabilities come from the metadata.

To make this testable with eunit, the side effecting rpc call has been wrapped in a fun which is stored in the state record. This makes it possible to stub out the rpc call and return canned values to exercise paths for the test.

### DIFF
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -103,7 +103,11 @@
                 last_ring_id :: term(),
                 supported :: [{node(), [{capability(), [mode()]}]}],
                 unknown :: [node()],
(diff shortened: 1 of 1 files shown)


hazen:
FWIW, Eunit tests pass and dialyzer is happy

hazen:
Gonna merge this so we can add to the testing cycle

hazen:
:+1: 3eae6b5

PR #853, opened 2016-08-09
Remove complexity of storing a function in state and passing through function calls.

JeetKunDoug:
By using meck in tests, we simplify production code, debugging, and replace
a pattern we don't use in Riak with a well-known testing methodology.


hazen:
Corresponding riak_test, `verify_removed_capability` passes

hazen:
:+1: 48c2635

hazen:
@borshop merge

PR #857, opened 2016-09-07
Bumped eleveldb dependency to 2.0.28

fadushin:


nickelization:
+1 e3485a6

nickelization:
Giving a +1 despite the failing checks, because Fred says the failures were caused by creating the PR before eleveldb was tagged. Bors should do it's normal integration testing, so we'll wait for that to pass before we merge.

fadushin:
@borshop merge

PR #858, opened 2016-09-15
Reverted cherry-pick of capabilities removal on downgrade

fadushin:
This change is causing issues with rolling restarts, so we are reverting it until we can come up with a better strategy to handle capabilities with mixed clusters on downgrade.


JeetKunDoug:
+1 f3d833c

fadushin:
@borshop merge

hazen:
@andytill FYI

PR #861, opened 2016-09-20
Use ETS table instead of application environment variables in riak_core_throttle

jvoegele:
Previously, `riak_core_throttle` had used application environment variables for maintaining its state. However, due to race conditions that manifested during application shutdown, errors would occur that prevented applications from shutting down cleanly. By replacing the use of application environment variables with ETS tables instead, we avoid these errors and shutdown cleanly.

Because `riak_core_throttle` is not itself a process, it cannot own its underlying ETS table. Thus, the `riak_core_table_owner` gen_server process was introduced for this purpose.


fadushin:
+1 4067bc4

hazen:
@borshop merge

PR #875, opened 2016-11-17
Support finding a user by their associated metadata

paulhenrich:


paulhenrich:
Related to:

- https://github.com/basho/riak_test/pull/1199
- https://github.com/basho/riak_s3_api/pull/26

src/riak_core_security.erl
```diff
@@ -82,6 +98,28 @@
...
 -type userlist() :: all | [string()].
+-type username() :: string().
+-type metadata_key() :: string().
+-type metadata_value() :: term().
+-type options() :: [{metadata_key(), metadata_value()}].
+
+-spec find_by_metadata(metadata_key(), metadata_value()) -> [{username(), options()}].
+find_by_metadata(Key, Value) ->
+    riak_core_metadata:fold(
+      fun({_Username, [?TOMBSTONE]}, Acc) ->
```
paulhenrich: Rejecting tombstones seems pretty common in this file. We just copied and pasted it, but I'd like to refactor it into a named function.

src/riak_core_security.erl
```diff
@@ -82,6 +98,28 @@
 -type bucket() :: {binary(), binary()} | binary().
 -type permission() :: {string()} | {string(), bucket()}.
 -type userlist() :: all | [string()].
+-type username() :: string().
+-type metadata_key() :: string().
+-type metadata_value() :: term().
+-type options() :: [{metadata_key(), metadata_value()}].
```
hazen, line 103: Love all the new types! Keep up the great work. 👍

src/riak_core_security.erl
JeetKunDoug: It would be good if we could extract this multi-headed anonymous function into a named fun - would have to have a very small anonymous fun here for the closure over {Key, Value} but otherwise just call, what, `include_user_with_option({_UserName, [_Options]} = Data, Key, Value, Acc)` or something like that?

src/riak_core_security.erl
```diff
@@ -82,6 +98,28 @@
...
+
+-spec find_by_metadata(metadata_key(), metadata_value()) -> [{username(), options()}].
+find_by_metadata(Key, Value) ->
```
JeetKunDoug: This is specifically `find_user_by_metadata` - we either need to make it generic (pass in a `type` for users/groups/grants and maybe a matching function?) or rename this to `find_user_by_metadata` - I think the latter is probably appropriate as the former will make this function fairly complex.

JeetKunDoug (changes requested):
Looks good generally - a few comments about naming and the large(ish) anonymous function.

src/riak_core_security.erl
paulhenrich: good point, I'll change it to be find_user_by_metadata.

src/riak_core_security.erl
```diff
@@ -106,7 +144,7 @@ print_sources(Sources) ->
                   atom_to_list(Source), io_lib:format("~p", [Options])] ||
             {Users, CIDR, Source, Options} <- GS]).
 
--spec print_user(Username :: string()) ->
+-spec print_user(username()) ->
```
hazen: Seems like keeping in variable name where you can will help readability.

paulhenrich:
@JeetKunDoug I think all requested changes have been made. Merge?

JeetKunDoug (approved):

JeetKunDoug:
+1

jvoegele:
👍

jvoegele:
👍 c78d436

PR #877, opened 2016-11-22
Im/feature/fold bucket type

ian-mi:
Create fold function for custom buckets belonging to a bucket type.

### DIFF
--- a/src/riak_core_bucket.erl
+++ b/src/riak_core_bucket.erl
@@ -242,6 +242,21 @@ bucket_key({_Type, _Name}=Bucket) ->
 bucket_key(Name) ->
     {bucket, Name}.
 
+%% @doc Fold over all buckets belonging to the given bucket type. The
+%% provided fold function should accept three arguments: The bucket
+%% name, bucket properties, and accumulator and return a new
+%% accumulator value.
+-spec fold_type(bucket_type(), fun((Bucket :: binary(), properties(), Acc) -> Acc), Acc) -> Acc.
+fold_type(Type, F, Acc0) ->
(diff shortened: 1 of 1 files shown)


src/riak_core_bucket.erl
```diff
@@ -24,17 +24,18 @@
 -module(riak_core_bucket).
 
 -export([append_bucket_defaults/1,
-    set_bucket/2,
```
paulhenrich, line 27: While we're at it, what do you think of making them one-per line and sorting? Take all the diff-pain in one go.

src/riak_core_bucket.erl
```diff
@@ -240,6 +241,17 @@ bucket_key({_Type, _Name}=Bucket) ->
 bucket_key(Name) ->
     {bucket, Name}.
 
+-spec fold_type(bucket_type(), fun((binary(), properties(), Acc) -> Acc), Acc) -> Acc.
```
paulhenrich: I had a bit of trouble wrapping my head around this mostly because I am still learning about the metadata system. Jason and I were thinking it might be more clear if the signature specified that the Acc is a proplist?

What do you think about:
```
-spec fold_type(bucket_type(), fun((binary(), properties(), Acc) -> Acc), Acc) ->
                       Acc when Acc :: [{atom(), term()}].
```

A doc comment might help here too.

ian-mi: Acc can be any type used as an accumulator. I will add a doc to clarify.

edit: I generally prefer to have the accumulator on the left for left folds but I am hopefully sticking to erlang tradition by putting it on the right.

jvoegele:
👍

ian-mi:
This should no longer be needed

PR #879, opened 2016-12-01
Add thumbs

davidx:
This adds a thumbs config

### DIFF
--- a/src/riak_core_test_util.erl
+++ b/src/riak_core_test_util.erl
@@ -28,14 +28,19 @@
 -export([setup_mockring1/0,
          fake_ring/2,
          stop_pid/1,
-         wait_for_pid/1]).
+         wait_for_pid/1, stop_pid/2]).
 -include_lib("eunit/include/eunit.hrl").
 
 stop_pid(Other) when not is_pid(Other) ->
     ok;
 stop_pid(Pid) ->
+    stop_pid(Pid, kill).
+
+stop_pid(Other, _ExitType) when not is_pid(Other) ->
+    ok;
+stop_pid(Pid, ExitType) ->
     unlink(Pid),
-    exit(Pid, shutdown),
+    exit(Pid, ExitType),
     ok = wait_for_pid(Pid).
 
 wait_for_pid(Pid) ->
--- a/src/riak_core_test_util.erl
+++ b/src/riak_core_test_util.erl
@@ -31,6 +31,10 @@
          wait_for_pid/1, stop_pid/2]).
 -include_lib("eunit/include/eunit.hrl").
 
+stop_pid(undefined) ->
+    ok;
+stop_pid(Name) when is_atom(Name) ->
+    stop_pid(whereis(Name));
 stop_pid(Other) when not is_pid(Other) ->
     ok;
 stop_pid(Pid) ->
--- a/src/riak_core_test_util.erl
+++ b/src/riak_core_test_util.erl
@@ -28,7 +28,9 @@
 -export([setup_mockring1/0,
          fake_ring/2,
          stop_pid/1,
-         wait_for_pid/1, stop_pid/2]).
+         wait_for_pid/1,
+         stop_pid/2,
+         unlink_named_process/1]).
 -include_lib("eunit/include/eunit.hrl").
 
 stop_pid(undefined) ->
@@ -57,6 +59,9 @@ wait_for_pid(Pid) ->
             {error, didnotexit}
     end.
 
+
+unlink_named_process(Name) when is_atom(Name) ->
+    unlink(whereis(Name)).
 
 setup_mockring1() ->
     % requires a running riak_core_ring_manager, in test-mode is ok
--- a/test/bucket_eqc_utils.erl
+++ b/test/bucket_eqc_utils.erl
@@ -1,0 +1,56 @@
+%%%-------------------------------------------------------------------
+%%% @author doug
+%%% @copyright (C) 2016, <COMPANY>
+%%% @doc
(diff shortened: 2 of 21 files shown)


davidx:
created webhook after this pr. fyi

davidx:
thumbot retry

ooshlablu:
This appears to be a legit issue, not a thumbs issue.

JeetKunDoug:
So this build passes on buildbot, and it looks like it's having issues writing to the current directory in the thumbs run - not a _code_ issue, but an _environment_ issue running the test... going to ask Thumbs to retry and see if it's transient or not.

JeetKunDoug:
thumbot retry

davidx:
Strangely a different error when running it by hand in the shell. https://gist.github.com/davidx/e2114e0b465d548d703699be3f42695f
I wonder where it is trying to open that file.

JeetKunDoug:
So this passes all tests locally, and on Buildbot - must be something specific to Thumbs' environment?

JeetKunDoug:
thumbot retry

JeetKunDoug:
thumbot retry

JeetKunDoug:
@thumbot retry

JeetKunDoug:
@thumbot retry

src/riak_core_test_util.erl
```diff
@@ -28,14 +28,23 @@
...
 -include_lib("eunit/include/eunit.hrl").
 
+stop_pid(undefined) ->
+    ok;
+stop_pid(Name) when is_atom(Name) ->
+    stop_pid(whereis(Name));
 stop_pid(Other) when not is_pid(Other) ->
     ok;
 stop_pid(Pid) ->
+    stop_pid(Pid, kill).
```
hazen, line 43: You do want to `kill` instead of `shutdown` now?

test/bucket_eqc_utils.erl
```diff
@@ -0,0 +1,56 @@
+%%%-------------------------------------------------------------------
+%%% @author doug
+%%% @copyright (C) 2016, <COMPANY>
```
hazen: Maybe we should just use the regular Basho boilerplate?

test/riak_core_capability_tests.erl
```diff
@@ -0,0 +1,60 @@
+%% -------------------------------------------------------------------
+%%
+%% riak_core: Core Riak Application
+%%
+%% Copyright (c) 2007-2012 Basho Technologies, Inc.  All Rights Reserved.
```
hazen: We could update to 2017...

test/core_vnode_eqc.erl
```diff
@@ -52,17 +52,25 @@ simple_test_() ->
     {setup,
      fun setup_simple/0,
      fun(OldVars) ->
-             riak_core_ring_manager:stop(),
-	     application:stop(exometer),
-	     application:stop(lager),
-	     application:stop(goldrush),
-             [ok = application:set_env(riak_core, K, V) || {K,V} <- OldVars],
-             ok
+         unlink(whereis(riak_core_ring_manager)),
```
lehoff: How about adding a util function, say `unlink_named(ProcessName)`, and use that to make the code uniform?

test/node_watcher_qc.erl
```diff
@@ -44,47 +44,61 @@ qc_test_() ->
...
+                begin
+                    %% Setup ETS table to recv broadcasts
+                    ets:new(?MODULE, [ordered_set, named_table, public]),
+                    ets:insert_new(?MODULE, {bcast_id, 0}),
+
+                    %% Start the watcher
+                    {ok, Pid} = riak_core_node_watcher:start_link(),
+
+                    %% Internal call to the node watcher to override default broadcast mechanism
+                    gen_server:call(riak_core_node_watcher, {set_bcast_mod, ?MODULE, on_broadcast}),
```
lehoff: I would prefer if we added an exported function, `riak_core_node_watcher:set_broadcast_module(Module)` and wrap it with a `-ifdef(TEST)`, so that it is only available for testing. 

This will make it possible for us to change the implementation of the `riak_core_node_watcher` without having to change the tests as well.

We might as well clean that up while we are at it.

test/node_watcher_qc.erl
```diff
@@ -44,47 +44,61 @@ qc_test_() ->
...
     application:set_env(riak_core, gossip_interval, 250),
     application:set_env(riak_core, ring_creation_size, 8),
-
-    %% Start supporting processes
-    riak_core_eventhandler_sup:start_link(),
-    riak_core_ring_events:start_link(),
-    riak_core_node_watcher_events:start_link(),
-
-    %% meck used for health watch / threshold
+    RingEventHandlerSup = maybe_start_link(riak_core_eventhandler_sup),
```
lehoff: In line with the style used by `application`, I would prefer `ensure_started/1` over `maybe_start_link/1`.

lehoff (changes requested):

src/riak_core_test_util.erl
JeetKunDoug: yea - discussed w/ @lehoff when we did it - decided to default to kill

test/bucket_eqc_utils.erl
JeetKunDoug: oops - yea, totally forgot to fix that :) Will do.

lehoff (approved):

lehoff:
+1

davidx:
+1

bsparrow435:
+1

PR #881, opened 2016-12-05
Handle case where there are no existing users the same as if no users matched

jvoegele:
Previously, the `find_one_user_by_metadata` function would return `{error, not_found}` for the case when no users matched the metadata query, but would return an empty list when there were no existing users registered with `riak_core_security`. This change is to make both of those cases return the same value: `{error, not_found}`.


paulhenrich:
👍

PR #882, opened 2016-12-06
Merge sweeper code to develop

nickelization:
This is just some support code Mikael wrote for allowing tombstones in AAE trees, which prevents expired data from being accidentally resurrected. The code that uses this will be merged into riak_kv.


JeetKunDoug:
+1

PR #885, opened 2016-12-13
Add riak_core_security:find_unique_user_by_metadata

ian-mi:
This functions identically to find_one_user_by_metadata but returns
{error, not_unique} if there are multiple users with matching
metadata.


JeetKunDoug:
+1

paulhenrich:
+1 with a note to self to revisit our find_one caller to see if we can replace it with this.

PR #888, opened 2016-12-13
Add count/2 function to riak_core_util

nickelization:
This is useful for determining how many elements of a list match a certain criteria, without having to waste memory on intermediate lists, and without having to rewrite the fold yourself every time.

Similar to Enum.count from Elixir.

Also includes a simple EQC prop for testing.


lehoff:
👍

PR #890, opened 2016-12-19
Include node() in getting TS table type status

paegun:
* change to include the node in getting the (TS table) type status.

### DIFF
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -719,7 +719,7 @@ all_members() ->
 -spec get_remote_type_status(BucketType::binary()) ->
     {AllProps::[any()], BadNodes::[node()]}.
 get_remote_type_status(BucketType) ->
-    rpc:multicall(all_members() -- [node()],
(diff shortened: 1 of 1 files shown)


src/riak_core_claimant.erl
```diff
@@ -719,7 +719,7 @@ all_members() ->
 -spec get_remote_type_status(BucketType::binary()) ->
     {AllProps::[any()], BadNodes::[node()]}.
 get_remote_type_status(BucketType) ->
-    rpc:multicall(all_members() -- [node()],
+    rpc:multicall(all_members(),
```
macintux, line 722: Given that we only call this from one spot in the claimant, I'd like to see the function renamed.

macintux: I guess that means we should rename `get_remote_ddl_compiled_status` for consistency.

macintux:
+1

hmmr:
+1

macintux:
No Thumbs, merging.

PR #891, opened 2016-12-30
Port of Add thumbs (#879) to TS [JIRA: RIAK-3246]

hazen:
`cherry-pick` of https://github.com/basho/riak_core/pull/879

* add thumbs
* Hopefully resolve test process stop issues.

- Add a call to `meck:unload/0` in setup for all tests that use meck
  in order to make sure meck processes are unloaded _before_ starting
  up new ones. It appears eunit doesn't always call a teardown function
  before moving on to another test, so they can't be depended upon.
- unlink before stopping `riak_core_ring_manager` and `riak_core_ring_events`
  in `core_vnode_eqc` - this seems to be what was actually causing the
  "unexpected exit" of a test process.
- add `stop_pid/2` in riak_core_test_util so you can pass an exit type
  which was being used by a test and failing
- unlink processes that are `start_link`ed to tests
- Fix sync command typo to stop riak_core_ring_events
* Add some additional fixes to stopping processes started with `start_link`.
- Stop proesses start_linked
- Add a `spawn` wrapper around a test to see if it helps with isolating tests (it did).
* `meck:unload()` before starting `core_vnode_eqc` test.

Ends up, the `meck` calls in `riak_core_util` were breaking
`core_vnode_eqc`, which doesn't even _use_ `meck`. Call `meck:unload()`
at end of test in `riak_core_util` (not in test teardown, which can also
cause problems), and again at the beginning of `core_vnode_eqc` just in
case some new test later on tries to break it again.
* Move `riak_core_capability` tests to their own file.

Because of `meck` issues we need to keep unit tests and modules
separate for modules that are mecked in other tests.
* Address review comments on test fixes.

- add `set_broadcast_module/2` to riak_core_node_watcher, wrapped in
  an `-ifdef(TEST)` so we can potentially change the implementation
  later without having to change the test, which used to do a direct
  `gen_server:call` to the `set_bcast_mod` `handle_call` function.
- Change `riak_core_ring_manager:stop/0` from a cast to a call. As a
  cast, it leaves us vulnerable to potential race conditions in test
  startup/shutdown. Because it was only a test function anyway, making
  the caller wait for the stop seems reasonable. Also wrapped it in
  `-ifdef(TEST)` to make sure it's not called from production code.
- Add `unlink_named_process` to `riak_core_test_util` and use it
  where we had `unlink(whereis(name_of_process))`
- remove last copy/paste of `stop_pid` function from `bg_manager_eqc`
  and `core_vnode_eqc` and use `riak_core_test_util` functions instead.
- Fix copyright notice in bucket_eqc_utils.
- Found that we needed to call `meck:unload` after each test in
  `bprops_eqc` and `btypes_eqc` to avoid a meck/coverdata issue.
  Moved setup and unload of `meck`s in `bucket_eqc_utils` to the
  `per_test_setup` function and dropped `setup_cleanup/0` as it was no
  longer doing anything.
* Update `bprops_eqc` and `btypes_eqc` to use `eqc:testing_time` instead
of `eqc:numtests` as it's possible for it to time out via `eunit` if
you allow them to do 100 iterations.

### DIFF
--- a/.thumbs.yml
+++ b/.thumbs.yml
@@ -1,0 +1,11 @@
+minimum_reviewers: 2
+merge: true
+build_steps:
+  - make clean
+  - make deps
+  - make compile
+  - make test
+  - make xref
+  - make dialyzer
+org_mode: true
+timeout: 1800
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -109,6 +109,13 @@
 -define(ETS, riak_capability_ets).
 -define(CAPS, '$riak_capabilities').
 
+-ifdef(TEST).
+-compile(export_all).
+-type state() :: #state{}.
+-export_type([state/0]).
+
+-endif.
+
 %%%===================================================================
 %%% API
 %%%===================================================================
@@ -156,6 +163,12 @@ get(Capability, Default) ->
         _:_ ->
             Default
     end.
+
+-ifdef(TEST).
+%% @doc Exported for testing - takes opaque state record and returns negotiated
+get_negotiated(State) ->
+    State#state.negotiated.
+-endif.
 
 %% @doc Return a list of all negotiated capabilities
 all() ->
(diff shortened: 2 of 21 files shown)


hazen:
@JeetKunDoug @lehoff Here's the TS version of https://github.com/basho/riak_core/pull/879

hazen:
create jira issue

macintux:
+1

Every change is to test code and has been reviewed by wiser folk than me. Who am I to say they lie?

paegun:
+1

PR #893, opened 2017-01-11
Fix RIAK-2643 by adding debug logging for meta and ring

bsparrow435:
There have been a few issues where rapidly updating the ring results in
sub optimal behavior. These have been difficult to debug due to the
lack of logging in the `riak_core_ring` module. We’ve added debug
logging which can be enabled as needed to log ring metadata
merges(buckets, real, etc) and ring membership and resulting rings from
a reconciliation. For 2.2.1, we should add a new lager sink just for
logging ring operations.


JeetKunDoug:
+1

hazen:
+1

PR #894, opened 2017-01-11
Fix RIAK-2169 by fixing stat name `dropped_vnode_requests`

bsparrow435:
The stat `dropped_vnode_requests_total` was not increment correctly in Riak stats because it was incorrectly projected in `riak_core_stat`.  Fix this error by setting the correct stat name: `dropped_vnode_requests`.


JeetKunDoug:
+1

hazen:
+1

PR #895, opened 2017-01-18
Worker Pool Replacement Review ONLY

tburghart:
### DO NOT MERGE!!!

This PR exists ***ONLY*** to make commenting easier to deal with.


src/riak_core_job_manager.erl
```diff
@@ -0,0 +1,1310 @@
...
+}).
+
+% 'running' dictionary, RunnerRef => full record
+-record(rd, {
+    c   = 0             ::  non_neg_integer(),      % count
+    d   = dict:new()    ::  ?dict_t(rkey(), rrec()) % data
+}).
+
+-record(state, {
+    rmax                            ::  pos_integer(),
```
nickelization: Some more descriptive field names would definitely be helpful here, or at least some comments as to what `rmax`/`qmax`/`hmax`/etc stand for.

src/riak_core_job_manager.erl
```diff
@@ -0,0 +1,1310 @@
...
+%
+% cancel(JobId :: jid(), Kill :: boolean())
+%   ->  {ok, killed | canceled | history, job()}
+%       | {error, running, job()} | false | {error, term()}
+%
+handle_call({cancel, JobId, Kill}, _, State) ->
+    case c_find(JobId, State#state.loc) of
+
+        {_, Ref} when erlang:is_reference(Ref) ->
+            case Kill of
```
nickelization: Too many long, nested case statements here. At minimum, we should break this section out into something like a `cancel_by_reference` helper function with separate clauses for `Kill` being `true` or not.

src/riak_core_job_manager.erl
```diff
@@ -0,0 +1,1310 @@
...
+-spec handle_call(Msg :: term(), From :: {pid(), term()}, State :: state())
+        -> {reply, term(), state()} | {stop, term(), term(), state()}.
+%% @private
+%% @end
+%
+% config() -> config() | {error, term()}.
+%
+handle_call(config, _, State) ->
+    {reply, [
+        {?JOB_SVC_CONCUR_LIMIT, State#state.rmax},
```
nickelization: Why do we call the concurrency limit `rmax` instead of `cmax`?

src/riak_core_job_manager.erl
```diff
@@ -0,0 +1,1310 @@
...
+            JobOrId
+    end,
+    gen_server:call(?JOBS_MGR_NAME, {find, JobId}).
+
+-spec reconfigure() -> ok.
+%%
+%% @doc Re-read application environment configuration and adjust accordingly.
+%%
+reconfigure() ->
+    gen_server:cast(?JOBS_MGR_NAME, ?job_svc_cfg_token).
```
nickelization: I read the comments where this token is defined, but this still feels a bit odd and unidiomatic to me. Why are we using a special token here, and not just an atom like `reload_job_config` or something? In theory, I guess it's possible that someone else might write code somewhere else in Riak that uses the same atom and then somehow accidentally sends it to the wrong process, but that seems exceedingly unlikely in practice.

src/riak_core_job_manager.erl
tburghart: `rmax` is the maximum allowable size of the `run` collection.
`qmax` is the maximum allowable size of the `que` collection.
`hmax` is the maximum allowable size of the `hist` collection.

src/riak_core_job_manager.erl
tburghart: maximum Running jobs

src/riak_core_job_manager.erl
tburghart: perhaps ... I debated it when writing that function, I'll consider it more

src/riak_core_job_manager.erl
tburghart: The communication between the `manager` and `service` in this case is private - there's no public API to tell the `service` to reconfigure itself.
I wanted to be sure it wouldn't accidentally be used for something else, and atoms are constant-cost whether simple or complex.
I debated whether to support reconfiguring on the fly at all, as it's fraught with peril, at least if you're lowering limits, but I have reason to believe the CS folks might want it in their bag of tricks.

src/riak_core_job_manager.erl
nickelization: Hah, thanks for the info, I meant comments in the code though, not for my own personal edification ;)

src/riak_core_job_sup.erl
```diff
@@ -0,0 +1,208 @@
...
+-spec stop_runner(Runner :: pid()) -> ok | {error, term()}.
+%% @private
+%% @doc Stop a job runner process.
+%%
+stop_runner(Runner) when erlang:is_pid(Runner) ->
+    case supervisor:terminate_child(?MODULE, Runner) of
+        ok ->
+            ok;
+        {error, not_found} = Error ->
+            % Errr, is it a runner that might have been started by a previous
```
nickelization: If the supervisor died, any runners running underneath it should die too, since they are linked. This extra cleanup code is unnecessary, and it may cause delays in the execution of this function since the `receive` block will always time out.

nickelization:
It looks like all your new Erlang source files were created with the executable permissions bit set, and a couple of existing files were modified to have that bit set as well. I assume this was a mistake?

tburghart:
Well, that _IS_ interesting ... I wonder where that came from.
The file modes changed at commit 5eb16bdd0f68deea075713c663475c3c2f2e0d1a, and I'm thinking perhaps I took a snapshot of them on a USB thumb drive, which would likely be a FAT32 filesystem, and restored from it, which usually defaults to the execute permission being on because FAT doesn't have that bit.
Have to keep that in mind, and I'll fix them up when I address other comments here.

src/riak_core_job_sup.erl
tburghart: That comment is not wholly accurate.  It's defensive code because I was moving a _LOT_ of stuff around and wanted to handle the case of a stray runner process.
It needs a couple of things:
* A loud error in the logs and/or on the console saying it occurred.
* Either a 'todo' comment that it should be removed when the code has sufficient mileage on it that we're comfortable it won't happen.
* OR an async process to confirm it's a stray runner and act accordingly.

src/riak_core_job_manager.erl
tburghart: Ok, I suppose I could spare a few keystrokes on a comment ;)

tburghart:
@nickelization I _think_ I've addressed all of your comments

PR #896, opened 2017-01-20
Remove dialyzer.ignore-warnings file

nickelization:
Turns out none of the warnings in this file actually exist anymore 😂 

More generally, I can't remember the last time I've written code that generated a warning that wasn't fixable. Rather than emptying out dialyzer.ignore-warnings, I'm just going to blow it away altogether to discourage people from relying on it in the future. We can always add it back if we really need it for some reason.


bsparrow435:
👍

JeetKunDoug:
+1

JeetKunDoug:
Thanks for noticing this - we should certainly be attempting to get rid of these things.

PR #897, opened 2017-01-20
Update behavior specs to use -callback directives [JIRA: RIAK-3243]

nickelization:
This PR updates the `riak_core_vnode_worker` and `riak_core_coverage_fsm` behaviors to use `-callback` directives instead of the old-style `behaviour_info` system. I've done my best to figure out all the proper type specs for the behaviors, and Dialyzer passes so I think this is all correct, though if anyone knows of a way we could tighten up the specs any further that'd always be welcome.

### DIFF
--- a/src/riak_core_coverage_fsm.erl
+++ b/src/riak_core_coverage_fsm.erl
@@ -67,8 +67,6 @@
 -behaviour(gen_fsm).
 
 %% API
--export([behaviour_info/1]).
-
 -export([start_link/3]).
 
 -ifdef(TEST).
@@ -87,28 +85,30 @@
          terminate/3,
          code_change/4]).
 
--spec behaviour_info(atom()) -> 'undefined' | [{atom(), arity()}].
-behaviour_info(callbacks) ->
-    [
-     {init, 2},
-     {process_results, 2},
-     {finish, 2}
-    ];
-behaviour_info(_) ->
-    undefined.
+-type mod_state() :: term().
+-callback init(from(), RequestArgs :: [term()]) ->
+    {Request :: term(), vnode_selector(), NVal :: pos_integer(), primary_vnode_coverage(),
+     NodeCheckService :: module(), VNodeMaster :: atom(), timeout(), mod_state()}.
+-callback process_results(Results :: term(), mod_state()) ->
+    {ok, mod_state()} |
+    {done, mod_state()} |
+    {error, term()}.
+-callback finish(clean | {error, term()}, mod_state()) -> {stop, normal, mod_state()}.
 
 -define(DEFAULT_TIMEOUT, 60000*8).
 
 -type req_id() :: non_neg_integer().
 -type from() :: {atom(), req_id(), pid()}.
+-type vnode_selector() :: all | allup.
+-type primary_vnode_coverage() :: all | pos_integer().
 
 -record(state, {coverage_vnodes :: [{non_neg_integer(), node()}],
                 mod :: atom(),
                 mod_state :: tuple(),
                 n_val :: pos_integer(),
                 node_check_service :: module(),
-                vnode_selector :: all | allup,
-                pvc :: all | pos_integer(), % primary vnode coverage
+                vnode_selector :: vnode_selector(),
+                pvc :: primary_vnode_coverage(),
                 request :: tuple(),
                 req_id :: req_id(),
                 required_responses :: pos_integer(),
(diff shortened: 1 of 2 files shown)


src/riak_core_coverage_fsm.erl
```diff
@@ -87,28 +85,30 @@
...
-behaviour_info(_) ->
-    undefined.
+-type mod_state() :: term().
+-callback init(from(), RequestArgs :: [term()]) ->
+    {Request :: term(), vnode_selector(), NVal :: pos_integer(), primary_vnode_coverage(),
+     NodeCheckService :: module(), VNodeMaster :: atom(), timeout(), mod_state()}.
+-callback process_results(Results :: term(), mod_state()) ->
+    {ok, mod_state()} |
+    {done, mod_state()} |
+    {error, term()}.
```
ian-mi, line 95: Maybe `{ok | done, mod_state()} | {error, term()}` is clearer?

src/riak_core_coverage_fsm.erl
```diff
@@ -87,28 +85,30 @@
...
--spec behaviour_info(atom()) -> 'undefined' | [{atom(), arity()}].
-behaviour_info(callbacks) ->
-    [
-     {init, 2},
-     {process_results, 2},
-     {finish, 2}
-    ];
-behaviour_info(_) ->
-    undefined.
+-type mod_state() :: term().
```
ian-mi, line 88: It's curious that the type of `mod_state` in the state record is `tuple()`. Should this match `mod_state()`?

src/riak_core_vnode_worker.erl
```diff
@@ -37,12 +38,11 @@
         modstate :: any()
     }).
 
--spec behaviour_info(atom()) -> 'undefined' | [{atom(), arity()}].
-behaviour_info(callbacks) ->
-    [{init_worker,3},
-     {handle_work,3}];
-behaviour_info(_Other) ->
-    undefined.
+-callback init_worker(partition(), Args :: term(), Props :: [{atom(), term()}]) ->
```
ian-mi, line 43: Should `Props` be a proplist with type `[atom() | {atom(), term()}]` or `[proplists:property()]`?

ian-mi (commented):
Maybe a mod_state() type could be added to the riak_core_vnode_worker module to match the style of riak_core_coverage_fsm?

src/riak_core_coverage_fsm.erl
nickelization: I think `mod_state` in the state record should be `term()`, not `tuple()`. Was going to change it there as well because I see no reason to restrict it, but then I forgot. Will change and update.

src/riak_core_coverage_fsm.erl
nickelization: That would be shorter, but I don't feel like it's clearer: there are three distinct types of result that `process_results` can return, so having each one split out into its own top-level type in the spec feels more intuitive to me.

We can squash the first two together, but only because they happen to bear a superficial resemblance in their structure, not because they necessarily share anything else in common.

src/riak_core_vnode_worker.erl
nickelization: I'm not sure if it's really important, since every implementation of `init_worker` that we have ignores the `Props` argument 😆 . I guess it depends on whether we want to encourage people to use `proplists` or not, and I'm not a huge fan of the `proplists` module, so my vote would be to leave it as is...but admittedly that's fairly arbitrary ;). If you feel strongly about it, I can change it. Or we can also always leave it for now, and then widen the spec in the future if we decide we want to make it more permissive.

nickelization:
@ian-mi Thanks for the review! I updated some things based on your comments. Let me know if you're satisfied with the changes or if there's anything else you'd like to see.

src/riak_core_vnode_worker.erl
```diff
@@ -38,11 +38,12 @@
         modstate :: any()
```
ian-mi: Perhaps this could use mod_state() as well?

nickelization: Good call. Done.

nickelization: If there's anything else, let me know. Otherwise I'll rebase all these fixup commits back into the original checkin.

ian-mi:
+1

bashopatricia:
create jira issue

Basho-JIRA:
This was already reviewed and merged to develop on Jan 26: marking as closed. 

***[posted via JIRA by Nick Marino]***

PR #898, opened 2017-02-14
riak_core_info_service

JeetKunDoug:
This PR implements the `riak_core_info_service` set of processes - a supervisor and one or more children that are created to support NIFs that need to call in to Riak to get information.
- `riak_core_info_service` - API
- `riak_core_info_service_sup` - supervisor for `riak_core_info_service_process` processes.
- `riak_core_info_service_process` - a lightweight process that can respond to a specified request by invoking a function and calling a callback to return the results.

See `riak_core_app.erl` for an example of registering and starting a service.

TODO:
- [x] Decide if hard-coding the `eleveldb` service registration is acceptable, or if we want to use a more indirect method of registering the callbacks in this case (cuttlefish schema perhaps? Do we have other ways of doing this that don't require `eleveldb` to directly call in to `riak_core`?)
- [x] Perhaps make the `riak_core_info_service_process` a `gen_server` at this point.
- [x] Add copyright banners


src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,113 @@
+-module(riak_core_info_service_process).
```
tburghart, line 22: At this point you've pretty much rolled your own `gen_server`, so it doesn't seem to me that the argument against using the OTP behavior has any basis left.

JeetKunDoug: I was thinking the same thing - simple `gen_server` with a few `handle_info` function heads should work just as well, and be less boilerplate code. I didn't know what the original reasoning was - maybe @macintux can weigh in and if he agrees we can just turn this into a `gen_server` process.

macintux: The original "API" was leveldb sending custom Erlang messages that would not be routed via OTP. I felt strongly that implementing an OTP behavior would be effectively lying about the role the process played.

I haven't looked at the new code to see what's been done to it.

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,69 @@
...
+         handle_cast/2,
+         handle_info/2,
+         terminate/2,
+         code_change/3]).
+
+
+%% `gen_server' implementation
+
+start_link(Registration, Shutdown, Source, ResponseHandler) ->
+    {ok, Pid} = gen_server:start_link(?MODULE, [Shutdown, Source, ResponseHandler], []),
```
tburghart: Trivial, but why not just create the `#state{}` record here and pass it through intact?

JeetKunDoug: Probably just an "order of operations thing," as in, I morphed the code in such a way that this is how it ended up... good call though, will update.

src/riak_core_app.erl
```diff
@@ -145,3 +146,12 @@ register_capabilities() ->
...
     ok.
+
+%% TODO: This belongs in riak_kv - an issue will be created to move it, but time constraints.
+start_eleveldb_info_service() ->
+    Registration = {eleveldb, set_metadata_pid, []},
+    Shutdown = undefined, %%{eleveldb, metadata_service_shutdown, []}
+    InfoSource = {riak_core_bucket, get_bucket, []},
+    ResultsHandler = {eleveldb_metadata, handle_metadata_response, []},
+    {ok, _Pid} = riak_core_info_service:start_service(Registration, Shutdown, InfoSource, ResultsHandler),
+    ok.
```
macintux: Here's a nickel, kid. Get yourself a real editor.

(With apologies to Scott Adams. http://dilbert.com/strip/1995-06-24)

(No end of line marker on the last line in the file.)

JeetKunDoug: fixed

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,69 @@
+-module(riak_core_info_service_process).
+-behaviour(gen_server).
+
+-record(state, {
+    source = undefined :: riak_core_info_service:callback(),
```
macintux: Suggest a different name; I first thought `source` was the source of the request. Perhaps `service_provider`?

macintux:
I'm reasonably certain the whitelist for mapreduce will prevent this from being misused, but it's worth verifying my memory is correct.

src/riak_core_info_service_process.erl
JeetKunDoug: +1 - makes sense to me... will think about the name a bit but I like `service_provider`

JeetKunDoug: yea - go with `service_provider` unless you come up with something better - please make sure to rename things like the variables in `riak_core_app.erl` to match whatever you come up with.

macintux:
I think I'm about ready to call this sufficient. Want to add a bit of testing first.

macintux:
Ready for review.

src/riak_core_app.erl
```diff
@@ -145,3 +146,12 @@ register_capabilities() ->
       end,
       Capabilities),
     ok.
+
+%% TODO: This belongs in riak_kv - an issue will be created to move it, but time constraints.
```
raghavkarol, line 152: Is the `eleveldb` app also initialised and available when using a different backend, e.g., `bitcask`?

raghavkarol (commented):
@macintux, @JeetKunDoug - LGTM, please see comments.

raghavkarol:
+1

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,120 @@
...
+    apply_callback(Registration, [Pid]),
+    {ok, Pid}.
+
+init(State) ->
+    {ok, State}.
+
+handle_info({invoke, ProviderParams, HandlerContext}, State) ->
+    handle_invoke_message(ProviderParams, HandlerContext, State);
+
+handle_info(callback_shutdown, State) ->
```
raghavkarol, line 69: `handle_shutdown_message` will be called twice when handling `callback_shutdown`. Perhaps change this clause to only return `{stop, normal, State}` and call `apply_callback(Shutdown, ...` from only from `terminate`?

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,120 @@
...
+    handle_shutdown_message(State).
+
+terminate(_Reason, State) ->
+    handle_shutdown_message(State).
+
+code_change(_OldVsn, State, _Extra) ->
+    State.
+
+handle_call(_Request, _From, _State) ->
```
raghavkarol: Perhaps just return `{stop, not_implemented, State}.` instead?

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,120 @@
...
+    erlang:error(not_implemented).
+
+handle_cast(_Request, _State) ->
```
raghavkarol: Perhaps just return `{stop, not_implemented, State}.` instead?

src/riak_core_app.erl
JeetKunDoug: Because `riak_core` uses `eleveldb` for its hashtree implementation, it's always available - the app itself doesn't have initialization of its own (especially now that we've moved this over to `riak_core`.

src/riak_core_info_service_process.erl
JeetKunDoug: Perhaps also somehow include the message that came in (the _Request) in the reason so when it's logged we know which message showed up that we didn't understand?

src/riak_core_info_service_process.erl
macintux: Good catch.

macintux: Fixed.

src/riak_core_info_service_process.erl
macintux: Fixed.

src/riak_core_info_service_process.erl
macintux: Fixed.

raghavkarol:
+1

src/riak_core_app.erl
raghavkarol: Understood, and that also answers another question I had about why `riak_core` depends on `eleveldb`.

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,124 @@
...
+-endif.
+
+%% `gen_server' implementation
+
+start_link(Registration, Shutdown, Provider, ResponseHandler) ->
+    State = #state{shutdown = Shutdown,
+                   service_provider = Provider,
+                   response_handler = ResponseHandler},
+    {ok, Pid} = gen_server:start_link(?MODULE, State, []),
+    apply_callback(Registration, [Pid]),
```
tburghart: Shouldn't this be called from `init`?
If the registration callback fails, we don't want the service merrily chugging along.
At a minimum, the result of the callback should be checked so `start_link` fails if the registration does.

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,124 @@
...
+    handle_invoke_message(ProviderParams, HandlerContext, State);
+
+handle_info(callback_shutdown, State) ->
+    {stop, normal, State}.
+
+terminate(_Reason, State) ->
+    handle_shutdown_message(State).
+
+code_change(_OldVsn, State, _Extra) ->
+    State.
```
tburghart: Return `{ok, State}`

src/riak_core_info_service.erl
```diff
@@ -0,0 +1,104 @@
...
+%%%      === Handler Parameters ===
+%%%      The arguments to `Handler' will be, in order:
+%%%      <ol><li>Each item, if any, provided in the list in the 3rd element of the registration tuple</li>
+%%%      <li>The result from `Provider' wrapped in a 3-tuple:
+%%%        <ol><li>The result</li>
+%%%            <li>The list of parameters passed to `Provider'</li>
+%%%            <li>The opaque `HandlerContext' (see {@section Request Message})</li>
+%%%        </ol></li></ol>
+%%%
+%%%
```
tburghart: Handler Results _REALLY_ should be specified.
As it stands, I'm uncomfortable with the lack of return value checks, and of course there's not anything to check when they're not specified.

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,124 @@
...
+    lager:warning("Terminating - unknown message received: ~p", [Request]),
+    {stop, not_implemented, State}.
+
+handle_invoke_message(ProviderParams, HandlerContext,
+                      #state{service_provider = Provider,
+                             response_handler = ResponseHandler } = State) ->
+    Result = apply_callback(Provider, ProviderParams),
+    Reply = {Result, ProviderParams, HandlerContext},
+    apply_callback(ResponseHandler, [Reply]),
+    {noreply, State}.
```
tburghart: The lack of error checking (and reporting!) scares me.

tburghart (changes requested):
At a minimum, there's a contract violation that dialyzer should report.

I'm also more than a little uncomfortable with the lack of callback error checking.

src/riak_core_info_service.erl
macintux: Reworking this per discussions with Ted.

Notable change to behavior: `undefined` will no longer be an acceptable callback.

macintux:
With the new error handling code, this has definitely reached a point where `riak_test` is indicated. Will do the needful.

macintux:
Rather than do ugly stuff with `riak_test` to expose the callbacks to the devrel node(s), I did less-ugly stuff with a unit test inside `riak_core_info_service`.

I also caught some glitches along the way: dialyzer fix, create a shutdown callback for eleveldb. That last one will require some careful thought by someone not awake until 5am (i.e., not me).

src/riak_core_info_service_process.erl
```diff
@@ -81,20 +90,55 @@ terminate_not_implemented(Request, State) ->
...
+%% The original response to the callback will be wrapped in an `ok' or
+%% `exit' tuple, primarily so that the service process can terminate
+%% if something goes wrong during registration.
+-spec apply_callback(callback(), list(term())) -> {ok, term()} |
+                                                  {error, term()}.
 apply_callback({Mod, Fun, StaticArgs}, CallSpecificArgs) ->
     %% Invoke the callback, passing static + call-specific args for this call
-    erlang:apply(Mod, Fun, StaticArgs ++ CallSpecificArgs).
+    Args = StaticArgs ++ CallSpecificArgs,
+    _ = case (catch erlang:apply(Mod, Fun, Args)) of
```
raghavkarol: with `catch`, a `throw(...)` in `erlang:apply(Mod, Fun, Args)` matches `Result` and is indistinguishable from a return value. If this is not the intention, a `try-catch` should be used instead.

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,168 @@
...
+
+%% The original response to the callback will be wrapped in an `ok' or
+%% `exit' tuple, primarily so that the service process can terminate
+%% if something goes wrong during registration.
+-spec apply_callback(callback(), list(term())) -> {ok, term()} |
+                                                  {error, term()}.
+apply_callback({Mod, Fun, StaticArgs}, CallSpecificArgs) ->
+    %% Invoke the callback, passing static + call-specific args for this call
+    Args = StaticArgs ++ CallSpecificArgs,
+    _ = case (catch erlang:apply(Mod, Fun, Args)) of
```
tburghart, line 111: `_ = ` not needed because each case returns from the function

src/riak_core_info_service_process.erl
```diff
@@ -0,0 +1,168 @@
...
+            {'EXIT', {Reason, _Stack}}=Exit->
+                %% provider called erlang:error(Reason)
+                lager:warning("~p:~p/~B exited (~p)",
```
tburghart, line 114: `lager` calls, otoh, can return structured results, so dialyzer will flag them when the option to check returns is on.

src/riak_core_info_service_process.erl
tburghart: I think that's suitable.  The code in here is pretty much the example I sent John, and I thought about that situation while throwing it together.  If we were to use a `try/catch`, we'd effectively be saying a return by `throw` is an error, when it has plenty of uses as a simple `break` out of a loop.

tburghart (approved):

macintux:
Test failure is interesting. Also interesting that the status box on this PR doesn't make it obvious that there is a failure with the build: https://github.com/basho/riak_core/pull/898#issuecomment-283011210

The two new tests pass when running eunit on just the module locally, now trying eunit across entire project.

Update: they fail locally as well, which will make this easier to fix. Please no one provide a 2nd approval yet.

src/riak_core_info_service_sup.erl
```diff
@@ -0,0 +1,50 @@
...
+
+-export([init/1]).
+
+-define(SERVER, ?MODULE).
+
+start_link() ->
+    supervisor:start_link({local, ?SERVER}, ?MODULE, []).
+
+start_service(Registration, Shutdown, Provider, Handler) ->
+    supervisor:start_child(?SERVER, [Registration, Shutdown, Provider, Handler]).
```
hazen, line 37: Is this a valid way to start a child? http://erlang.org/doc/man/supervisor.html#start_child-2

Here's my failed attempt at a child spec:
```
    ChildSpec = {?SERVER, {riak_core_info_service_process, start_link, [Registration, Shutdown, Provider, Handler]}, permanent, 5000, worker, [riak_core_info_service_process]},
    supervisor:start_child(?SERVER, ChildSpec).
```

macintux: `simple_one_for_one` supervisors handle child specs differently. See the 4th paragraph in that section of the document.

macintux:
Testing problem identified, fixed. Also some other cleanup on the tests. Ready for re-review.

src/riak_core_info_service.erl
```diff
@@ -124,121 +124,144 @@ start_service(Registration, Shutdown, Provider, Handler) ->
...
-    throw(no_supervisor);
-wait_for_sup(Regname, Count) ->
-    case whereis(Regname) of
-        undefined ->
-            timer:sleep(50),
-            wait_for_sup(Regname, Count-1);
-        _ ->
+%% Launching and terminating the supervisor is an async operation with
+%% all the non-determinism that that implies. So, we wait.
+sup_wait(Name, Fun) ->
```
hazen: Much cleaner design. Like it!

tburghart:
+1

src/riak_core_app.erl
```diff
@@ -152,8 +152,11 @@ register_capabilities() ->
 %% TODO: This belongs in riak_kv - an issue will be created to move it, but time constraints.
 start_eleveldb_info_service() ->
     Registration = {eleveldb, set_metadata_pid, []},
-    Shutdown = {?MODULE, start_eleveldb_info_service, []},
+    Shutdown = {?MODULE, shutdown_eleveldb_info_service, []},
```
tburghart: Changing this to `{eleveldb, remove_metadata_pid, []}` should shut xref and dialyzer up, with the added benefit of being consistent with registration and eliminating the separate `shutdown_eleveldb_info_service/1` function altogether.

macintux:
xref and dialyzer are unhappy because Thumbs doesn't support multiple unnerved branches yet. Both pass locally using the proper eleveldb branch.

I agree it's reasonable to remove the indirection anyway. Will do.

tburghart:
+1 9eda941

raghavkarol (approved):
LGTM - I suggest using common test instead of eunit for tests that need fixture setup. Each common test case is a separate process, so flushing messages is not required. Also, as of now, rebar3 doesn't have an options to run a single eunit test case (rebar does), while it is possible to do this for ct

raghavkarol:
+1

PR #900, opened 2017-02-24
Fix riak_core_util:compose functions

jvoegele:
Fix the `compose/1` and `compose/2` functions in `riak_core_util`:
* Compose lists of functions from right-to-left to match mathematical definition.
* Add type variables to the type specs to express the intent that all functions being composed together should have the same type signature (a single argument of type `X`, returning a value of type `X`).
* Add a test case that exposed the bug in the previous implementation that composed from left-to-right.

### DIFF
--- a/src/riak_core_util.erl
+++ b/src/riak_core_util.erl
@@ -360,19 +360,22 @@ multi_keydelete(KeysToDelete, N, TupleList) ->
 
 %% @doc Function composition: returns a function that is the composition of
 %% `F' and `G'.
--spec compose(fun(), fun()) -> fun().
+-spec compose(fun((X) -> X), fun((X) -> X)) -> fun((X) -> X).
 compose(F, G) when is_function(F), is_function(G) ->
     fun(X) ->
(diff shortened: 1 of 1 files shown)


paulhenrich:
Where is this called? I went looking for callers in case anyone was relying on the wrong behavior, but I couldn't find any.

paulhenrich:
Not a caller, but while grepping I found another wrong implementation in https://github.com/basho/riak_repl/blob/develop/src/riak_repl_ring.erl#L498

paulhenrich:
👍 Couldn't find any callers with a cursory search of GH and other OSS repos. We should probably still add something to the release notes (cc @angrycub)

angrycub:
I agree.  We should note this in the release notes.  Could you write a comment containing the release note for this pull request.

jvoegele:
@angrycub here is a comment for the release notes:

The `riak_core_util:compose/1` function has been modified to compose functions from right-to-left to match the mathematical definition of function composition.

PR #903, opened 2017-03-03
More robustness additions for the info service

macintux:
* Do not let `riak_core` crash on startup if the `eleveldb` callbacks are not present
* Verify the callbacks (3 of them, anyway) are available at registration time

PR #904, opened 2017-03-03
More robustness additions for the info service

macintux:
* Do not let riak_core crash on startup if the eleveldb callbacks are not present
* Verify the callbacks (3 of them, anyway) are available at registration time
* Add context to the registration and shutdown callbacks for future eleveldb work

### DIFF
--- a/src/riak_core_info_service.erl
+++ b/src/riak_core_info_service.erl
@@ -26,7 +26,7 @@
 %%%      == Callbacks ==
 %%%      The dependency needs to know to which pid it should send
 %%%      requests for `riak_core' data. The pid will be sent to the
-%%%      `Registration' callback.
+%%%      `Registration' and `Shutdown' callbacks.
 %%%
 %%%      The `Provider' callback is a function inside `riak_core' that returns
 %%%      information that the dependency needs.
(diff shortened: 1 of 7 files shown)


macintux:
Replaces #903 because git weirdness

src/riak_core_info_service.erl
```diff
@@ -227,6 +248,10 @@ exception_test() ->
     teardown(Sup),
     ok.
 
+no_callback_test() ->
+    Key = 'no_callback_test',
+    ?assertMatch({error, _}, riak_core_info_service:start_service(?BOGUS_REG, ?SHUT(Key), ?GET_RING, ?HANDLER(Key))).
```
fadushin: Should this be `?assertMatch({error, {not_callable, _Fun}}, ...`?

src/riak_core_info_service.erl
```diff
@@ -110,7 +110,27 @@
...
 start_service(Registration, Shutdown, Provider, Handler) ->
-    riak_core_info_service_sup:start_service(Registration, Shutdown, Provider, Handler).
+    case verify_callable([Registration, Shutdown, Handler]) of
+        true ->
+            riak_core_info_service_sup:start_service(Registration, Shutdown, Provider, Handler);
+        {false, {_Module, _Function, _Arity}=Fun} ->
+            {error, {not_callable, Fun}}
+    end.
+
+verify_callable([]) ->
```
tburghart, line 120: Channeling Nick to point out that our style guidelines prefer not doing our own recursion.
This could be neatly handled with `lists:all/2` or, for reporting, `lists:filtermap/2`

macintux:
Do not approve this; `develop` is probably going to be rolled back a couple of commits to manage the breakage this PR is designed to fix.

src/riak_core_info_service.erl
macintux: I'm happier with custom recursion in cases where we fail early, although admittedly there are only 3 functions to check.

macintux:
Should be safe to review this PR now. It fortunately incorporates the earlier approved #898 and Doug's Thumbs fix to prevent an auto-merge.

/cc @tburghart @raghavkarol

macintux:
We may need to rethink the supervisor restart intensity and period values and/or `permanent` vs `transient`. If we register `eleveldb` and it throws errors from its callbacks, will the process keep restarting to the point where the supervisor shuts down?

macintux:
Manual testing points to a couple of issues:
* `erlang:function_exported/3` only works properly against loaded modules
* This broke `eleveldb` registration but no log messages were apparently emitted

Will keep digging.

macintux:
I was wrong. The log messages were present, just further back in the log than I expected.

```2017-03-08 09:04:31.955 [warning] <0.685.0>@riak_core_app:start_eleveldb_info_service:162 Info service failed to register for eleveldb: {not_callable,{eleveldb_metadata,handle_metadata_response,1}}```

Adding `code:ensure_loaded` to `verify_callable`.

macintux:
As I feared, an ill-behaved registration can force a restart of the supervisor (and thus lose any other registered consumers).

Ruminating.

raghavkarol:
+1

macintux:
The backend bits (eleveldb, leveldb*) have been merged to develop, so this PR should be fully functional without further branch-wrangling.

JeetKunDoug:
+1

PR #906, opened 2017-03-08
Removing vestiges of riak_search

fadushin:

PR #907, opened 2017-03-10
Remove legacy search, continued

nickelization:
This removes one more bit of code that we missed in #906. I posted about it in the review comments there before I realized the previous PR had already been merged.


bsparrow435:
+1

PR #916, opened 2018-02-27
let's make travis work again

ghost:
no point in building against the following -  R15B03, R15B02, R15B01, R15B


nickkeers (approved):
:+1:

PR #917, opened 2018-03-01
Add participate_in_2i_coverage riak option

russelldb:
Allows marking of nodes 'down' in calculations for 2i coverage
queries.  In create_plan, it queries each node for the value of the
participate_in_2i_coverage and treats those with the value set
explicitly to `false` as 'down' for the calculation of avaiable
vnodes.

See related riak_kv commit that adds a default value to the KV schema.
This is an operator feature. The operator must bring down the node,
change the config, and start the node, for it to have effect.

### DIFF
--- a/src/riak_core.erl
+++ b/src/riak_core.erl
@@ -159,7 +159,13 @@ standard_join(Node, Ring, Rejoin, Auto) ->
                                                   node(),
                                                   gossip_vsn,
                                                   GossipVsn),
-            {_, Ring5} = riak_core_capability:update_ring(Ring4),
+            ParticipateInCoverage = app_helper:get_env(riak_core,participate_in_coverage),
+            Ring4a =
+                riak_core_ring:update_member_meta(node(),
+                                                  Ring4,
+                                                  node(),
+                                                  participate_in_coverage, ParticipateInCoverage),
+            {_, Ring5} = riak_core_capability:update_ring(Ring4a),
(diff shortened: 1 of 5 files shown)


src/riak_core.erl
```diff
@@ -159,7 +159,13 @@ standard_join(Node, Ring, Rejoin, Auto) ->
                                                   node(),
                                                   gossip_vsn,
                                                   GossipVsn),
-            {_, Ring5} = riak_core_capability:update_ring(Ring4),
+            ParticipateInCoverage = app_helper:get_env(riak_core,participate_in_coverage),
```
ghost, line 162: All this Ring2, Ring3, Ring4, Ring4a, Ring5, Ring6 might be nicer as a fold - but that's outside the scope of this change request.

bryanhuntesl:
LGTM

ghost:
The structure looks OK - but I notice there are no tests for this new functionality - are they in a separate repository?

ghost:
Duh, just read comments above.

russelldb:
Don't you think maybe your comments about the schema from https://github.com/basho/riak_kv/pull/1664 should be on _this_ PR?

ghost:
I don't mind which PR is used to do it. the cuttlefish one could be merged in (to here) first.

russelldb:
I don't understand.

ranisen:
Looks ok for me.

PR #918, opened 2018-05-03
Move to gen_fsm_compat

Licenser:
With erlang 20 gen_fsm is marked as depricated to make place for
gen_statem. While gen_statem is sure great it feels just silly to
move perfectly good gen_fsm code to gen_statem just because.

As gen_fsm will eventually go away keeping it and just turning off
warnings also isn't the best solution it will just postpone the
problem.

This is the third option, gen_fsm_compat is a 1:1 copy of R19.3
gen_fsm that will keep working without deprication warnings and
in result make a move to 20+ a lot easier.


ghost (commented):
Great idea, although perhaps better for now to express the gen_fsm_compat dependency as git rather than hex.

Licenser:
I was about to do that but then I saw that exometer already is a hex dependency so it seems hex was already required one way or another.

ghost:
https://github.com/basho/riak_core/commit/d253e871258f73671da20a8b57e10b6fa719de65#r28837140

ghost:
Added comment on the change to hex dependency.

martincox:
Fixed in develop-3.0.

PR #920, opened 2018-05-21
Delete supervisor_pre_r14b04.erl

ghost:


martincox:
Will need to drop refs to this in the code base also, I guess? In `riak_core_vnode_sup:start_vnode/3` for example. Doesn't seem like a simple deletion of the mod, unless there's more to follow.

ghost:
Yes, interesting. If you leave the issue open I'll do an investigation mid June

https://github.com/basho/riak_core/blob/1d65cf705debafb5676063e9f5d59c9b545d7d46/src/riak_core_vnode_sup.erl#L40

PR #924, opened 2019-01-15
Develop 2.9

martinsumner:
Merge into basho repo develop 2.9.  2.9 hash two riak_core changes:

- There are changes to support the vnode soft limit monitoring, to allow for vnodes to report their approximate queue size in response to poll messages, so that Riak may route around slow vnodes (for example when choosing a put co-ordinator)

- The introduction of support for node worker-pool strategies (workers shared across vnodes on a given node) to be used to rate-limit the resources consumes by work which can be queued

### DIFF
--- a/docs/node_worker_pool.md
+++ b/docs/node_worker_pool.md
@@ -1,0 +1,116 @@
+# Node Worker Pool
+
+
+## Background
+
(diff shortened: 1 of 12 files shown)


martinsumner:
@martincox - develop 2.9 update for riak_core

PR #925, opened 2019-02-18
Update for OTP20/rebar3.

cmeiklejohn:
* gen_fsm_compat changes
* Fix seeds to use proper format
* Add nowarn_export_all on modules that explicitly export all.
* Switch to os:timestamp where possible.


jadeallenx:
Damn, brings back good memories. 😢

martinsumner:
@martincox 

We have conflicts with this now, as you have made some direct commits.  We need to work out how to resolve.

martincox:
My bad. :man_facepalming:

PR #929, opened 2019-03-07
remove and ignore rebar.lock

UlfNorell:


martinsumner:
Removing locks across the board.  This is causing continuous issues with getting a stable build as we change.  We will not .gitignore the rebar.lock in the root riak repo.

PR #930, opened 2019-03-07
Use Kyorai poolboy

ThomasArts:


martinsumner:
Chosen kyorai of devinus due to proving in riak_core_ng

PR #932, opened 2019-03-08
Fix calls to rand:seed in riak_core_claim

rjmh:


ThomasArts:
After discussion with @martinsumner it seems a delicate issue and we need to make sure it works by fixing the eqc tests first.

PR #934, opened 2019-03-14
Fix circular dependency.

martincox:
Remove folsom from included_applications - is started by exometer_core now, conflicts when building release.


nsaadouni:
+1

Issue #943, opened 2019-10-27
riak_core_claimant tick can occur after close


martinsumner:
When a node is leaving a cluster the final trigger to exit is a call to `riak_core_ring_manager:refresh_my_ring/0`

https://github.com/basho/riak_core/blob/9f7d0538e01e248dc4f5e0efff328737f23d8fb9/src/riak_core_ring_manager.erl#L377-L388

The refresh_my_ring call will be prompted from `riak_core_ring_handler:ensure_vnodes_started/1`:

https://github.com/basho/riak_core/blob/9f7d0538e01e248dc4f5e0efff328737f23d8fb9/src/riak_core_ring_handler.erl#L61-L99

This will in turn be prompted by the management tick in the `riak_core_vnode_manager`.

The `refresh_my_ring` logic will generate a fresh ring, persist, before prompting riak_core to close.

However, this will prompt the close process to start a structured shutdown of the application, and that shutdown might not be immediate.  As it is not immediate, before the shutdown is complete the next tick event may occur on the `riak_core_claimant`:

https://github.com/basho/riak_core/blob/9f7d0538e01e248dc4f5e0efff328737f23d8fb9/src/riak_core_claimant.erl#L632-L644

This can create an issue for `riak_ensemble`.  `riak_ensemble` is dependent on certain changes to the root ensemble being prompted by the claimant (a singleton claimant within the cluster). 

https://github.com/basho/riak_core/blob/9f7d0538e01e248dc4f5e0efff328737f23d8fb9/src/riak_core_claimant.erl#L830-L863

If the single to claimant is a node which has been set to `leave`, then we need that claimant node to prompt the removal changes to the root ensemble.  However, if the schedule tick required to make this happen occurs after the ring has been refreshed - this may lead to a bad situation.  In this case the refreshed ring will not have the local node() removed, but instead the local node() will be the only member of the ring.

Now when bootstrapping the root_ensemble as a result of the tick, the code will see all members of the root ensemble other than the local node as having been removed - and will spawn a process to prompt their removal.

https://github.com/basho/riak_core/blob/9f7d0538e01e248dc4f5e0efff328737f23d8fb9/src/riak_core_claimant.erl#L861

This is erroneous - in fact the local node is the only one which should be removed.

This leads to an intermittent failure of the `riak_test` `ensemble_remove_node` (which can made to fail fairly reliably by increasing timeouts).  It fails as the `riak_ensemble_manager:cluster/0` has no members.

[...]

PR #945, opened 2019-10-31
Develop 3.0 lastgaspring

martinsumner:
When a node leaves a cluster before it closes it will change the ring to a fresh ring (which knows only of the local node).  This is necessary to ensure that restarting the node does not lead to any attempt to rejoin.

There is an issue though, that the close of riak is not immediate.  It is, as one would expect, an orderly shutdown of the dependencies, before riak_core is the final application to close.

During this closing process, there are multiple processes which are monitoring the ring and may react to the change in the ring.  At this stage, as the shutdown is not complete these processes may take unnecessary action, and as the node is still connected to he cluster at an erlang level - this could leak into the rest of the cluster.

This was causing significant problems for `riak_ensemble`, where ensembles would be changed through the cluster incorrectly.  In some cases there would then be a version mismatch that meant that this was never corrected back.

This PR introduces 'lastgasp' metadata to the ring.  If the ring is being changed to reflect shutdown on departure this metadata is set _after_ the ring has been persisted (so it is not seen on restart).

There are related PRs then due in `riak_kv` and `riak_repl` with ring_handlers amended to check this `lastgasp` status before reacting to a ring change.

These changes have stabilised ensemble riak_test tests (without actually changing `riak_ensemble`), and also reduce unexpected crashes during the shutdown processes caused by the false starting of elections/vnodes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ dialyzer_unhandled_warnings
 _build
 rebar.lock
 .DS_Store
+nonode@nohost
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -35,6 +35,7 @@
          commit/0,
          clear/0,
          ring_changed/2,
+         pending_close/2,
          create_bucket_type/2,
          update_bucket_type/2,
          bucket_type_status/1,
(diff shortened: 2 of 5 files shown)


martinsumner:
https://github.com/basho/riak_core/issues/943
https://github.com/basho/riak_ensemble/issues/129

ThomasArts (approved):
QuickCheck tests pass

PR #947, opened 2019-11-05
Update .travis.yml

martinsumner:
Test with dialyzer, xref and eunit on travis


martinsumner:
@nsaadouni could you enable for riak_core?  Hopefully should work with this PR

PR #948, opened 2019-11-05
Fix broken test

martinsumner:
Refactoring broke a test - need to stop and start claiminant from `refresh_my_ring` test

### DIFF
--- a/src/riak_core_claimant.erl
+++ b/src/riak_core_claimant.erl
@@ -84,6 +84,16 @@
 %%% API
 %%%===================================================================
 
+
+-ifdef(TEST).
+
+-export([stop/0]).
+
+stop() ->
+    gen_server:call(claimant(), stop).
+
+-endif.
+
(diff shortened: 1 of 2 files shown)


martinsumner:
@ThomasArts - I must have messed up somewhere ... unit tests were not passing after my latest PR.  This fixes it.

src/riak_core_claimant.erl
```diff
@@ -84,6 +84,16 @@
 %%% API
 %%%===================================================================
 
+
+-ifdef(TEST).
+
+-export([stop/0]).
+
+stop() ->
```
ThomasArts, line 92: By only including this in test, you signal that this `stop()` function should better not be called from any other module. But, clearly, the handle_call is there and stop can therewith be triggered.

So the added value of the `ifdef(TEST)` is rather limited.

ThomasArts (approved):

PR #949, opened 2019-11-09
fixing platform deprecation, newlines, and lager upgrade

v-kat:
I wasn't able to get all the checks passing so let me know if there's any other issues

### DIFF
--- a/src/riak_core_gen_server.erl
+++ b/src/riak_core_gen_server.erl
@@ -45,16 +45,16 @@
 %% compliance with the License. You should have received a copy of the
 %% Erlang Public License along with this software. If not, it can be
 %% retrieved via the world wide web at http://www.erlang.org/.
-%% 
+%%
 %% Software distributed under the License is distributed on an "AS IS"
 %% basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See
 %% the License for the specific language governing rights and limitations
 %% under the License.
-%% 
+%%
 %% The Initial Developer of the Original Code is Ericsson Utvecklings AB.
 %% Portions created by Ericsson are Copyright 1999, Ericsson Utvecklings
 %% AB. All Rights Reserved.''
-%% 
+%%
 %%     $Id$
 %%
 -module(riak_core_gen_server).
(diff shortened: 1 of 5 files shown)


martinsumner:
Thanks!  Will take a look at this on Monday.

src/riak_core_tcp_mon.erl
```diff
@@ -458,7 +458,7 @@ ssl_test_() ->
         spawn(fun () ->
             %% server
             {ok, S} = ssl:transport_accept(LS),
-            ok = ssl:ssl_accept(S),
+            ok = ssl:handshake(S),
```
martinsumner: http://erlang.org/doc/man/ssl.html#handshake-1 - looks like it cannot just return OK, will always return a socket

v-kat: Thanks. Updated. I might have not run the ssl suite locally :(

src/riak_core_gen_server.erl
```diff
@@ -1015,8 +1015,8 @@ format_status(Opt, StatusData) ->
 		      Name
 	      end,
     Header = lists:concat(["Status for generic server ", NameTag]),
-    Log = sys:get_debug(log, Debug, []),
-    Specfic = 
+    Log = sys:get_log(Debug),
```
martinsumner: src/riak_core_gen_server.erl
1018: Call to missing or unexported function sys:get_log/1

martinsumner: Dialyzer doesn't like this in OTP 20

martinsumner:
The SSL riak_test tests all pass.  However issue with sys:get_log/1.

src/riak_core_tcp_mon.erl
```diff
@@ -458,8 +458,8 @@ ssl_test_() ->
         spawn(fun () ->
             %% server
             {ok, S} = ssl:transport_accept(LS),
-            ok = ssl:ssl_accept(S),
-            ssl_recv_loop(S)
+            {ok, SslSock} = ssl:handshake(S),
```
martinsumner: This will need to use `deprecated_21` to switch back to ssl_accept in OTP 20

martinsumner:
To pass `make test`, the branch needs to be updated from basho riak_core `develop-3.0` as this resolves a problem which cancels a number of the tests.

martinsumner:
This looks good now after being updated from upstream.  CI tests pass.

martinsumner (approved):

PR #954, opened 2020-02-12
Fix log

martinsumner:
https://github.com/basho/riak/issues/1004


ThomasArts (approved):
That was a subtle one. good catch

PR #960, opened 2020-05-08
When using Eunit related functions, TEST must be defined

ThomasArts:


ThomasArts:
This is encapsulated in next PR

PR #968, opened 2021-05-12
Update rebar.config to use https

l2dy:
Unify git protocol to the secure `https://`.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -20,16 +20,16 @@
 {plugins, [{eqc_rebar, {git, "https://github.com/Quviq/eqc-rebar", {branch, "master"}}}]}.
 
 {deps, [
-  {lager, {git, "git://github.com/erlang-lager/lager.git", {tag, "3.8.0"}}},
+  {lager, {git, "https://github.com/erlang-lager/lager.git", {tag, "3.8.0"}}},
   {poolboy, {git, "https://github.com/basho/poolboy.git", {tag, "riak_kv-3.0.0"}}},
-  {riak_sysmon, {git, "git://github.com/basho/riak_sysmon.git", {tag, "2.2.0"}}},
-  {clique, {git, "git://github.com/basho/clique.git", {tag, "0.3.11"}}},
-  {eleveldb, {git, "git://github.com/basho/eleveldb.git", {tag, "riak_kv-3.0.0"}}},
+  {riak_sysmon, {git, "https://github.com/basho/riak_sysmon.git", {tag, "2.2.0"}}},
+  {clique, {git, "https://github.com/basho/clique.git", {tag, "0.3.11"}}},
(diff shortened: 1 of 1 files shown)


martinsumner:
A while back, everything across the active basho repos were standardised on `git` not `https`, I don't recall the reasoning at the time.

I've looked around for stuff on the relative security of https via ssh/git - [found this](https://gist.github.com/grawity/4392747).

I have no objections to changing back, but do you have a specific reference or reasoning for https vs ssh?  Based on my reading and limited understanding it isn't clear that one option is secure, but the other not.  Apologies if the reason is obvious, it might just be me being dumb.

martincox:
You can always force git to use your preferred protocol by modifying your ~/.gitconfig like so:

> [url "https://"]
>         insteadOf = git://
> [url "https://"]
>         insteadOf = ssh://
> [url "https://github.com/"]
>         insteadOf = git@github.com:

I use https because of firewall restrictions. There will be various deps that aren't consistent with the basho repos, some may use https and others ssh, so the only way to truly control this is by setting it on your machine and forcing it across the board.

l2dy:
> I have no objections to changing back, but do you have a specific reference or reasoning for https vs ssh? Based on my reading and limited understanding it isn't clear that one option is secure, but the other not. Apologies if the reason is obvious, it might just be me being dumb.

HTTPS pull is anonymous for public repositories, great fit for CI and throwaway environments. SSH on the other hand requires users to set up SSH keys and add those keys to their GitHub accounts, which is more effort involved and managing secrets can be difficult in certain environments. Also there is the firewall thing @martincox mentioned.

git:// is vulnerable to MITM attacks, and it's not even mentioned in GitHub's clone options and docs. https://docs.github.com/en/github/getting-started-with-github/about-remote-repositories#choosing-a-url-for-your-remote-repository

l2dy:
@martinsumner What do you think?

martinsumner:
There's not much point in updating riak_core, unless I do all the deps ... and that's quite a long-winded and fiddly job.  Given the existence of the workaround suggested by @martincox - I'm minded to leave it at `git` and let people implement `insteadOf` in their local config if they have particular concerns.

I checked some other major Erlang projects - Cowboy, VerneMQ and they both use `git` references in rebar.config, so I don't think it is a fault, and so it should be left for the individual to override if they wish.

l2dy:
OK. My point is that git:// is insecure and not even officially supported by GitHub, but sure the workaround is valid.

l2dy:
On January 11, 2022, GitHub will start to temporarily stop accepting the unencrypted Git protocol, and on March 15, 2022 changes will be made permanent. https://github.blog/2021-09-01-improving-git-protocol-security-github/

@martinsumner All the deps indeed need to be fixed and we'd better do it in advance.

martinsumner:
As we update repos with new PRs we're going to start changing these from git to https.  Release 3.0.8 will cover quite a few dep changes, and we can pick the rest as part of 3.0.9

PR #972, opened 2021-10-01
Refactor leave - balance and tail violations

martinsumner:
Some curious and unexpected cluster plans when running leave operations, prompted a review of the lead code.

This PR corrects a number of issues found:

- The first part of a leave plan is to plan to handoff all vnodes on the leaving nodes to other members of the cluster, in a way that does not break the target_n_val.  The logic of this attempt_simple_leave function was incorrect, as it would produce plans that would include tail violations (where the location was safe when the ring was viewed as a list, but not when it was considered as a ring where the head wraps round to the tail).  This issue is corrected.

- The attempt_simple_leave did not attempt to balance the vnodes in any way, leading in some cases to unbalanced plans (and leaves that might not be supportable as a receiving node may not have sufficient space).  There is now a preference to make choices that lead to more balanced outcomes - and a workaround to skip this stage and move straight to a freshly balanced cluster plan (`riak_core.full_rebalance_onleave`).

- When a full rebalance was forced, as it was not possible to produce a safe outcome to `attempt_simple_transfer`, then a full rebalance would be triggered.  However a deprecated function was used, one which could lead to tail violations (ignoring the improvements made in Riak 2.2.5).  The updated function is now used.


src/riak_core_gossip.erl
```diff
@@ -367,73 +371,287 @@ remove_from_cluster(Ring, ExitingNode) ->
...
-                Other = hd(lists:delete(ExitingNode, Members)),
-                TempRing = lists:foldl(
-                             fun({I,N}, R) when N == ExitingNode ->
-                                     riak_core_ring:transfer_node(I, Other, R);
-                                (_, R) -> R
-                             end,
-                             Ring,
-                             AllOwners),
-                riak_core_claim:claim_rebalance_n(TempRing, Other)
+                HN = hd(lists:delete(ExitingNode, Members)),
```
ThomasArts, line 386: I see that the code is not new, but are we absolutely sure that Members always has at least 2 elements, such that we have an element left to take as head?

src/riak_core_gossip.erl
```diff
@@ -367,73 +371,287 @@ remove_from_cluster(Ring, ExitingNode) ->
...
-    attempt_simple_transfer(Seed, Ring, Owners,
-                            TargetN,
-                            ExitingNode, 0,
-                            [{O,-TargetN} || O <- riak_core_ring:claiming_members(Ring),
-                                             O /= ExitingNode]).
-attempt_simple_transfer(Seed, Ring, [{P, Exit}|Rest], TargetN, Exit, Idx, Last) ->
-    %% handoff
-    case [ N || {N, I} <- Last, Idx-I >= TargetN ] of
+-ifdef(TEST).
+-type transfer_ring() :: [{integer(), term()}].
```
ThomasArts, line 400: interesting... a different type when testing.

src/riak_core_gossip.erl
```diff
@@ -367,73 +371,287 @@ remove_from_cluster(Ring, ExitingNode) ->
...
+                                        force_rebalance.
+attempt_simple_transfer(Seed, Ring, ExitingNode) ->
+    ForceRebalance =
+        app_helper:get_env(riak_core, full_rebalance_onleave, false),
+    case ForceRebalance of
+        true ->
+            force_rebalance;
+        false ->
+            TargetN = app_helper:get_env(riak_core, target_n_val),
+            Owners = riak_core_ring:all_owners(Ring),
```
ThomasArts: This is almost the same code as in line 383 and following

src/riak_core_gossip.erl
martinsumner: It is quite a long chain of functions - but we get here because at some stage there has been a call to riak_core-claimant:maybe_update_ring/5 - https://github.com/basho/riak_core/blob/1f7e204548148f7dda9727113838d40b559310e3/src/riak_core_claimant.erl#L1235-L1254

This checks that claiming_members is non-empty

src/riak_core_gossip.erl
martinsumner: Refactored so Owners/Members are determined only once

src/riak_core_gossip.erl
```diff
@@ -367,73 +371,291 @@ remove_from_cluster(Ring, ExitingNode) ->
...
+            {4, n4}, {5, n5}, {6, n3}, {7, n2}],
+    SomeTime = {1632,989499,279637},
+    FixedSeed = rand:seed(exrop, SomeTime),
+    {ok, R1} =
+        simple_transfer(R0,
+                        {fun test_ring_fun/3, 3, n4},
+                        R0,
+                        {FixedSeed,
+                            [],
+                            lists:keydelete(n4, 1, count_nodes(R0))}),
```
ThomasArts, line 573: The Counts are computed using Members, which does not include the `leaving` node, hence allowed to delete the leaving node here.

src/riak_core_gossip.erl
```diff
@@ -367,73 +371,291 @@ remove_from_cluster(Ring, ExitingNode) ->
...
+%% although this may result in a large number of transfers
+-spec attempt_simple_transfer(transfer_ring(),
+                                term(),
+                                random:ran(),
+                                [{integer(), term()}],
+                                [term()]) ->
+                                    {ok, transfer_ring()}|
+                                        target_n_fail|
+                                        force_rebalance.
+attempt_simple_transfer(Ring, ExitingNode, Seed, Owners, Members) ->
```
ThomasArts, line 423: The function relies on the fact that the list of owners is sorted. However, the resulting Ring is not necessarily sorted.

ThomasArts (approved):

PR #977, opened 2021-10-27
Fix property to test even when nodes are unavailable

ThomasArts:
But do not test the hot_node in that case


src/riak_core_coverage_plan.erl
```diff
@@ -30,7 +30,7 @@
 -include_lib("eqc/include/eqc.hrl").
 -export([prop_cover_partitions/0,
             prop_distribution/0,
-            prop_find_coverage_partitions/0, 
+            prop_find_coverage_partitions/0,
```
ThomasArts, line 33:
```suggestion
```

martinsumner (approved):

PR #979, opened 2021-11-04
Mas i1801 monitorworkerq

martinsumner:
Add monitoring via riak stats to the worker queues.  Monitor both queue time, and work time.

### DIFF
--- a/src/riak_core_worker_pool.erl
+++ b/src/riak_core_worker_pool.erl
@@ -125,9 +125,6 @@ ready({work, Work, From} = Msg,
         #state{pool=Pool, queue=Q, monitors=Monitors} = State) ->
     case poolboy_checkout(Pool, State) of
         {full, State0} ->
-            riak_core_stat:update({worker_pool,
-                                    queue_event,
-                                    State#state.pool_name}),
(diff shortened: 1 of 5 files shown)


martinsumner:
https://github.com/basho/riak_test/pull/1359

src/riak_core_worker_pool.erl
```diff
@@ -278,3 +291,56 @@ discard_queued_work(Q, Mod) ->
             ok
     end.
 
+-spec poolboy_checkin(pid(), pid(), state()) -> {ok, state()}.
+poolboy_checkin(Pool, Worker, State) ->
```
ThomasArts: It is not wrong, but changing state as a side effect of a little function called makes it harder to follow the code. I kind of like to have state updates very declarative in a small scope.
In this case that would mean that you pass the pool name in as last argument, instead of the state, and return the checkout or an error. Then you do the state update in the function calling this one...

I do see that this may give some code duplication and probably it's fine this way, but it is harder to get an idea of what happens to the state in the main "loop".

ThomasArts (approved):

ThomasArts (approved):

PR #981, opened 2022-02-07
Fix2 bg manager eqc

ThomasArts:
Fix the model such that operations get_lock and get_token with dead processes do not update the state. Those operations are no-ops, but it takes some time for the system to have the monitor DOWN signal actually make it a no-op.


ThomasArts:
This does not work, the implementation is non-deterministic and this fix just reveals another issue

PR #992, opened 2022-11-23
Mas i1121 passringdir

martinsumner:
Break down the functions required for reip, so that it can be called from riak_kv_console without knowledge of the cluster name or of the ring directory.  This makes it easier to write a reip admin command to be used when riak_core is not loaded with its standard configuration


hmmr:
+1

Issue #994, opened 2023-01-02
Handoff and TCP RECV timeout


martinsumner:
## Summary of Confusion/Problem:

Handoffs on busy riak_kv clusters often fail, with the reported issue being TCP RECV timeouts.

The code of handoff includes references to support very old versions:

https://github.com/basho/riak_core/blob/riak_kv-3.0.12/src/riak_core_handoff_sender.erl#L129-L133

Also there is reference to a re-write to resolve confusion happening soon:

https://github.com/basho/riak_core/blob/riak_kv-3.0.12/src/riak_core_handoff_sender.erl#L198-L202

Handoffs timing out are disruptive, as the handoff must start from the beginning (in terms of the fold over the vnode), and re-send all of the data at the next attempt.

When leaving/joining this can be controlled through handoff concurrency, but this isn't perfect, and still can lead to continuous failures.

Hinted handoffs can be even more problematic, particular when receiving vnodes are already subject to high loads.  There is very little tuning/testing that can be done to predict what a safe transfer limit might be.  There is also a vicious circle that can form - as if hinted handoffs fail, the receiving cluster may be subject to increasing amounts of anti-entropy and read-repair load that will cause hinted handoffs to continue to fail. 

This issue is exacerbated that the timeout's reported as TCP RECV timeouts are not necessarily low-level network timeouts, but are more likely as a result of the receiver failing to respond to a OSI L7 SYNC message.  There is also a confusion of timeout settings:

Calling the function `get_handoff_receive_timeout/0` returns the handoff_timeout, which defaults to the TCP_TIMEOUT:

https://github.com/basho/riak_core/blob/riak_kv-3.0.12/src/riak_core_handoff_sender.erl#L514-L515

There is also a different `handoff_receive timeout` on the sender, based on a receiver setting:

https://github.com/basho/riak_core/blob/riak_kv-3.0.12/src/riak_core_handoff_sender.erl#L293_L298

Where the default is to be a third of the receiver setting, if set, but by default it is not set, and does not default to a third of the default chosen when unset on the receiver side:

https://github.com/basho/riak_core/blob/riak_kv-3.0.12/src/riak_core_handoff_receiver.erl#L78

How are these timeouts to be tuned to prevent timeouts?  A lower handoff_receive_timeout on the sender side, will lead to more frequent SYNCs - and hence make a backlog leading to a timeout less likely; but have different consequences within the receiver.

## Proposal:

There are some broad changes proposed:

[...]

PR #995, opened 2023-01-03
Mas i994 handoffsync

martinsumner:
Refactor handoff to avoid TCP RECV error - see https://github.com/basho/riak_core/issues/994.

- Clarify meaning of terms:

  - `handoff_timeout` if the receiver does not respond to a sender message within this period, then the sender should assume the receiver is down and exit the fold.  The sender can exchange sync messages every `handoff_acksync_threshold` batches to ensure that batches do not accumulate at the receiver causing the sender to fail to get a timely response to a future request. 

  - `handoff_receive_timeout`  if the receiver does not receive a message from the sender within this timeout, it will assume that the sender is down and exit the process.  The sender will interrupt the fold if it has not sent a batch in the last `handoff_receive_timeout div 3` to send a sync message as a keepalive to prevent this timeout if it is alive.

- The default value of `handoff_acksync_threshold` is changed to 1 from 25.  At 25 batches, and 250 bytes every object size, a backlog of 100K handoff items could develop at the receiver - requiring a sub-1ms response time for each handoff item to avoid triggering the timeout.  Alternative default of 1 is less likely to see the timeout, at the small cost of an additional sync round-trip per batch.

- A `handoff_batch_threshold_count` has been added to generate a threshold for a batch when this count of objects is reached.  The `handoff_batch_threshold` is also respected (which triggers a batch on size in bytes).  This is to prevent very large batches by count when object sizes are small.

- The ?PT_MSG_OLDSYNC is now properly deprecated.  The receiver will still respond, should a cluster contain a handoff sender still on a previous version, but the sender will now only send the ?PT_MSG_SYNC message.

- Reformatting of code to fit column width.

- Standardise handling of sync messages through `riakc_core_handoff_sender:send_sync/3` function, and now error logs raised at the actual message where an error is received.  The item queue is redacted in the error generated by a throw, so that important details are not obfuscated by a large list of binaries.

- A new log has been added to log progress in sending batches every `handoff_acklog_threshold` batches (default 100).  This allows for progress to be tracked through log indexers, rather than manually by operators calling the `riak admin transfers` command.

- All code for (legacy) non-batched handoff removed.

### DIFF
--- a/src/riak_core.erl
+++ b/src/riak_core.erl
@@ -160,7 +160,7 @@ standard_join(Node, Ring, Rejoin, Auto) ->
             Ring6 = maybe_auto_join(Auto, node(), Ring5),
             riak_core_ring_manager:set_my_ring(Ring6),
             ok = riak_core_gossip:send_ring(Node, node()),
-            ok = riak_core_metadata_manager:waitfor_exchange(Node)
+            ok = riak_core_metadata_manager:attempt_exchange(Node)
 
     end.
 
--- a/src/riak_core.erl
+++ b/src/riak_core.erl
@@ -140,25 +140,28 @@ standard_join(Node, Ring, Rejoin, Auto) ->
             {error, different_ring_sizes};
         _ ->
             GossipVsn = riak_core_gossip:gossip_version(),
-            Ring2 = riak_core_ring:add_member(node(), Ring,
-                                              node()),
+            Ring2 =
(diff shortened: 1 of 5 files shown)


martinsumner:
https://github.com/basho/riak_test/pull/1369

yorkshire-steve (approved):

yorkshire-steve (approved):

PR #1002, opened 2023-04-18
Swapping to another algorithm

ThomasArts:
This PR allows a new algorithm for node placement in a ring where rack awareness, or location awareness, is taken into consideration.
Thus, one can provide a target n-val for locations in this new version and consecutive nodes will be placed at distinct locations.


ThomasArts:
replaced by PR1003

aramallo:
👏🏼

PR #1008, opened 2023-05-26
Improved location claim algorithm

martinsumner:
This PR adds a new v4 claim algorithm, which significantly improves the efficiency of claim with location awareness.

The algorithm has a brute force approach, whereby it will in the plan stage potentially do an exhaustive search of the possible swaps and moves, in order to provide a ring which meets the target n_val for both nodes and locations.  Firstly by trying to minimise the number of changes, and if that is unsuccessful it will attempt to find a valid solution without concern for the plan complexity. 

To enable the feature, set via riak.conf:

`choose_claim_fun` -> `choose_claim_v4`
`full_rebalance_onleave` -> `on`
`target_n_val` -> `4` (default)
`target_location_n_val -> `3` (default)

riak_test testa are added and updated in https://github.com/basho/riak_test/pull/1375.

There is a write-up in docs/claim-version4.md.

In some cases cluster plans may take much longer than at present, potentially > 60s.

The new claim algorithm was developed by [Ulf Norell and Thomas Arts](http://www.quviq.com/contact-2/about/).


martinsumner:
@systream - as you did the original location awareness, I suspect you may have an interest in testing this.

martinsumner:
https://github.com/basho/riak_test/pull/1375

ThomasArts (approved):
Looks good to me.

PR #1014, opened 2024-02-16
Fix  dbg:stop_clear/0 is deprecated and will be removed in OTP 27;

asyncmind0:
- using dbg:stop/0 instead

### DIFF
--- a/src/riak_core_capability.erl
+++ b/src/riak_core_capability.erl
@@ -133,6 +133,7 @@ register(Capability, Supported, Default, LegacyVar) ->
 register(Capability, Supported, Default) ->
     register(Capability, Supported, Default, undefined).
 
+            
 %% @doc Query the current negotiated mode for a given capability, throwing an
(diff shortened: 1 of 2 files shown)


martinsumner:
The function `dbg:stop` changed OTP 24 -> OTP 25 to make it the same as `dbg:stop_clear/0`, but in OTP 24 the two functions still have different behaviour:

https://www.erlang.org/docs/24/man/dbg#stop-0
https://www.erlang.org/docs/24/man/dbg#stop_clear-0
https://www.erlang.org/docs/25/man/dbg#stop-0

So to implement this changeI think  you need to check the OTP_RELEASE and do stop if OTP_RELEASE >= 25, and stop_clear otherwise.

There's a question over whether we should just remove riak_core_tracer altogether.  I don't think it is used anywhere other than `riak_kv_get_fsm`.  @nsaadouni @tburghart - do either of your teams make use of this riak_core_tracer?

PR #1015, opened 2026-04-13
Erlang 28 Workflow

albsch:


albsch:
Sorry, my mistake...

Bob-The-Marauder:
See if OpenRiak will take your update? https://github.com/OpenRiak/

albsch:
The two projects have diverged too much, I think. It's the other way around, `riak_core_lite` tries to incorporate features from `OpenRiak` while still trying to be as simple as possible.

